PR #2, opened 2016-06-13
Snapdragon: Adapt sysroot instructions

julianoes:
Since the file Flight_qrlSDK.zip has disappeared from the Intrinsyc
download page, we go back to the trusty sysroot provided by Qualcomm.


LorenzMeier:
Thanks!

PR #3, opened 2016-06-22
re-add all things that got lost when importing from the previous repo

bkueng:
Quite a few things got lost in the transition


kd0aij:
looks like this commit also disappeared: https://github.com/kd0aij/Devguide/commit/d4a55184aa17aeb66c72d5b8bcca2744836ee83e

bkueng:
@kd0aij you're right, there was more. I didn't see that the old repo was moved and more commits were pushed.
See here: https://github.com/PX4/Devguide/pull/4

PR #11, opened 2016-07-11
Add documentation for new logger & system-wide replay

bkueng:


advanced-replay.md
```diff
@@ -0,0 +1,129 @@
...
+  to be disabled for replay.
+
+- Optional: setup parameter overrides in the file
+  `build_posix_sitl_default_replay/src/firmware/posix/rootfs/replay_params.txt`.
+  This file should contain a list of `<param_name> <value>`, like:
+```
+EKF2_GB_NOISE 0.001
+```
+  By default, all parameters from the log file are applied. When a parameter
+  changed during recoding, it will be changed as well at the right time during
```
kd0aij, line 72: typo: recording

kd0aij:
@bkueng Thanks, that's very helpful.

advanced-replay.md
bkueng, line 72: Thanks, fixed

PR #13, opened 2016-07-12
Updated PPA - gcc-arm-embedded toolchain

eduherminio:
PPA by terry.guo that contained the toolchain doesn't seem to be available any more.
I wasn't able to continue my Linux installation following that step in the guide, so after a little bit of research I found the alternative solution [here](http://wiki.paparazziuav.org/wiki/Installation/Linux#gcc-arm-embedded_toolchain).
`sudo add-apt-repository ppa:team-gcc-arm-embedded/ppa`
Finally I've managed to end the installation and a first build with no additional trouble.

PR #21, opened 2016-07-27
Update Ubuntu PPA for gcc-arm-none-eabi

julianoes:
- Use new PPA as suggested by #13.
- Actually install the new toolchain from the ppa (install gcc-arm-embedded instead of gcc-arm-none-eabi). This installs version 5.4.
- Suggest to use version 5.4 in boutique installation as well.
- Removal of version 4.9 in boutique installation because gdb is unable to open the build artefacts which is probably a bug resolved in versions 5+.

The firmware built by gcc version 5.4.1 has been tested on the bench, no errors observed so far.

### DIFF
--- a/starting-installing-linux-boutique.md
+++ b/starting-installing-linux-boutique.md
@@ -129,9 +129,9 @@ Log out and log in for changes to take effect! Also remove the device and plug i
 
 ### Toolchain Installation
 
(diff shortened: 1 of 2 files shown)


LorenzMeier:
Nice!

PR #26, opened 2016-08-02
Logging: add SD Cards performance results

### LINKED ISSUE #25: Add docs for best logging practices
@bkueng Could you create a page or section for the SD card recommendations and a brief explanation why this matters (500 ms allowed block time per SD card standard and associated buffering needs?)

I think that would be very valuable for developers.

bkueng:
fixes #25


LorenzMeier:
Nice!

PR #31, opened 2016-08-08
Add new info about bash on windows.

lovettchris:
The PX4 build succeeds using [BashOnWindows](https://github.com/Microsoft/BashOnWindows)

### DIFF
--- a/starting-installing-windows.md
+++ b/starting-installing-windows.md
@@ -11,3 +11,10 @@ Download and install these on your system:
   * [PX4 USB Drivers](http://pixhawk.org/static/px4driver.msi) (32/64 bit systems)
 
 Now continue to run the [first build](starting-building.md)!
+
+## NEW! Bash on Windows
+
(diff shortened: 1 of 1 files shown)


LorenzMeier:
That's pretty cool. How do you install the env for bash on windows?

lovettchris:
You can download [windows 10 anniversary update](https://www.microsoft.com/en-us/software-download/windows10) then enable the windows feature called "Windows Subsystem for Linux (Beta)" using the "Turn Windows features on or off" dialog available from the "Programs and Features" control panel.  Then simply open a command prompt and type "bash".

lovettchris:
Note there's a few more steps to get openjdk, using webupd8 repo outlined [here](http://stackoverflow.com/questions/32942023/ubuntu-openjdk-8-unable-to-locate-package).  But this doesn't seem to be needed for the build.

bkueng:
rebased & applied

PR #32, opened 2016-08-18
Updated the blsp.config documentation to recommend the use of the 2-w…

jywilson:
…ire notation to prevent the use of the pins on the same connector for RTS-CTS flow control.


jywilson:
Thanks for the fast merge!  Much appreciated.

PR #37, opened 2016-08-27
Adding Crazyflie 2 hardware entry

dennisss:
Going along with https://github.com/PX4/Firmware/pull/5360.


hardware-crazyflie2.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Usage
+
+The power button has the same functionality as with the stock firmware.
+
+Once the right yellow LED has stopped blinking and the bottom left blue light has started blinking, the boot sequence is complete.
+
+Connect via USB to QGC to finish configuration:
+- Note: Do not use the Airframe tab of QGC
+- Manually set `SYS_AUTOSTART` to 4900 in the parameter options and `SYS_AUTOCONFIG` to 1 (and then restart).
```
LorenzMeier: It sounds like we should auto-detect the board instead. See https://github.com/PX4/Firmware/blob/master/ROMFS/px4fmu_common/init.d/rcS#L158. It would be better to automatically set the SYS_AUTOSTART value and SYS_AUTOCONFIG in the startup script if the board config is detected.

Then we can scrap the entire section.

LorenzMeier:
You don't have a flight video by chance which we could embed here?

dennisss:
It's a bit rough right now (https://youtu.be/jD7XJkyekpQ), drivers seem to be working, but I'll need to do some PID tuning to get an 'embed-able' video. I think it's the only quad in it's size class on PX4, so I've started with the default settings which are a bit high.

hardware-crazyflie2.md
dennisss: Updated it to auto-detect if not already set here: https://github.com/PX4/Firmware/pull/5360/commits/3c210c1cb545cbaa8a2b56338ff4b818180ae2d3#diff-e5c16c5209413f30e964883f74027b51R172

dennisss:
Updated with a better short video of manual flight. PID gains updated in original Firmware PR.

PR #48, opened 2016-10-03
document out-of-tree modules

bkueng:
...and make sure RPi is called as root.


LorenzMeier:
Gitbook test failed

bkueng:
saw that, but I think it's the gitbook installation. I wouldn't know which of my changes could have caused that.

julianoes:
@bkueng Can you rebase and force push to trigger another build? Or ammend something to the last commit to change the hash.

bkueng:
all good now

PR #56, opened 2016-10-18
Fixing link with raspberry pi page

barnjamin:
Broken link, I _think_ this should fix it assuming all .md suffixes are converted to HTML if they're found?


LorenzMeier:
Thanks!

PR #57, opened 2016-10-20
Flightplot updates

bkueng:
update repo link & adds an iframe for binary downloads

Not sure if the iframe works, in the preview it did not render.


LorenzMeier:
You can't use HTML in Gitbook. Can you try to find an alternate method for that?

bkueng:
Are you sure? https://mlewistw.gitbooks.io/gitbook-introduction/content/chapter_3__other_cool_things/html_is_allowed.html suggests otherwise.
And https://toolchain.gitbook.com/syntax/markdown.html#html too.

LorenzMeier:
It doesn't actually work. The build passed but failed on deployment. I reverted it to be a simple link.

PR #73, opened 2016-12-10
Additional VTOL Key Configuration Parameter

GabrielKoenig:
RC_MAP_TRANS_SW should be listed in Key Configuration Parameter


LorenzMeier:
Thanks!

PR #74, opened 2016-12-11
Invert PWM signal instead of RC

GabrielKoenig:
The correct inversion of a channel should be added.


LorenzMeier:
Thanks, awesome!

PR #76, opened 2016-12-17
(typo) roll is controlled by aileron

Boyang--Li:
I think here is a typo.
Roll is controlled by aileron rather than elevator.


LorenzMeier:
Thanks!

PR #79, opened 2017-01-11
git examples, show workflow

ChristophTobler:
I added a first example on how to contribute. It would be good if we can add common pitfalls like merge conflics etc.?

### DIFF
--- a/git-examples.md
+++ b/git-examples.md
@@ -7,6 +7,7 @@ Adding a feature to PX4 follows a defined workflow. In order to share your contr
 * Fork the Firmware (see [here](https://help.github.com/articles/fork-a-repo/#fork-an-example-repository))
 * Clone your fork to your local computer<br>
 ```sh
+cd ~/wherever/
 git clone https://github.com/<your git name>/Firmware.git
 ```
 * Go into the new directory, initialize and update the submodules, and add the original upstream Firmware<br>
@@ -15,12 +16,18 @@ cd Firmware
 git submodule update --init --recursive
 git remote add upstream https://github.com/PX4/Firmware.git
 ```
-* Make the changes that you want to add
+* You should have now two remote repositories: One repository is called upstream that points to the PX4 Firmware,
+and one repository that points to your own fork of the PX4 repository. 
+* This can be checked with the following command:
+```sh
+git remote -v
+```
+* Make the changes that you want to add to the current master.
 * Create a new branch with a meaningful name that represents your feature<br>
 ```sh
-git checkout -b <your branch name>
+git checkout -b <feature branch name>
 ```
-you can use the command ```git branch``` to make sure you're on the right branch
+you can use the command ```git branch``` to make sure you're on the right branch.
 * Add your changes that you want to be part of the commit<br>
 ```sh
 git add .
@@ -30,10 +37,11 @@ to add all your files or<br>
 git add <file name>
 ```
 to add single files
-* Commit your files that you added one step before and add a meaningful message explaining your changes<br>
+* Commit the added files with a meaningful message explaining your changes<br>
 ```sh
 git commit -m "<your commit message>"
 ```
+For a good commit message, please refer to [Contributing](starting-contributing.md) section.
 * Some time might have passed and the [upstream master](https://github.com/PX4/Firmware.git) has changed. PX4 prefers a linear commit history and uses [git rebase](https://git-scm.com/book/de/v1/Git-Branching-Rebasing). To include the newest changes from upstream in your local branch, switch to your master branch<br>
 ```sh
 git checkout master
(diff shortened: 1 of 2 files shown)


kd0aij:
Nice! That will be very useful to new contributors, (and me too). 
Shouldn't that be a forced push to origin after rebase, though?

ChristophTobler:
@kd0aij Glad you appreciate it. We had that discussion right now :) A forced push should only be necessary if you already pushed to your fork, then some new upstream master changes come up, you rebase locally and push again (if I got that correctly). But we should add that to pitfalls

kd0aij:
re: commit messages, a pointer to here would be good: http://dev.px4.io/starting-contributing.html#commits-and-commit-messages

Stifael:
@kd0aij the pointer is already in there at https://github.com/PX4/Devguide/pull/79/commits/7c2eebb9722435683df3ffd312d9634000cc2778#diff-9fbaa3fefb54bf7f66ffad4b91886548R44.

Stifael (commented):
There is wrong comment. It should say: to your fork repo...

ChristophTobler:
@bkueng Can you have a quick look at it? I'd say we merge this and add more examples as we stumble across them

kd0aij:
what is the "PX4 submodule"?

git-examples.md
```diff
@@ -0,0 +1,63 @@
+# GIT Examples
+## Adding a feature to PX4
```
bkueng: Maybe 'contributing code to PX4'? A feature seems to specific, as it could also be a bug-fix, refactoring, ....

git-examples.md
```diff
@@ -0,0 +1,63 @@
...
+```
+* Make the changes that you want to add
+* Create a new branch with a meaningful name that represents your feature<br>
+```sh
+git checkout -b <your branch name>
+```
+you can use the command ```git branch``` to make sure you're on the right branch
+* Add your changes that you want to be part of the commit<br>
+```sh
+git add .
```
bkueng: I would not suggest that. It usually does not end up in a good commit (as you probably end up having unrelated changes in one commit). Using a gui (like git gui [0]), or `git add -p` [1] would IMO be prefered. This way you review your own changes one more time.

[0] https://git-scm.com/book/en/v2/Git-in-Other-Environments-Graphical-Interfaces
[1] http://nuclearsquid.com/writings/git-add/

git-examples.md
```diff
@@ -0,0 +1,63 @@
...
+```sh
+git rebase master
+```
+* Now you can push your local commits to your fork<br>
+```sh
+git push origin <your branch name>
+```
+* You can verify that the push was successful by going to your forked repository in your browser: ```https://github.com/<your git name>/Firmware.git```<br>
+There you should see the message that a new branch has been pushed to your fork.
+* Now it's time to create a pull request (PR). On the right hand side of the "new branch message" (see one step before), you should see a green button saying "Compare & Create Pull Request". Then it should list your changes and you can (must) add a meaningful title (short) and message (explain what you did for what reason. Check [other pull requests](https://github.com/PX4/Firmware/pulls) for comparison)
```
bkueng: 'explain what you did for what reason'. That's quite important, and I see it often not being done for new contributors, so we should highlight that.
You can also add that the commit message can just be copied if it's a single commit.

bkueng (commented):
I made some comments.

I think more experienced git users can safely skip this. They just need to know that we're using a rebasing workflow. You might want to mention that in the beginning.

Stifael:
@kd0aij. Maybe px4 submodule is the wrong name. I was talking about a submodule used in the firmware, such as the sitl gazebo one

git-examples.md
```diff
@@ -72,15 +72,15 @@ There you should see the message that a new branch has been pushed to your forke
...
 ```sh
 git checkout -b feature_branch
 ```
-* Go go submodule subdirectory
+* Go to submodule subdirectory
 ```sh
 cd <path to submodule>
 ```
-* PX4 submodule might not necessarilty point to the newest commit. Therefore, first checkout out master and pull the newest upstream code.
+* PX4 submodule might not necessarily point to the newest commit. Therefore, first checkout master and pull the newest upstream code.
```
kd0aij, line 79: I don't understand this; does "git pull" update submodule commit pointers, or do you need to use a "git submodule xxx" command?

dagar: Usually with submodules the answer is whatever's most inconvenient. I end up running `make submodulesclean` all the time jumping between branches.

Stifael: I am sorry. the steps that I wrote have to be done if you like to have a submodule update in the Firmware.
But anyway, the way I understand it (and I will add it to dev guide afterwards) is as follow (try to explain it with a simple example):
Assume we are working in the Firmware. At some point we realize that there is a bug in the submodule sitl_gazebo. In order to fix it, one approach is to go to the sitl_gazebo directory and do a pull such that we have the newest commits. Then we fix the bug, make a new branch that describes the bug, then do an add and a commit, and finally push it to the submodule origin. Then we go online and do a PR for that bug_fix. 
Once the bug_fix is in the submdolue master, you need to do a PR for submodule update. To do so you have to follow the steps I wrote in the guide. I will update it now

PR #83, opened 2017-01-20
rename mavlink messaging and add overview link + stream rate

ChristophTobler:
I renamed and restructured the mavlink page a bit. Added the link to the overview and added an example of how to change streaming rate


ChristophTobler:
@bkueng Could you please review?

bkueng:
Yep, looks good!

PR #87, opened 2017-02-04
small updates, mostly spelling

dagar:


dagar:
hmm, I used the gitbook editor to make the change and it looks like it altered some white space as well. I'll review/cleanup.

dagar:
Gitbook editor whitespace changes seem to be harmless or for the better.

Please review.

LorenzMeier:
I generally prefer their web editor because the integration is a lot easier.

bkueng (approved):

LorenzMeier:
Github sync from Gitbook is broken again. It might be worthwhile to try to use the web editor from Gitbook until we figured that one out.

dagar:
I think the only solution is to force the sync one way or the other and then manually restore anything that was lost.

PR #91, opened 2017-03-02
heap profiling: Add instruction for Ubuntu

simonegu:


advanced-gdb-debugging.md
```diff
@@ -28,27 +28,46 @@ Stack usage is calculated with stack coloring and thus is not the current usage,
...
+##### Ubuntu:
+```bash
+sudo apt-get install google-perftools libgoogle-perftools-dev
+```
+
+#### Start heap profiling
+
+First of all, build the firmware as follows:
+```bash
+build_posix_sitl_default
```
bkueng: shouldn't it be `make posix_sitl_default`?

advanced-gdb-debugging.md
```diff
@@ -28,27 +28,46 @@ Stack usage is calculated with stack coloring and thus is not the current usage,
...
+
+In another terminal, type:
 ```bash
 cd build_posix_sitl_default/tmp
 export HEAPPROFILE=/tmp/heapprofile.hprof
-env LD_PRELOAD=/lib64/libtcmalloc.so ../src/firmware/posix/px4 posix-configs/SITL/init/lpe/iris
+export HEAP_PROFILE_TIME_INTERVAL=30
+```
+
+And entry this depending on your system:
```
bkueng: *enter*

advanced-gdb-debugging.md
```diff
@@ -28,27 +28,46 @@ Stack usage is calculated with stack coloring and thus is not the current usage,
...
+##### Ubuntu:
+```bash
+env LD_PRELOAD=/usr/lib/libtcmalloc.so ../src/firmware/posix/px4 ../../posix-configs/SITL/init/lpe/iris
+google-pprof --pdf ../src/firmware/posix/px4 /tmp/heapprofile.hprof.0001.heap > heap.pdf
+```
+
 It will generate a pdf with a graph of the heap allocations.
 The numbers in the graph will all be zero, because they are in MB. Just look at the percentages instead. They show the live memory (of the node and the subtree), meaning the memory that was still in use at the end.
 
 If it does not generate heap dumps while running the `px4` app you might need to change the settings of the profiler. On some systems it is necessary to set an interval time when to write the dumps:
```
bkueng: this line can be removed now

bkueng (commented):
Thanks for the update

ndepal:
I noticed a line (which you did not change in this commit) that does not make a lot of sense:

> Then advertise the debug value topic (one advertisement for different published names is sufficient). Put this in front of your main loop:

The next line is a new title. It looks like something is missing here.

bkueng:
@ndepal I remove the text, it was duplicated from some other place.

PR #95, opened 2017-03-16
replay: document ekf2 replay with ulog

bkueng:
@priseborough


priseborough:
The text explains how to extract the parameters to a file replay_params.txt, but doesn't explain that the parameters in  replay_params.txt are then adjusted before repeating the replay when assessing the effect of parameter changes.

bkueng:
I extended it.

priseborough (approved):

PR #97, opened 2017-03-21
Update Intel Aero RTF instructions

lucasdemarchi:
Not sure how this can be "compiled" to check it's rendering properly, but this should be the update to the instructions for the current version of the BSP.


lucasdemarchi:
@LorenzMeier 

I think we also need to change the picture on this page. I'm not comfortable with doing it without having a way to visualize the change though.

lucasdemarchi:
Fixed the picture.

PR #102, opened 2017-03-26
Include Aero in README

LorenzMeier:


LorenzMeier:
This is mostly here to test the Gitbook integration.

PR #106, opened 2017-04-02
Added a note to use the LPE version of the firmware

kPanesar:
Currently, there is no way for a developer to know that the parameters mentioned on the "External Position Estimation" page are not enabled by the default firmware build. A note was added at the top of the file to use the LPE version of the PX4 firmware.


LorenzMeier:
Nice, thanks!

PR #113, opened 2017-04-10
Hgw add parameters file manually

hamishwillee:
This adds Firmware parameters to the devguide (an updated version of those here: https://pixhawk.org/firmware/parameters ). As per https://github.com/PX4/Firmware/issues/6913 the plan is that eventually this updated will happen automatically.


dagar:
Are you going to include parameters.md in both the user guide and the dev guide?
I'll create an issue for the auto update.

hamishwillee:
> Are you going to include parameters.md in both the user guide and the dev guide?

That is a very good question. I hadn't planned to, but actually it makes quite a lot of sense for *users* to have this information handy (even though is for more "advanced" users). 

- What do you think - useful for people constructing their own drones? Doing tuning.
- Will it be possible to send it to both places?

dagar:
Yes, we can do both easily. Just show me where we need to commit the output and let's figure out how to get a Github PX4 jenkins user with appropriate permissions.

hamishwillee:
As per other thread, this should go in DevGuide replacing my manually created file: https://dev.px4.io/parameter_reference.html. In user guide it will go under advanced configuration: https://hamishwillee.gitbooks.io/px4-user-guide/content/en/advanced_config/

In both cases, perhaps wait a week for multilingual-system work to take place?

PR #116, opened 2017-04-10
Refactor: clean up image locations

bkueng:


hamishwillee:
This is excellent. It is a precondition for multi-lingual installation that images are (at least by default) outside of the docs tree. That allows all translations to use the common image, but create their own image tree or have local images for translated pictures.

Issue #92, opened 2017-03-07
Richquotes - notes, warnings, tips etc currently not working


hamishwillee:
Plugins for heading anchors and richquotes are in conflict: 
https://github.com/erixtekila/gitbook-plugin-richquotes/issues/25

This is just for tracking.

hamishwillee:
@bkueng I also note that in [some places](https://raw.githubusercontent.com/PX4/Devguide/master/hardware-snapdragon.md) we use explicit div markup like this
```
<aside class="warning">
Although the Snapdragon uses DF13 connectors, the pinout is different from Pixhawk.
</aside>
```
Can I assume that this was the old way of doing asides, and now the intention is to use richquotes plugin?

bkueng:
> Can I assume that this was the old way of doing asides, and now the intention is to use richquotes plugin?

yes

hamishwillee:
Thanks @bkueng - I've created #117 to fix up all the incorrect cases. Merge if they look OK!

hamishwillee:
@bkueng I think we should disable the heading anchors plugin. This will enable the notes, tips etc to work. 

While it would be nice to have the little link showing the anchor (for easy copying) I think that is not important (note that the actual anchor is still created by the page toc plugin)

What do you think?

bkueng:
I agree it would be nice to have it, but if you can still get the link via TOC, then yeah, let's disable it.

hamishwillee:
OK, done. We should retest anchors in future though. It is a cool plugin.

hamishwillee:
Fixed by https://github.com/PX4/Devguide/pull/680https://github.com/PX4/Devguide/pull/680

PR #117, opened 2017-04-11
Change all use off old-style aside markup to use richquotes plugin

hamishwillee:
This changes all files to use the [richquotes syntax](https://www.npmjs.com/package/gitbook-plugin-richquotes). Even though richquotes currently have a bug #92 , this makes it easier for quote markup to show up, as it renders as quote by default.


LorenzMeier:
Thanks!

PR #119, opened 2017-04-18
Add Airframes Reference Page

bkueng:
@hamishwillee you can add this to the user guide as well, where you see it fit.


hamishwillee:
> @hamishwillee you can add this to the user guide as well, where you see it fit.

Will do. Thanks.

hamishwillee:
For Quadrotor Wide all of the layouts have these as "specific". Since they are the same, shouldn't they be "common"?

MAIN1: motor 1
MAIN2: motor 2
MAIN3: motor 3
MAIN4: motor 4

hamishwillee:
Another possible layout that I like is: 
image

hamishwillee:
OK, merged it in: https://docs.px4.io/en/airframe_reference.html. Had to put it in the root of the installation because otherwise all the image references broke. I wonder if we could make the script dynamically take the path?

Anyway, last question,. The links to go build logs on pixhawk.org (e.g. https://pixhawk.org/platforms/planes/z-84_wing_wing). Why does Pixhawk.org host build logs? I will be moving airframes docs to user guide from dev guide, but we should presumably remove the pixhawk.org version too and have just "one place"?

bkueng:
> For Quadrotor Wide all of the layouts have these as "specific". Since they are the same, shouldn't they be "common"?

Only the first 4 are the same, some of them have additional feed-through (and I double checked, they use different mixer files). To avoid confusion I always group all MAIN outputs together.

> Another possible layout that I like is:

Looks nice as well, yes. With the current, the separation between common header and airframes is clearer. I'm ok with either one of them.

> I wonder if we could make the script dynamically take the path?

Sure. Do you want to make a PR?


> Anyway, last question,. The links to go build logs on pixhawk.org (e.g. https://pixhawk.org/platforms/planes/z-84_wing_wing). Why does Pixhawk.org host build logs? I will be moving airframes docs to user guide from dev guide, but we should presumably remove the pixhawk.org version too and have just "one place"?

I agree. pixhawk.org is the predecessor of dev.px4.io (and I'm not sure if it also served as user-guide). I agree we should move it. Some information there is just outdated and some of it does not yet exist on the dev guide (the parameter reference was one of these).

I'm merging this then.

PR #120, opened 2017-04-19
Add plugin to enable favicon for docs

hamishwillee:
The favicon.ico file is already present (not very high res - we probably should get a better one). All this does is use the [custom-favicon](https://www.npmjs.com/package/gitbook-plugin-custom-favicon) plugin to ensure it is displayed (basically it copies a specified file to replace the default theme favicon in the output).

### DIFF
--- a/book.json
+++ b/book.json
@@ -8,7 +8,8 @@
         "-mermaid-2",
         "github",
         "language-picker",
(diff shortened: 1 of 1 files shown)


hamishwillee:
@bkueng FYI, added this :-)

PR #122, opened 2017-04-20
update Mount (Gimbal) page

bkueng:
@julianoes @mhkabir

### DIFF
--- a/en/advanced/gimbal_control.md
+++ b/en/advanced/gimbal_control.md
@@ -24,9 +24,14 @@ motion is required.
 
 ## Configure the gimbal mixer for AUX output
 
-If the `MNT_MODE_OUT` parameter is set to `AUX`, a mixer file has to be configured.
-The gimbal uses the control group #2 (see [Mixing and Actuators](../concept/mixing.md)).
-This is the basic mixer configuration:
+The gimbal configuration must be specified in a [main or auxiliary mixer file](../advanced/system_startup.md#starting-a-custom-mixer)
+(a mixer file has to be configured if the `MNT_MODE_OUT` parameter is set to
+`AUX`).
+
+> **Note** Read [Mixing and Actuators](../concept/mixing.md) for an explanation
+> of how mixers work and the format of the mixer file.
+
+A basic basic mixer configuration for a mount is shown below.
 
 ```
 # roll
@@ -45,7 +50,9 @@ O:      10000  10000      0 -10000  10000
 S: 2 2  10000  10000      0 -10000  10000
 ```
 
-Add those you need to your [main or auxiliary mixer](../advanced/system_startup.md#starting-a-custom-mixer).
+The gimbal uses the control group #2 (see [Mixing and Actuators](../concept/mixing.md)).
+Add the configuration values you need (depending on the gimbal hardware and
+selected output pins) to your own mixer.
 
 There is also a [generic quad airframe config](../airframes/airframe_reference.md#quadrotor-x)
 which includes a [mount mixer](https://github.com/PX4/Firmware/blob/master/ROMFS/px4fmu_common/mixers/mount.aux.mix).
--- a/en/advanced/gimbal_control.md
+++ b/en/advanced/gimbal_control.md
@@ -1,20 +1,24 @@
 # Gimbal Control Setup
 
-PX4 contains a generic mount/gimbal control driver with different input and output methods. Any input can be selected to drive any output.
-
-First, make sure the driver runs, using `vmount start`, then configure its parameters.
+PX4 contains a generic mount/gimbal control driver with different input and
+output methods. Any input method can be selected to drive any output.
 
 ## Parameters
-The parameters are described in [src/drivers/vmount/vmount_params.c](https://github.com/PX4/Firmware/blob/master/src/drivers/vmount/vmount_params.c). The most important ones are the input (`MNT_MODE_IN`) and the output (`MNT_MODE_OUT`) mode. By default, the input is disabled. Any input method can be selected to drive any of the available outputs.
+The parameters are described [here](../advanced/parameter_reference.md#mount). The most
+important ones are the input (`MNT_MODE_IN`) and the output (`MNT_MODE_OUT`)
+mode. By default, the input is disabled. After selecting the input mode, reboot
+the vehicle so that the mount driver starts. Any input method can be selected to
+drive any of the available outputs.
 
-If a mavlink input mode is selected, manual RC input can be enabled in
-addition (`MNT_MAN_CONTROL`). It is active as long as no mavlink message is received yet, or mavlink explicitly requests RC mode.
+If the input mode is set to `AUTO`, the mode will automatically be
+switched based on the latest input. To switch from mavlink to RC, a large stick
+motion is required.
 
+## Configure the gimbal mixer for AUX output
 
-
-### Configure the gimbal mixer for AUX output
-
-The gimbal uses the control group #2 (see [Mixing and Actuators](../concept/mixing.md)). This is the mixer configuration:
+If the `MNT_MODE_OUT` parameter is set to `AUX`, a mixer has to be configured.
+The gimbal uses the control group #2 (see [Mixing and Actuators](../concept/mixing.md)).
+This is the basic mixer configuration:
 
 ```
 # roll
@@ -33,7 +37,23 @@ O:      10000  10000      0 -10000  10000
 S: 2 2  10000  10000      0 -10000  10000
 ```
 
-Add those you need to your main or auxiliary mixer.
+Add those you need to your [main or auxiliary mixer](../advanced/system_startup.md#starting-a-custom-mixer).
+
+There is also a generic quad airframe config which includes a mount mixer:
+[Generic Quadrotor X config with mount](../airframes/airframe_reference.md#quadrotor-x).
+
+
+## SITL
+
+The Typhoon H480 model comes with a preconfigured simulated gimbal. To run it,
+use:
+```
+make posix gazebo_typhoon_h480
+```
+
+To just test the mount driver on other models or simulators, make sure the
+driver runs, using `vmount start`, then configure its parameters.
+
 
 ## Testing
 The driver provides a simple test command - it needs to be stopped first with `vmount stop`. The following describes testing in SITL, but the commands also work on a real device.
--- a/en/advanced/gimbal_control.md
+++ b/en/advanced/gimbal_control.md
@@ -22,14 +22,27 @@ If the input mode is set to `AUTO`, the mode will automatically be
 switched based on the latest input. To switch from mavlink to RC, a large stick
 motion is required.
 
-## Configure the gimbal mixer for AUX output
+## AUX output
 
-The gimbal configuration must be specified in a [main or auxiliary mixer file](../advanced/system_startup.md#starting-a-custom-mixer)
-(a mixer file has to be configured if the `MNT_MODE_OUT` parameter is set to
-`AUX`).
+If the output mode is set to `AUX`, a mixer file is required to define the
+mapping for the output pins and the
+[mount mixer](https://github.com/PX4/Firmware/blob/master/ROMFS/px4fmu_common/mixers/mount.aux.mix)
+is automatically selected (overriding any aux mixer provided by the
+airframe configuration).
 
+The output assignment is as following:
+- **AUX1**: Pitch
+- **AUX2**: Roll
+- **AUX3**: Yaw
+- **AUX4**: Shutter/retract
+
+### Customizing the mixer configuration
 > **Note** Read [Mixing and Actuators](../concept/mixing.md) for an explanation
 > of how mixers work and the format of the mixer file.
+
+The outputs can be customized by [creating a mixer
+file](../advanced/system_startup.md#starting-a-custom-mixer) on the SD card with
+name `etc/mixers/mount.aux.mix`.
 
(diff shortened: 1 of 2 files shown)


mhkabir:
Looks good. 👍

en/advanced-gimbal-control.md
```diff
@@ -1,20 +1,24 @@
...
+output methods. Any input method can be selected to drive any output.
 
 ## Parameters
-The parameters are described in [src/drivers/vmount/vmount_params.c](https://github.com/PX4/Firmware/blob/master/src/drivers/vmount/vmount_params.c). The most important ones are the input (`MNT_MODE_IN`) and the output (`MNT_MODE_OUT`) mode. By default, the input is disabled. Any input method can be selected to drive any of the available outputs.
-
-If a mavlink input mode is selected, manual RC input can be enabled in
-addition (`MNT_MAN_CONTROL`). It is active as long as no mavlink message is received yet, or mavlink explicitly requests RC mode.
-
+The parameters are described [here](parameter_reference.md#mount). The most
+important ones are the input (`MNT_MODE_IN`) and the output (`MNT_MODE_OUT`)
```
hamishwillee: FYI You can actually link specifically to a reference for a particular parameter using its name - e.g. parameter_reference.md#MNT_MODE_IN. The parameter file will probably move shortly, but still worth defining a link and I'll move that too. Up to you.

en/advanced-gimbal-control.md
```diff
@@ -1,20 +1,24 @@
...
+PX4 contains a generic mount/gimbal control driver with different input and
+output methods. Any input method can be selected to drive any output.
 
 ## Parameters
-The parameters are described in [src/drivers/vmount/vmount_params.c](https://github.com/PX4/Firmware/blob/master/src/drivers/vmount/vmount_params.c). The most important ones are the input (`MNT_MODE_IN`) and the output (`MNT_MODE_OUT`) mode. By default, the input is disabled. Any input method can be selected to drive any of the available outputs.
-
-If a mavlink input mode is selected, manual RC input can be enabled in
-addition (`MNT_MAN_CONTROL`). It is active as long as no mavlink message is received yet, or mavlink explicitly requests RC mode.
-
+The parameters are described [here](parameter_reference.md#mount). The most
```
hamishwillee: FMI Is the group id automatically (#mount) created now? Awesome

en/advanced-gimbal-control.md
```diff
@@ -33,7 +37,23 @@ O:      10000  10000      0 -10000  10000
 S: 2 2  10000  10000      0 -10000  10000
 ```
 
-Add those you need to your main or auxiliary mixer.
+Add those you need to your [main or auxiliary mixer](advanced-system-startup.md#starting-a-custom-mixer).
+
+There is also a generic quad airframe config which includes a mount mixer:
```
hamishwillee: I don't yet understand where and how mixers work, but it seems to me that it would be better to somehow link to where the "Generic Quadrotor X config with mount" actually defines the mixer (perhaps in github tree) than to the definition of the airframe.

en/advanced-gimbal-control.md
```diff
@@ -1,20 +1,24 @@
 # Gimbal Control Setup
 
-PX4 contains a generic mount/gimbal control driver with different input and output methods. Any input can be selected to drive any output.
-
-First, make sure the driver runs, using `vmount start`, then configure its parameters.
+PX4 contains a generic mount/gimbal control driver with different input and
```
hamishwillee: What is the purpose of a gimbal control driver? Ie it is often worth saying something along the line of what you can control here - i.e. When you would need to do this setup and what config is reasonable - e.g. I assume you're doing this because you have a new gimbal and the sort of thing you want to define is what axis it can work with? Or is this more a tuning thing. Does that make sense?

en/advanced-gimbal-control.md
```diff
@@ -1,20 +1,24 @@
...
+output methods. Any input method can be selected to drive any output.
 
 ## Parameters
-The parameters are described in [src/drivers/vmount/vmount_params.c](https://github.com/PX4/Firmware/blob/master/src/drivers/vmount/vmount_params.c). The most important ones are the input (`MNT_MODE_IN`) and the output (`MNT_MODE_OUT`) mode. By default, the input is disabled. Any input method can be selected to drive any of the available outputs.
```
hamishwillee: So it looks like you need to set MNT_MODE_IN because it is disabled by default and then reboot. Why? You don't say. Does this mean that for any gimbal use you need to set MNT_MOUNT_IN? Similarly MNT_MOUNT_OUT, why is that needed. 

Also we've jumped into parameters here, and then we go to "Configure the gimbal mixer for AUX output". However there is no "context" at the point you talk about parameters. Before you talk about params and config for aux output, it is worth a sentence saying "how everything fits together". Something like "The input/output mixer is configured using a specific file XXXX, using parameters to define the inputs and outputs blah blah blah".

en/advanced-gimbal-control.md
```diff
@@ -1,20 +1,24 @@
...
+important ones are the input (`MNT_MODE_IN`) and the output (`MNT_MODE_OUT`)
+mode. By default, the input is disabled. After selecting the input mode, reboot
+the vehicle so that the mount driver starts. Any input method can be selected to
+drive any of the available outputs.
 
+If the input mode is set to `AUTO`, the mode will automatically be
+switched based on the latest input. To switch from mavlink to RC, a large stick
+motion is required.
 
-### Configure the gimbal mixer for AUX output
```
hamishwillee: So what does it mean to configure it for AUX output? Is there any other way you can configure the mixer? Why woudl you do this.

hamishwillee:
I've added some review comments. This is clearly and improvement. This is probably great for someone who already knows what is going on, but IMO could use a little more context before jumping in to parameters.

bkueng:
@hamishwillee I expanded the first sections. It's true, it did assume some deeper understanding already, so it's good to have someone less familiar reviewing it.

en/advanced/gimbal_control.md
```diff
@@ -33,7 +45,23 @@ O:      10000  10000      0 -10000  10000
 S: 2 2  10000  10000      0 -10000  10000
 ```
 
-Add those you need to your main or auxiliary mixer.
+Add those you need to your [main or auxiliary mixer](../advanced/system_startup.md#starting-a-custom-mixer).
+
+There is also a generic quad airframe config which includes a [mount mixer](https://github.com/PX4/Firmware/blob/master/ROMFS/px4fmu_common/mixers/mount.aux.mix):
```
hamishwillee: Change this to:

There is also a [generic quad airframe config](../airframes/airframe_reference.md#quadrotor-x) which includes a [mount mixer](https://github.com/PX4/Firmware/blob/master/ROMFS/px4fmu_common/mixers/mount.aux.mix).

en/advanced/system_startup.md
```diff
@@ -47,3 +47,16 @@ The `extras.txt` can be used to start additional applications after the main sys
 ### Starting a custom mixer
 
 By default the system loads the mixer from `/etc/mixers`. If a file with the same name exists in `/fs/microsd/etc/mixers` this file will be loaded instead. This allows to customize the mixer file without the need to recompile the Firmware.
+#### Example
+The following example shows how to add a custom aux mixer: create a file on the
```
hamishwillee: Make this a bullet list. BTW, this whole update is really useful.

en/advanced/gimbal_control.md
```diff
@@ -1,20 +1,32 @@
...
+switched based on the latest input. To switch from mavlink to RC, a large stick
+motion is required.
 
+## Configure the gimbal mixer for AUX output
 
-
-### Configure the gimbal mixer for AUX output
-
-The gimbal uses the control group #2 (see [Mixing and Actuators](../concept/mixing.md)). This is the mixer configuration:
+If the `MNT_MODE_OUT` parameter is set to `AUX`, a mixer file has to be configured.
```
hamishwillee: Best to start with "what they do" and where first (ie "create a file"). Also to state that they need to read the mixing docs to understand the file formats. So perhaps something like:

---

The gimbal configuration must be specified in a [main or auxiliary mixer file](../advanced/system_startup.md#starting-a-custom-mixer) (a mixer file has to be configured if the MNT_MODE_OUT parameter is set to AUX). 

> **Note** Read [Mixing and Actuators](../concept/mixing.md) for an explanation of how mixers work and the format of the mixer file.

A basic basic mixer configuration is shown below. 
 ```
 Your example
  ```
The gimbal uses the control group #2 (see [Mixing and Actuators](../concept/mixing.md)). Add the configuration values you need to your own mixer.

hamishwillee: The further question is how do I work out what to do to "Add those you need"? Or is this more like .. "Add the configuration you need for your selected input message formats and output hardware ..."

hamishwillee:
Heaps better @bkueng! I've put a few more suggestions inline.

bkueng:
@hamishwillee I updated the sentence. Good to go?

hamishwillee:
Well, except for the other comments. Sorry, I thought I'd submitted them.

en/advanced/gimbal_control.md
bkueng: My knowledge here is actually only theoretical, as I have never set this up myself on real HW :)

I addressed your comments, see new commits.

en/advanced/system_startup.md
```diff
@@ -47,3 +47,17 @@ The `extras.txt` can be used to start additional applications after the main sys
 ### Starting a custom mixer
 
 By default the system loads the mixer from `/etc/mixers`. If a file with the same name exists in `/fs/microsd/etc/mixers` this file will be loaded instead. This allows to customize the mixer file without the need to recompile the Firmware.
+#### Example
+The following example shows how to add a custom aux mixer:
+  * Create a file on the SD card, `etc/mixers/gimbal.aux.mix` with your mixer
+	content.
+  * Then to use it, create an additional file `etc/config.txt` with this content:
+```
```
hamishwillee: Just FYI, you can indent this to the level of "Then to ..." above to put this within the preceding bullet point.

hamishwillee:
@bkueng This is clearly and improvement and IMO good enough to merge. I vote we do so. However it would be good to get further review from someone who is actually trying this and can comment on what was useful/missed/confusing. Any thoughts on someone to ask?

bkueng:
Hmm, no. @LorenzMeier maybe?

en/advanced/system_startup.md
LorenzMeier: @bkueng We should condition this on a parameter that you can enable to output a gimbal on the aux ports by default. The whole microSD approach is horrible from a user perspective.

I would like to automate this, move the bulk of these docs to the user guide and then only document how the gimbal driver works in the dev guide.

This is clearly unusable for an average user (and even for an average developer) as-is. It is way too complex.

LorenzMeier (changes requested):

LorenzMeier:
@bkueng We should condition this on a parameter that you can enable to output a gimbal on the aux ports by default. The whole microSD approach is horrible from a user perspective.

I would like to automate this, move the bulk of these docs to the user guide and then only document how the gimbal driver works in the dev guide.

This is clearly unusable for an average user (and even for an average developer) as-is. It is way too complex.

hamishwillee:
@LorenzMeier How fast is that sort of change likely to be? It sounds great, but if due in more than a few weeks I'd still merge this on the basis that it is better than what we had.

LorenzMeier:
This is doable today. Just moving the contents on the card into the startup script.

en/advanced/gimbal_control.md
```diff
@@ -1,20 +1,37 @@
 # Gimbal Control Setup
 
-PX4 contains a generic mount/gimbal control driver with different input and output methods. Any input can be selected to drive any output.
+If you want to control a gimbal with a camera (or any other payload) attached to
+the vehicle, you need to configure how you want to control it and how PX4 can
```
mhkabir, line 4: The formatting seems all wrong - I just noticed. 

None of this should be on newlines. It makes it look super-weird once published. Each  _paragraph_ should be separated by a newline, but not sentences or in-between sentences.

mhkabir (changes requested):

en/advanced/gimbal_control.md
bkueng: That's actually not the case. A paragraph is defined by an **empty** newline. I have my editor set to a line width of 80, it reduces the diff in general.

bkueng:
@LorenzMeier Ok I'll look into it.

en/advanced/gimbal_control.md
mhkabir: My point is that the line width of 80 is fine when editing offline and for text editor purposes, but it doesn't look good when published. I feel that a paragraph should be 1 continuous _line_, such that it flows properly in a browser.

bkueng: Ok, so I didn't make it clear enough: this will be reformated when published, as specified by the markdown language, see https://toolchain.gitbook.com/syntax/markdown.html, section **Paragraphs and Line Breaks**.

mhkabir: Ah - thanks for the clarification. I understand now :)

bkueng:
I updated the docs according to the latest FW changes (https://github.com/PX4/Firmware/pull/7117): the mixer file is now automatically selected, but can still be customized.

bkueng:
Related: https://github.com/PX4/Firmware/pull/7124

PR #124, opened 2017-04-21
Reorganise docs into folders

hamishwillee:
This reorganises docs into folders rather than completely flat structure. Several docs that were orphans have been added. All images links and doc links have been fixed.


hamishwillee:
This should be good to go @LorenzMeier @bkueng

bkueng:
This looks good. There are some problems:
I found the following 2 broken links:
```
uavcan/README.md
32:Next, follow the instructions in [UAVCAN Configuration](/uavcan-node-enumeration.md) to activate the UAVCAN functionalities in the firmware. Disconnect your power supply and reconnect it. After the power cycle all UAVCAN devices should be detected which is confirmed by a beeping motor on the Orel 20 ESCs. You can now continue with the general setup and calibration.
34:Depending on the used hardware, it can be reasonable to perform an update of the firmware on the UAVCAN devices. This can be done via the UAVCAN itself and the PX4 firmware. For more details please refer to the instructions in [UAVCAN Firmware](/uavcan-node-firmware.md).
```

The links in the top-level readme (visible on github) are broken too (with the language separation update).

The links on the Introduction page (`en/README.md`) are incorrect: they don't need to start with `../`

hamishwillee:
Hi @bkueng 

Thanks so much - really nice to have thorough checking. Normally I use the plugin "validate-links" to check for links but unfortunately I can't get it to work for this book.

Anyway, fixed those ones plus some other links I found (direct links to dev.px4.io rather than using .md links). Hopefully good to go now!

PR #127, opened 2017-04-25
try translated_commit:

FantasyJXF:
Should I add the following code to every file?
```
---
translated_commit: https://github.com/PX4/px4_user_guide/blob/e9d8ee6fc72c9d0111c14edaf7c585c0024382aa/book.json 
---
```

PR #128, opened 2017-04-25
updated the Chinese README

FantasyJXF:
Please check it


hamishwillee:
HI @FantasyJXF 

I will discuss with @bkueng and @LorenzMeier if there are any other gating requirements, but IMO this looks really good. I will add any additional tracking information as a "post process".

Does this PR replace your previous PR #127 ? 

For your information you do not need to submit a new PR for every change. What you could have done is committed changes to your branch, and these would show up in the old PR until the branch was merged. You only need a new PR if the old PR is too broken to add to. Hope that makes sense.

FantasyJXF:
My bad, I didn't think too much, maybe you should close the other PR and this one is more complete since its the leastest one.

Next time I will remember that.

hamishwillee:
So this looks OK - ie that you have committed on top of just one PR rather than opening a separate one (there is a time and a place for multiple separate PRs but usually all "related" changes should go in just one)

There are a couple of different types of ways of merging. The preferred way for merging with master is to use what is called a rebase. What this does is pull in the current master and then re-run all your changes on top of the new master - from a git point of view it looks like you started from the current master. This is better than merging different branches. You might want to look this up.

Regards
Hamish

hamishwillee:
@LorenzMeier Can you confirm there are no gating factors for me to merge the first Chinese translation? @bkueng is OK with this.

hamishwillee:
Closing - superceded by #131

PR #141, opened 2017-05-05
add preliminary modules documentation

bkueng:
This adds the current results from https://github.com/PX4/Firmware/pull/7145 to the dev-guide. I split it into one page per category.
@hamishwillee


LorenzMeier:
Awesome!

hamishwillee:
@bkueng I am so impressed by this. You should be cloned.

hamishwillee:
PS, and interestingly the Chinese team have already started to translate this ... https://github.com/PX4/Devguide/pull/142

PR #142, opened 2017-05-05
Translate PREFLIGHT CHECK

FantasyJXF:
Translate PREFLIGHT CHECK


zh/middleware/modules_communication.md
```diff
@@ -0,0 +1,117 @@
+# Modules Reference: Communication
```
hamishwillee, line 1: Shall I wait for this to be translated before I do a merge?

BTW, English version is auto generated from source but we won't do that for Chinese.

hamishwillee:
Looks great - please thanks song bao dong for me :-) Do you want me to wait until the module doc translation is complete before merging this ... or is merging sooner better for you?

zh/middleware/modules_main.md
```diff
@@ -0,0 +1,21 @@
+
+# Modules & Commands Reference
+The following pages document the PX4 modules, drivers and commands. They
+describe the provided functionality, high-level implementation overview and how
+to use the command-line interface.
+
+> **Note** **This is auto-generated from the source code** and contains the
```
hamishwillee, line 7: When you translate this bit, you should probably note that the "original" is translated from source and will be up to date, but that this translation may lag slightly.

zh/middleware/modules_communication.md
FantasyJXF: I've noticed that, though it would be better if you add a new English file and add it to the Chinese folders.

As for this translation, I rarely translate the `comment` part or the source code. So I just translated the rest of them

FantasyJXF:
I think it would be better if you could merge it sooner. 

Thanks.

hamishwillee:
Merged!

> I've noticed that, though it would be better if you add a new English file and add it to the Chinese folders.

It would be great, but I'm not doing all the changes and we can't mandate this for contributors. Better for the translators to track the changes.

That said, this is one of the things I hope to make easy to automate with translation support.

PR #145, opened 2017-05-10
Add sensor/topic debugging using listener via QGC

hamishwillee:
This topic adds:
- Link in uORB Messaging to explain that you can use this feature via QGC, even when connected over wireless link
- New topic on sensor/topic debugging with a little more detail. 

Incidentally, also add validate-links plugin and fixes to some links that were failing.


hamishwillee:
Changes will need to propagate to Chinese tree before this is merged. 

@LorenzMeier - In sidebar new topic is referred to as "Sensor/Topic Debugging". In page it has title  "Sensor/Topic Debugging using the Listener Command". Are these reasonable (do you think a normal user would find what they expect and need with these topics).

en/debug/sensor_uorb_topic_debugging.md
```diff
@@ -0,0 +1,17 @@
+# Sensor/Topic Debugging using the Listener Command
+
+The uORB is an asynchronous `publish()` / `subscribe()` messaging API used for
+inter-thread/inter-process communication. The `listener` command can be used from the *QGroundControl MAVLink Console* to inspect topic (message) values, including the current values published by sensors.
+
+> **Tip** This is a powerful debugging tool because it can be used even when QGC is connected over a wireless link (e.g. when the vehicle is flying).
+
+<span></span>
```
bkueng, line 8: Is this span needed?

en/debug/sensor_uorb_topic_debugging.md
```diff
@@ -0,0 +1,17 @@
+# Sensor/Topic Debugging using the Listener Command
+
+The uORB is an asynchronous `publish()` / `subscribe()` messaging API used for
+inter-thread/inter-process communication. The `listener` command can be used from the *QGroundControl MAVLink Console* to inspect topic (message) values, including the current values published by sensors.
```
bkueng, line 4: I'd probably start more general, that the `listener` command is available on the shell (you mention it at the bottom), and then go on to the QGC console (which is actually much more powerful as well, as you can monitor system load, start/stop tasks, etc. directly from QGC).

bkueng (commented):
I added 2 remarks, other than that it looks good.
The link validation will be very helpful.

en/debug/sensor_uorb_topic_debugging.md
hamishwillee: Yes. This allows two "notes" to be rendered following immediately after each other. Without this (or something similar) there will always be at least one full line/para between notes.

en/debug/sensor_uorb_topic_debugging.md
hamishwillee: @bkueng The text is deliberately designed this way (my original version was just as you have suggested!) The reason it exactly as you say - it is QGC that is interesting here - not the shell.

It is possible it is pushed "too far" down in which case this note could be in the set with the other ones.

Does that make sense?

As an aside, it would be great to talk more about debugging using QGC, but I know little about that yet. I have created https://github.com/mavlink/qgc-user-guide/issues/71 so we will eventually add more information in QGC on this topic.

bkueng: Makes sense. I think I'd still move the note up.
The QGC shell was very recently added. It's basically the same as the mavlink shell, it just can be a bit more comfortable if you want to use QGC at the same time. We can discuss this offline if you'd like.

en/debug/sensor_uorb_topic_debugging.md
bkueng: ok, nice trick.

en/debug/sensor_uorb_topic_debugging.md
hamishwillee: I'd like to discuss it as a separate topic/offline. I'll move up the note and merge for this case.

en/debug/sensor_uorb_topic_debugging.md
hamishwillee: It would want to be - took me ages to find a workaround :-)

hamishwillee:
@FantasyJXF FYI This change adds a new page. Matching pages were added to the Chinese tree.

FantasyJXF:
Thank you

PR #146, opened 2017-05-10
Update Aero naming in line with Intel branding

hamishwillee:
Intel requested we use proper branding for the aero: 

"For the underlying page title and body copy please again say “Intel® Aero Ready to Fly Drone”  Let’s please not abbreviate it to RTF in title or body copy.


hamishwillee:
@bkueng We should probably also update the airframes reference (can you add note to do that, non-urgently). Does this otherwise look OK to merge?

bkueng:
I'll update the airframe.

bkueng:
https://github.com/PX4/Devguide/commit/8a3fe7e31c9efc98b9e039c1af794274a4993f1a

hamishwillee:
Thanks very much.

PR #148, opened 2017-05-13
Update the Chinese Translations

FantasyJXF:
Update the middleware.Thank you Innoecho.


hamishwillee:
Looks great @FantasyJXF (and thanks to Innoecho. too!)

One thing to keep an eye on (not something I spotted here, but just for your information) is in-page anchors. If the English page links to an anchor (e.g. fred.html#some_heading_anchor_link) you will have to use a different anchor for Chinese content, since these are auto-generated by default based on heading text. It is possible to properly define anchors, but we don't use that functionality.

hamishwillee:
Actually, that may be the cause of this broken link ...

```
warn: link to "flight_controller/＃do-i-with-a-p1-or-p2-board" on page "flight_controller/snapdragon_flight_advanced.md" could not be resolved 
warn: link to "flight_controller/＃do-i-with-a-p1-or-p2-board" on page "flight_controller/snapdragon_flight_advanced.md" could not be resolved 
```

PR #157, opened 2017-05-20
Added a maintenance page with pick of churn tool

julianoes:
The amount of churn can be an indicator which files/parts might need
refactoring.


LorenzMeier:
Thanks!

PR #164, opened 2017-05-29
Instructions for required python packages

qiaoshun8888:
New installed Ubuntu doesn't come with these required packages.
It's the same as in [Mac OS instruction](https://dev.px4.io/en/setup/dev_env_mac.html).


mhkabir (approved):

bkueng:
There's already `python-serial` and `python-empy` on the `apt-get` list. Do they really need to be installed with pip too?

qiaoshun8888:
Updated.
Not really. I didn't notice that they are installed by `apt-get`. Just removed these two. Thanks.

bkueng:
Thanks

PR #165, opened 2017-05-30
update the linux installnation

FantasyJXF:
I checked the other translations and make fix sth .

thanks Li


hamishwillee:
Thanks very much. Merged.

PR #166, opened 2017-05-30
Add example of etc/extras.txt script and document potential boot failure

jlecoeur:
Adds documentation for the issue described in https://github.com/PX4/Firmware/issues/4846.


bkueng (approved):
Thanks!

PR #168, opened 2017-06-01
Update flight_modes.md

kd0aij:
correct name for multirotor manual angle control flight mode is "stabilize", not "angle"


dagar:
Thanks Mark.

PR #174, opened 2017-06-05
Adding Fast RTPS documentation

vicentemonge:


en/setup/fast-rtps-installation.md
```diff
@@ -0,0 +1,95 @@
...
+
+## [Requirements](http://eprosima-fast-rtps.readthedocs.io/en/latest/requirements.html#requirements)
+
+*eProsima Fast RTPS* requires the following packages to work.
+
+### Common Dependencies
+
+#### Gtest
+
+Gtest is needed to compile the tests when building from sources.
```
LorenzMeier: This is not clear enough. We require in our documentation installation instructions for the major platforms (e.g. apt-get install XYZ or brew install XYZ).

en/setup/fast-rtps-installation.md
```diff
@@ -0,0 +1,95 @@
...
+
+#### Java
+
+Java is required to make use of our buil-in code generation tool *fastrtpsgen*.
```
LorenzMeier: Not detailed enough. What version of Java? Is that version already installed with the rest of PX4? (we do have Java as dependency for jMAVSim)

LorenzMeier (changes requested):
Thanks, this is very helpful! To match the quality of the rest of the documentation we need a little more specific instructions in some areas.

en/setup/fast-rtps-installation.md
vicentemonge: We need 1.6 at least but jMAVSim needs 1.8. We will add in the document that we need 1.6 or newest.

mhkabir: It would make sense just to recommend 1.8. Too many options just confuse people.

vicentemonge: it seems good.

en/setup/fast-rtps-installation.md
vicentemonge: We think is not necessary since is for test purposes. We'll avoid that in this abstract.

LorenzMeier (approved):

PR #177, opened 2017-06-06
Navio2 update

nicolaerosia:


bkueng:
This makes sense. Note to self: we have support for RPi w/o shield, we should add a separate page for that.

nicolaerosia:
@bkueng that was the idea, we need to factor out common bits for Raspberry Pi into a separate page, and leave specific setup to different pages (raspberry_pi_navio, raspberry_pi_standalone, etc)

nicolaerosia:
@bkueng could you share details on the setup without shield?

bkueng:
It's in this PR: https://github.com/PX4/Firmware/pull/7200. Make target `posix_rpi_cross_no_shield` and this describes the setup: https://github.com/PX4/Firmware/blob/f16b5dfea157c3ef1a18e972479d9dcd9cf01087/posix-configs/rpi/px4_no_shield.config#L1:L5 (I haven't tried it myself however).

PR #178, opened 2017-06-07
Add dev guide for Airmind products.

flynxea:
Add dev guide for Airmind products.


airmind:
@dagar @LorenzMeier

mhkabir:
Nice!

dagar (approved):

airmind:
Thanks @dagar @mhkabir

hamishwillee:
Already merged, but here is some feedback/questions that might apply to an update. These might actually be for @dagar @mhkabir (e.g. where should files be located).

1. Is this a Pixhawk-series autopilot, and if so, which Pixhawk architecture is it compatible with.
2. There is no information on firmware update or airframe selection. Should at least comment on this, even if it is just to link to our other docs.
3. Should the vehicle pages actually be moved to airframes section? Normally airframes require some level of customisation and this is RTF, but still looks appropriate to me. 

In terms of content:
1. FYI only to @airmind The PDFs are actually pretty good, but could use a subedit for English
2. This doc could use a subedit for English and for style (e.g. "User guide" doesn't really need to be under a heading and in a "Note"). Lots of minor grammatical errors.
3. The big problem for me is the introduction, which I have pulled apart below:

>MindRacer is a fully stackable flight platform for miniature UAVs. 

* What does that mean? Stacking what on what? 
* I THINK you're referring to the fact that you have separate Flight controller, interface board, mrpdb boards, and that using these you can build a system that means you don't need to do any soldering, wiring. This whole stacking thing could use its own section explaining exactly what board does what, and how/whether they can be used together and independently. 

> Based on MindPX, MindRacer further scales down in formfactor 

scales down/compared to what? If you use scales you need to be comparative. Otherwise you can make statements about "smallest in class" or similar.

> while focused on providing modularity. 

See above. Is the thing small if you just use one part of the stack, or is the modulartiy a separate thing and if you use all the bits you get a big system.

> MindRacer is a platform rather than a flight controller. 

What do you mean by that? Where does the platform start and end - all the way to vehicle, or just at the "flight stack". Are there any limitations or hard codings implied by this platformisation? 

I guess what I'm saying is that when some people say platform they mean "something we expect you to take away and customise further".

> MindRacer implements the SEP(soldering-elimination-port) and WEP(wiring-elimination-protocol) concepts. Before SEP and WEP, soldering and wiring are always the major pain and efficiency killer during UAV manufacturing and tuning.

Good :-) Again, probably refer forward to information about how your modularisation works.

This looks like a pretty cool product!

dagar:
This is another board just like the pixhawk (px4fmu-v2) or pixhawk 3 (px4fmu-v4pro) that's fully supported in upstream PX4 (https://github.com/PX4/Firmware/tree/master/nuttx-configs). You can update it through QGC.

airmind:
@hamishwillee thanks a lot for the advise. will revise eventually @eflyinghi

mhkabir:
@dagar What do you think about adding a table on the build system page which lists all the CMake targets and links to the hardware-specific pages for each with a short description?

dagar:
Maybe list the make target within each board under the autopilot hardware section?

mhkabir:
Yeah that'd work too, but what I've seen is that people don't bother going to the hardware page.

hamishwillee:
@airmind YOu're welcome. I may well chase you up for some of this if/when I get time to start looking at Devguide improvements.

PR #184, opened 2017-06-15
Document Pixhawk 3 Pro

bkueng:


en/setup/building_px4.md
```diff
@@ -28,7 +28,18 @@ cd Firmware
...
 ```
 
-Note the syntax: 'make' is the build tool, 'px4fmu-v2' is the hardware / autopilot version and 'default' is the default configuration. All PX4 build targets follow this logic. A successful run will end with this output:
+Note the syntax: 'make' is the build tool, 'px4fmu-v2' is the hardware / autopilot version and 'default' is the default configuration. All PX4 build targets follow this logic.
+
+Build targets:
+* Crazyflie: `make crazyflie_default`
+* Intel® Aero RTF: `make aerofc-v1_default`
+* MindPX: `make mindpx-v2_default`
+* Pixhawk: `make px4fmu-v2_default`. You can use `make px4fmu-v3_default` for
```
mhkabir, line 37: We should put a note here that that the Pixhawk 1, 2 (the one in the Solo), 2.1 (from ProfiCNC) and Mini (from 3DR) are all supported using this build target.

bkueng: Sure, please go ahead and add them, I don't know all the boards.

PR #194, opened 2017-06-19
add astyle documenation

ChristophTobler:
I thought it would make sense to include that. Maybe general code style guidance would also help

### DIFF
--- a/en/contribute/README.md
+++ b/en/contribute/README.md
@@ -12,6 +12,17 @@ To contribute new functionality, [sign up for Github](https://help.github.com/ar
 
 All contributions have to be under the permissive [BSD 3-clause license](https://opensource.org/licenses/BSD-3-Clause) and all code must not impose any further constraints on the use.
 
+## Code style formatting
+
+PX4 uses [astyle](http://astyle.sourceforge.net/) for code formatting. Valid versions are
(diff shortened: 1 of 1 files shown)


bkueng (approved):
Looks good to me. @dagar are these versions ok?

dagar:
It's definitely possible to have differences between versions we can't control. Should we limit to the latest version only? I can update CI pretty easily.

ChristophTobler:
Sounds good to me. Should we get rid of https://github.com/PX4/astyle as it doesn't work at the moment and we don't have to maintain it?

dagar:
Yes let's get rid of it. We just need to be able to point Ubuntu users at an acceptable version (packaged).

ChristophTobler:
Do you think source code is not suitable? I can add instructions so that it is straight forward

LorenzMeier:
@ChristophTobler We should try to aim for a packaged version.

ChristophTobler:
Ok. I'll check which one works best

ChristophTobler:
Should we just create our own .deb file with e.g. checkinstall? Because for me only the source installation from https://sourceforge.net/projects/astyle worked. `apt-get install astyle` version produces a huge diff when running `make format`.

PR #201, opened 2017-06-22
RTK: cleanup & update

bkueng:
Goes together with user-guide update here: https://github.com/PX4/px4_user_guide/pull/30

@mhkabir  fyi


mhkabir:
This is great!

PR #204, opened 2017-06-27
Fix broken links from Korean build

hamishwillee:


hamishwillee:
@jeyong FYI, this fixes some broken links from the Korean section. It includes a new doc about Pixhawk 3 that will need translation.

PR #209, opened 2017-07-01
ROS: MAVROS Offboard control example: update SET MODE flag

TSC21:
This PR:
1. Changes some descriptions so it becomes more clear and updates them
2. Updates the `SET_MODE` cmd flag from `success` to `mode_sent`, according to https://github.com/mavlink/mavros/commit/599c5887ad2aadc124d7e1354f3a1597402118da.

So it becomes definite, this PR should probably wait for new MAVROS release on ROS buildfarm (as this change is only on current `master`).

### DIFF
--- a/en/ros/mavros_offboard.md
+++ b/en/ros/mavros_offboard.md
@@ -1,20 +1,20 @@
-# MAVROS offboard control example
+# MAVROS *Offboard* control example
 
-> **Caution** Offboard control is dangerous. If you are operating on a real vehicle be sure to have a way of gaining back manual control in case something goes wrong.
+> **Caution** *Offboard* control is dangerous. If you are operating on a real vehicle be sure to have a way of gaining back manual control in case something goes wrong.
 
-The following tutorial will run through the basics of offboard control through mavros as applied to an Iris quadcopter simulated in Gazebo. At the end of the tutorial, you should see the same behaviour as in the video below, i.e. a slow takeoff to an altitude of 2 meters.
+The following tutorial will run through the basics of *Offboard* control through MAVROS as applied to an Iris quadcopter simulated in Gazebo with SITL running. At the end of the tutorial, you should see the same behaviour as in the video below, i.e. a slow takeoff to an altitude of 2 meters.
 
 <video width="100%" autoplay="true" controls="true">
 	<source src="../../assets/sim/gazebo_offboard.webm" type="video/webm">
 </video>
 
 ## Code
-Create the offb_node.cpp file in your ros package and paste the following inside it:
+Create the `offb_node.cpp` file in your ROS package (by also adding it to you `CMakeList.txt` so it is compiled), and paste the following inside it:
 ```C++
 /**
  * @file offb_node.cpp
- * @brief offboard example node, written with mavros version 0.14.2, px4 flight
- * stack and tested in Gazebo SITL
+ * @brief Offboard control example node, written with MAVROS version 0.19.x, PX4 Pro Flight
+ * Stack and tested in Gazebo SITL
  */
 
 #include <ros/ros.h>
@@ -75,7 +75,7 @@ int main(int argc, char **argv)
         if( current_state.mode != "OFFBOARD" &&
             (ros::Time::now() - last_request > ros::Duration(5.0))){
             if( set_mode_client.call(offb_set_mode) &&
-                offb_set_mode.response.success){
+                offb_set_mode.response.mode_sent){
                 ROS_INFO("Offboard enabled");
             }
             last_request = ros::Time::now();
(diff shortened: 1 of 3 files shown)


en/ros/mavros_offboard.md
```diff
@@ -2,19 +2,19 @@
 
 > **Caution** Offboard control is dangerous. If you are operating on a real vehicle be sure to have a way of gaining back manual control in case something goes wrong.
 
-The following tutorial will run through the basics of offboard control through mavros as applied to an Iris quadcopter simulated in Gazebo. At the end of the tutorial, you should see the same behaviour as in the video below, i.e. a slow takeoff to an altitude of 2 meters.
+The following tutorial will run through the basics of `OFFBOARD` control through MAVROS as applied to an Iris quadcopter simulated in Gazebo with SITL running. At the end of the tutorial, you should see the same behaviour as in the video below, i.e. a slow takeoff to an altitude of 2 meters.
```
mhkabir: Offboard should not be in uppercase throughout the document.

andre-nguyen:
I think we should also redo the video. When I made it in 2015, takeoff setpoints weren't implemented yet. I think the takeoff is much faster now.

TSC21:
> Offboard should not be in uppercase throughout the document.

`OFFBOARD` is a flight mode and usually we see the flight modes referenced uppercased (`POSCTL`, `STABILIZED`, ...) - https://docs.px4.io/en/flight_modes/. That was my approach.

> I think we should also redo the video. When I made it in 2015, takeoff setpoints weren't implemented yet. I think the takeoff is much faster now.

@andre-nguyen on MAVROS side we still don't support takeoff setpoints (@tinito created mavlink/mavros#393 but it's staled right now). The video seems ok as an example of `OFFBOARD` control. So unless we add that support, this can still be our example.

en/ros/mavros_offboard.md
hamishwillee: In fairness, as stated below (https://github.com/PX4/Devguide/pull/209#issuecomment-312496575) some docs capitalise [Flight Modes (Overview)](https://docs.px4.io/en/config/flight_mode.html) and others don't [Basic Configuration > Flight Modes](https://docs.px4.io/en/config/flight_mode.html).

I have a leaning towards capitalisation because (hence [RTL](https://docs.px4.io/en/flight_modes/rtl.html), [LAND](https://docs.px4.io/en/flight_modes/land.html)) but it doesn't really matter as long as we are consistent.

@mhkabir Thoughts? (if you have a strong preference let me know. I can push whatever we agree into style guide/out through docs)

mhkabir: @hamishwillee The capitalised variant is the internal state machine representation. The user-facing documentation should have the flight modes like Offboard, Position, Altitude, Return, Land, Takeoff, etc. We should do a run through all the documentation and fix it everywhere. This would also make it consistent with QGC. 

image

ndepal:
This is not new in this PR, just noticed while glancing over this: Shouldn't [this line](https://github.com/PX4/Devguide/pull/209/files#diff-855b2f19c8d2fa3693cafe4536d0ad2dR49) be

```
while(ros::ok() && !current_state.connected){
```
(i.e. wait while _not_ connected)?

TSC21:
@ndepal have you noticed the exclamation point?

TSC21:
@hamishwillee can you round up this for a last review? Should we keep it as `OFFBOARD` or just "Offboard"?

en/ros/mavros_offboard.md
hamishwillee: As long as we're consistent that is fine. I've got some work in the user guide to make it like this.

hamishwillee:
As per @mhkabir comment https://github.com/PX4/Devguide/pull/209#discussion_r125213866 we should be consistent, and use the text as per QGC  - since that is what the user sees. 

I still think it is important that it is easy to see, so perhaps change `OFFBOARD`  to *Offboard mode*, *Offboard control*, *Offboard command* as appropriate. Seem reasonable?

If this is a step too far, feel free to merge - if we're going to have to do this globally we'll catch it when we get to it!

TSC21:
> I still think it is important that it is easy to see, so perhaps change `OFFBOARD` to *Offboard mode*, *Offboard control*, *Offboard command* as appropriate. Seem reasonable?

Yes I agree. I will be consistent with this on this PR and maybe we can have a look at all the DevGuide in another PR and fix where with don't have this consistency. What do you think? (Btw, I don't have write permissions over the DevGuide repo, so the control is yours)

hamishwillee:
Works for me! Fixing everything else for consistency is not a super high priority, but I will add it to my list.

TSC21:
@hamishwillee does it seem ok to you now? Maybe we should run a sanity check after this PR on the rest of the documentation to keep the consistency

hamishwillee:
THanks. I think there may be one more case of `OFFBOARD` - see "We set the custom mode to"

I haven't otherwise fully tested this, though I will at some point.

TSC21:
> THanks. I think there may be one more case of `OFFBOARD` - see "We set the custom mode to"

It was on purpose, given that the "custom mode" is code related. (Meaning that you code `OFFBOARD` and not `Offboard`).

hamishwillee:
Awesome :-

TSC21:
@hamishwillee can you get this in then?

hamishwillee:
@TSC21 Often I can't do  technical review" because I don't know enough about the area. I have merged since it makes sense to me and is an iteration of an existing doc - but as normal case, great to get a technical expert in area to comment too.

TSC21:
Well in that case it could be me xD jocking but yeah on these cases it's always good to have someone that can take a look. Though this was a very simple change that was introduced on upstream MAVROS.

hamishwillee:
Exactly :-) You an always make a statement "just requires editorial review/sanity check" and I won't worry so much.

TSC21:
Ahah yeah will take that into account next time ;) thanks @hamishwillee

PR #215, opened 2017-07-07
ULog: update ulog file spec to version 2

bkueng:
This documents the changes from https://github.com/PX4/Firmware/pull/7520

@ecmnet fyi this adds new ULog messages, maybe relevant for MAVGCL. Other than that existing parsers are compatible.


bkueng:
Merging, as the Firmware changes are already merged.

PR #219, opened 2017-07-13
Adding eProsima micro RTPS documentation

vicentemonge:
To add the PX4-Fast RTPS bridge documentation in the middleware section, including usage and examples.


en/middleware/micrortps.md
```diff
@@ -0,0 +1,338 @@
+# PX4-FastRTPS bridge
+
+
+This bridge adds communication capabilities between a **PX4 Autopilot** and a **Fast RTPS** application through serial ports or
+UDP sockets using **CDR serialization**. The goal is to provide a DDS (Data Distribution Service) interface to PX4. This interface allows sharing information also with the upcoming release of ROS (Robot Operating System), ROS2.
+
+DDS is an standard from the OMG (Object Management Group) providing a real-time pub/sub middleware widely used in aerospace, defense and IoT applications, while in Robotics has been adopted as the middleware for ROS2.
+
+![alt text](res/1_general-white.png)
```
hamishwillee: None of these images listed in this doc exist/are being displayed. The images should be in the /assets folder (so perhaps /assets/middleware/rtps). They should also have some clear distinguishing name so that it is obvious what they are for (this makes it easier to find them later). The "alt text" should also be descriptive.

hamishwillee:
This should probably cross-link to https://dev.px4.io/en/setup/fast-rtps-installation.html ?

en/middleware/micrortps.md
```diff
@@ -0,0 +1,338 @@
+# PX4-FastRTPS bridge
+
+
+This bridge adds communication capabilities between a **PX4 Autopilot** and a **Fast RTPS** application through serial ports or
+UDP sockets using **CDR serialization**. The goal is to provide a DDS (Data Distribution Service) interface to PX4. This interface allows sharing information also with the upcoming release of ROS (Robot Operating System), ROS2.
```
hamishwillee: Would be good to turn **CDR serialization** into a link. I don't know what this term means.

en/middleware/micrortps.md
```diff
@@ -0,0 +1,338 @@
...
+
+This covers the entire spectrum of communications.
+
+![alt text](res/4_both-white.png)
+
+These pieces of code are generated within the normal PX4 Firmware generation process. Also can be generated under demand calling the script **generate_microRTPS_bridge.py** placed in *Tools* folder, see section below.
+
+### Generating the client and the agent
+
+**NOTE**: Before continue we need to have installed Fast RTPS. Visit its installation
```
hamishwillee: Should this link be to https://dev.px4.io/en/setup/fast-rtps-installation.html ( ../setup/fast-rtps-installation.md) ?

hamishwillee: Note syntax in gitbook looks like this:
```
> **Note** Before ...
```

en/middleware/micrortps.md
```diff
@@ -0,0 +1,338 @@
...
+
+DDS is an standard from the OMG (Object Management Group) providing a real-time pub/sub middleware widely used in aerospace, defense and IoT applications, while in Robotics has been adopted as the middleware for ROS2.
+
+![alt text](res/1_general-white.png)
+
+## Automatic code generation
+
+The support for the functionality is mainly done within three new (automatically generated) code blocks.
+
+-  CDR serialization support is generated for every uORB topic adding a interfaces what for *sensor_combined.msg* topic looks like this:
```
hamishwillee: I THINK you mean 
"The toolchain enables CDR serialisation support by adding two interface methods for every uORB topic: `serialize_*` and  `deserialize_*`. For the *sensor_combined.msg* topic the methods look like this:"

en/middleware/micrortps.md
```diff
@@ -0,0 +1,338 @@
...
+   sensor_combined
+   telemetry_status
+   wind_estimate
+   ...
+   )
+```
+
+The client application will be generated in *build_OURPLATFORM/src/modules/micrortps_bridge/micrortps_client/* folder and the agent will be created in *src/modules/micrortps_bridge/micrortps_agent/* folder.
+
+**NOTE**: In the process of the agent generation we use a Fast RTPS tool called **fastrtpsgen**. If you don't have installed Fast RTPS in the default path you need to specify the directory of installation of fastrtpsgen setting the environment variable **FASTRTPSGEN_DIR** before make executing in this way:
```
hamishwillee: "If you don't have installed" -> If you haven't installed.

en/middleware/micrortps_hellow_world.md
```diff
@@ -0,0 +1,115 @@
+# Hello world
```
hamishwillee: Good to have more context for search engine - "RTPS Hello World Example"

hamishwillee: This file does not appear to be in the SUMMARY.md. It will therefore not be generated as HTML.

en/middleware/micrortps_throughput_test.md
```diff
@@ -0,0 +1,105 @@
+# Throughput test
```
hamishwillee: Title should have more context. Use First Letter capitalisation "RTPS Throughput Test". Is this in SUMMARY.md.

Cross link from other docs so it can be found.

en/middleware/micrortps_hellow_world.md
hamishwillee: Cross-link to other docs about RTPS so this can be found.

en/middleware/micrortps_hellow_world.md
```diff
@@ -0,0 +1,115 @@
...
+
+As a basic example we go to explain how implement a simple use case what sends information of some sensors (*sensor_combined* uORB topic) in the direction PX4-Fast RTPS and receives information as a text message (*log_message* uORB topic) sent in the other direction.
```
hamishwillee: Can we have a link to the example source files in github?

hamishwillee:
@VicenteEprosima This looks pretty good, but there are some important issues that should be fixed up. Most important being:
1. Images aren't in this PR. They should be in /assets/ sub folder and have names that are easy to identify as being part of rtps/this topic.
2. A couple of the docs aren't in the SUMMARY.md. They need to be or they won't be built by the toolchain.

@bkueng This needs a good English and "Gitbook" subedit. I don't want to dump all of that as comments - would be better for me to do that myself. How do you normally manage that? Submit the content and then do work to apply fixes as a patch or PR direct in the submitter's branch?

bkueng:
> @bkueng This needs a good English and "Gitbook" subedit. I don't want to dump all of that as comments - would be better for me to do that myself. How do you normally manage that? Submit the content and then do work to apply fixes as a patch or PR direct in the submitter's branch?

We usually do one of the following: either working on the same branch if the author gives you permissions, or opening a new PR that includes the commits from this one (after the author replied to the initial review).

hamishwillee:
Hi @VicenteEprosima , so as above https://github.com/PX4/Devguide/pull/219#issuecomment-315688267 let me know whether you can give me access to the branch after you've made your comments/fixes.

vicentemonge:
Hi @hamishwillee, of course, we are working still in the changes. I'll let you know when we're done. 

Thank you!

hamishwillee:
@VicenteEprosima Thanks very much for letting me know. Sorry I jumped in too early :-)

vicentemonge:
@hamishwillee I have [invited](https://github.com/VicenteEprosima/Devguide/invitations) you to my Devguide fork. You can start to tune the documentation. Thanks for your work!

en/setup/fast-rtps-installation.md
```diff
@@ -1,13 +1,21 @@
...
+<img src="../../assets/fastrtps/eprosima_logo.png" style="float:left;"/> [eProsima Fast RTPS](http://eprosima-fast-rtps.readthedocs.io/en/latest/) is a C++ implementation of the RTPS (Real Time Publish Subscribe) protocol, which provides publisher-subscriber communications over unreliable transports such as UDP, as defined and maintained by the Object Management Group (OMG) consortium. RTPS is also the wire interoperability protocol defined for the Data Distribution Service (DDS) standard, again by the OMG.
 
-**NOTE**: The information here was extracted from original [_eprosima Fast RTPS_ documentation](http://eprosima-fast-rtps.readthedocs.io/en/latest/)
+> **Tip** Fast RTPS is used by PX4 to enable publish/subscribe of uOrb topics to be extended to offboard components, including robotics and simulator tools. For more information see: [Micro RTPS](../middleware/micrortps.md).
```
hamishwillee: @VicenteEprosima I updated this page.  This paragraph is the most important change for you to comment on as it cross-links to your docs.
1. Is this a correct description of what the bridge seeks to achieve?
2. I named your doc "Micro RTPS" but you called it "DDS/ROS2 interface". I think we need a name that "relates" this to RTPS.

vicentemonge: @hamishwillee 
1. A more accurate description is found in the beginning of micrortps.md
2. I changed DDS for RTPS
~~3. We need a link to our readthedocs~~ I see now all the changes and it's ok
*Please update to last commit

vicentemonge:
@vilhjalmur89 maybe you could push your changes to this branch. I think the new Hello World you already done should be there. What do you think @hamishwillee?

hamishwillee:
I think you need to wait a few minutes for me to untangle our merged changes - I'm trying to submit :-)

hamishwillee:
@VicenteEprosima  @vilhjalmur89 OK, I just spent a bunch of time fixing up the micrortps.md doc and making sure that the other two docs build without gitbook errors. There are some issues with this that I will add as comments. 

@vilhjalmur89 If you want to merge your changes in then fine. In particular for the examples that would be great - they are contained and the errors are pretty easy to fix. From my perspective I just don't want to "stamp" on each other, so after my comments I won't do anything until next week.

PS I also did not try to merge your changes - I didn't know enough when I started to understand what you were trying to do.

en/SUMMARY.md
```diff
@@ -61,6 +61,9 @@
 * Middleware and Architecture
   * [uORB Messaging](middleware/uorb.md)
   * [MAVLink Messaging](middleware/mavlink.md)
+  * [DDS/ROS2 interface](middleware/micrortps.md)
```
hamishwillee: Is there a better name than this? Perhaps "PX4/RTPS Bridge" or "RTPS DDS"  Not sure, but I wouldn't see this and think "hey yes, that's how I publish uORB topics externally"

en/SUMMARY.md
```diff
@@ -61,6 +61,9 @@
...
+    * [MicroRTPS HelloWorld Example](middleware/micrortps_hello_world.md)
```
hamishwillee: For the sidebar (only) we could arguably remove "MicroRTPS" because you get the context from the parent item

en/setup/fast-rtps-installation.md
```diff
@@ -1,13 +1,21 @@
...
+<img src="../../assets/fastrtps/eprosima_logo.png" style="float:left;"/> [eProsima Fast RTPS](http://eprosima-fast-rtps.readthedocs.io/en/latest/) is a C++ implementation of the RTPS (Real Time Publish Subscribe) protocol, which provides publisher-subscriber communications over unreliable transports such as UDP, as defined and maintained by the Object Management Group (OMG) consortium. RTPS is also the wire interoperability protocol defined for the Data Distribution Service (DDS) standard, again by the OMG.
 
-**NOTE**: The information here was extracted from original [_eprosima Fast RTPS_ documentation](http://eprosima-fast-rtps.readthedocs.io/en/latest/)
+> **Tip** Fast RTPS is used by PX4 to enable publish/subscribe of uORB topics to be extended to offboard components, including robotics and simulator tools. For more information see: [Micro RTPS](../middleware/micrortps.md).
```
hamishwillee: @VicenteEprosima I edited this doc to be more readable and match other docs in the guide. This is the most important comment as it is cross link to your document. Question, is this an accurate description of what the bridge is trying to achieve?

I'm not completely happy with the title of the micrortps.md doc. When that settled the name in this paragraph will have to change to match it.

en/middleware/micrortps.md
```diff
@@ -0,0 +1,344 @@
+# RTPS/ROS2 interface: the PX4-FastRTPS bridge
+
+This bridge adds communication capabilities between a *PX4 Autopilot* and a *Fast RTPS* application, through serial ports or
```
hamishwillee: I think this whole section needs work because it isn't clear why you would bother creating a DDS service. 

As I understand it the point of RTPS is to be able to publish and subscribe uORB topics beyond the flight controller so that external components can subscribe as though they were on the flight controller. This allows deep integration with simulators and off board APIs. 

Furthermore that we have RTPS that is a "data neutral" mechanism for sharing data. What this bridge does is serialise our data, create a bridge between uorb and RTPS, and open a transport layer for the messages.

en/middleware/micrortps.md
```diff
@@ -0,0 +1,344 @@
...
+UDP sockets, using [CDR serialization](https://en.wikipedia.org/wiki/Common_Data_Representation). The goal is to provide a DDS (Data Distribution Service) interface to PX4. This interface will also allow sharing information with the forthcoming release of ROS2 (Robot Operating System).
+
+RTPS is the underlying protocol of DDS, an standard from the OMG (Object Management Group) providing a real-time pub/sub middleware widely used in aerospace, defense and IoT applications, while in Robotics has been adopted as the middleware for ROS2.
+
+Fast RTPS implements the latest version of the RTPS protocol and a minimum DDS API, resulting in a very lightweight implementation of the standard and full access to the RTPS fine settings.
+
+![micro RTPS general scheme](../../assets/middleware/micrortps/micrortps_general_scheme.png)
```
hamishwillee: I don't find this image useful. You really can't infer anything about the system/intention of the document. The arrows don't appear to mean anything. Nor do where they come from or go to.

en/middleware/micrortps.md
```diff
@@ -0,0 +1,344 @@
...
+  ```sh
+  void serialize_sensor_combined(const struct sensor_combined_s *input, char *output, uint32_t *length, struct microCDR *microCDRWriter);
+  void deserialize_sensor_combined(struct sensor_combined_s *output, char *input, struct microCDR *microCDRReader);
+  ```
+  
+  > **Note** *CDR serialization* provides a common format for exchanging serial data between different platforms.  
+
+- A *client application* that sends and receives the CDR serialized data from topics through a selected UART or UDP port.
+
+  ![micro RTPS client](../../assets/middleware/micrortps/micrortps_client.png)
```
hamishwillee: Image mixes too many concepts - data flow, code blocks. 

What I'd expect to see here is a diagram that either shows the inputs to the client application and the outputs (ie the the data flow within the application - serialisation, then out to uart. )
From a code/build perspective you might show the inputs/outputs of the build process. 

But what we have here is a mix of both. I don't know what you're trying to show.

en/middleware/micrortps.md
```diff
@@ -0,0 +1,344 @@
...
+
+- An *agent application* that shares the CDR serialized data via (using **Fast RTPS**) to DDS service clients (e.g. ROS2). An important step in this part is the generation of a IDL file for each topic which is the translation of the corresponding MSG file. For the case of *sensor_combined* topic the **sensor_combined_.idl** file is generated from **sensor_combined.msg** file.
+
+  ![micro RTPS agent](../../assets/middleware/micrortps/micrortps_agent.png)
```
hamishwillee: See previous comment. Not sure what you're trying to show. I'd be interested in either how messages flow, or how the code generator works.

en/middleware/micrortps.md
```diff
@@ -0,0 +1,344 @@
...
+  void deserialize_sensor_combined(struct sensor_combined_s *output, char *input, struct microCDR *microCDRReader);
+  ```
+  
+  > **Note** *CDR serialization* provides a common format for exchanging serial data between different platforms.  
+
+- A *client application* that sends and receives the CDR serialized data from topics through a selected UART or UDP port.
+
+  ![micro RTPS client](../../assets/middleware/micrortps/micrortps_client.png)
+
+- An *agent application* that shares the CDR serialized data via (using **Fast RTPS**) to DDS service clients (e.g. ROS2). An important step in this part is the generation of a IDL file for each topic which is the translation of the corresponding MSG file. For the case of *sensor_combined* topic the **sensor_combined_.idl** file is generated from **sensor_combined.msg** file.
```
hamishwillee: So the bit where it says "An important step in this part " mixes data flow with code generation. I think we need to separate the topics.

en/middleware/micrortps.md
```diff
@@ -0,0 +1,344 @@
...
+
+  ![micro RTPS agent](../../assets/middleware/micrortps/micrortps_agent.png)
+
+The diagram below covers the entire spectrum of communications.
+
+![micro RTPS both sides summary](../../assets/middleware/micrortps/micrortps_boths_sides.png)
```
hamishwillee: Again, confusing. I think using .idl and .msg as labels does not help - are you indicating that these files are sent/received (I very much doubt it!). 

The box (everywhere) with PX4 + RTSP is confusing. If the micro RTPS was on side connecting to the RTPS on other side perhaps more useful?

en/middleware/micrortps.md
```diff
@@ -0,0 +1,344 @@
...
+
+These pieces of code are generated within the normal PX4 Firmware generation process. They can also can be generated explicitly by calling the script **Tools/generate_microRTPS_bridge.py** (see section below).
```
hamishwillee: I added a section heading for the script - perhaps we should link to it?

en/middleware/micrortps.md
```diff
@@ -0,0 +1,344 @@
...
+
+### Automatically generate client and the agent code
+
+> **Note** Before continuing we need to have [installed Fast RTPS](../setup/fast-rtps-installation.md).
```
hamishwillee: Do we really need Fast RTPS installed at this point (for generation)?

en/middleware/micrortps.md
```diff
@@ -0,0 +1,344 @@
...
+   ...
+   )
+```
+
+The client application will be generated in *build_OURPLATFORM/src/modules/micrortps_bridge/micrortps_client/* folder and the agent will be created in *src/modules/micrortps_bridge/micrortps_agent/* folder.
+
+> **Note** In the process of the agent generation we use a Fast RTPS tool called *fastrtpsgen*. If you haven't installed Fast RTPS in the default path you need to specify the directory of installation of *fastrtpsgen* setting the environment variable `FASTRTPSGEN_DIR` before executing *make*. 
+> On Linux/Mac this is done as shown below:
+>
+>  ```sh
```
hamishwillee: I changed this to "on Linux/Mac". Would you ever need to do this on Windows?

en/middleware/micrortps.md
```diff
@@ -0,0 +1,344 @@
...
+$ python Tools/generate_microRTPS_bridge.py -h
+usage: generate_microRTPS_bridge.py [-h] [-s *.msg [*.msg ...]]
+                                    [-r *.msg [*.msg ...]] [-a] [-c]
+                                    [-t MSGDIR] [-o AGENTDIR] [-u CLIENTDIR]
+                                    [-f FASTRTPSGEN]
+
+optional arguments:
+  -h, --help            show this help message and exit
+  -s *.msg [*.msg ...], --send *.msg [*.msg ...]
+                        Topics to be sended
```
hamishwillee: Should be "topics to be sent"

en/middleware/micrortps.md
```diff
@@ -0,0 +1,344 @@
...
+                        src/modules/micrortps_bridge/micrortps_agent
+  -u CLIENTDIR, --client-outdir CLIENTDIR
+                        Client output dir, by default
+                        src/modules/micrortps_bridge/micrortps_client
+  -f FASTRTPSGEN, --fastrtpsgen-dir FASTRTPSGEN
+                        fastrtpsgen installation dir, by default /bin
+  --no-delete           Do not delete dir tree output dir(s)
+```
+
+- The argument `--send/-s` means that the application from PX4 side will send these messages, and the argument `--receive/-r` specifies which messages is going to be received.
```
hamishwillee: "messages is going" should be "messages are going"

So you can declare multiple messages for -s and -r? Can you call this multiple times to generate the new code?

I also assume that if you do this you don't need the .cmake changes?

en/middleware/micrortps.md
```diff
@@ -0,0 +1,344 @@
...
+- The `-f` option may be needed if *Fast RTPS* was not installed in the default location (`-f /path/to/fastrtps/installation/bin`).
+
+An example of use:
+
+```sh
+$ cd /path/to/PX4/Firmware
+$ python Tools/generate_microRTPS_bridge.py -s msg/sensor_baro.msg -r msg/sensor_combined.msg
+```
+
+Checking the correct installation:
```
hamishwillee: What are we checking the correct installation of? The script? Or do you mean "generation"

Is this just for the use of the script or also relevant to the automatic process?

en/middleware/micrortps.md
```diff
@@ -0,0 +1,344 @@
...
+src/modules/micrortps_bridge/micrortps_client
+├── CMakeLists.txt
+├── microRTPS_client.cpp
+├── microRTPS_transport.cxx
+└── microRTPS_transport.h
+ 0 directories, 4 files
+```
+
+
+## PX4 Firmware: the micro RTPS client
```
hamishwillee: What does having "PX4 Firmware" as a prefix tell us in this heading? Would this be better as just "RTPS client application"?

en/middleware/micrortps.md
```diff
@@ -0,0 +1,344 @@
...
+
+The agent application runs as both a uORB node and as a micro RTPS node. As a uORB node the agent could be subscribed to several topics as well as publish under internal uORB topics. The application receives from a internal publishers the messages, serializes the struct and writes it trough an UART
```
hamishwillee: What are you trying to say with "As a uORB node the agent could be subscribed to several topics as well as publish under internal uORB topics"

I think you're saying that it may be subscribed to internal uORB topics, and act as an external publisher over RTPS AND that it may be a publisher of uORB topics FROM remote uORB systems that it receives via RTPS. Correct?

en/middleware/micrortps.md
```diff
@@ -0,0 +1,344 @@
...
+device or UDP port selected by the user. Also will be reading from the UART device or UDP port and then publish the info
+to the internal subscribers.
+
+Steps to use the auto generated application:
```
hamishwillee: The changes to cmake would seem to be some sort of aside or subtopic - no one should ever need to actually do this but we need to keep the information for someone adding the feature at another time to a new target. Do you think we can/should separate this?

en/middleware/micrortps.md
```diff
@@ -0,0 +1,344 @@
...
+  > micrortps_client start #by default -t UART -d /dev/ttyACM0 -u 0 -l 10000 -w 1 -b 460800 -p 1
+  ```
+
+<span></span>
+> **Note** If the selected UART port is busy, it's possible that the MAVLink application is already being used. If it is the case, you can stop MAVLink from NuttShell by typing:
+  ```sh
+  > mavlink stop-all
+  ```
+ 
+## Fast RTPS: the Micro RTPS agent
```
hamishwillee: What value does having prefix "Fast RTPS" have here in the title.?

en/middleware/micrortps.md
```diff
@@ -0,0 +1,344 @@
...
+
+The *Fast RTPS* side will be explained taking a *Raspberry Pi* board to run an application as example.
```
hamishwillee: The text below here to "To create the application, compile the code:" seems to have some duplication and repetition.

What isn't clear is what is the role of the agent, and how it fits with the client. What I guess is that this is the PX4 side code that implements FastRTPS protocol and talks to the protocol on the remote system (e.g. RaPi). I THINK that this is sharing information with the client. So uOrb -> client (serialise) -> agent (rtps on px4)  sends rtps -> port to remote system -> rtps on remote system ....

en/middleware/micrortps.md
```diff
@@ -0,0 +1,344 @@
...
+  $ sudo usermod -a -G dialout pi
+  ```
+
+2. You need to stop the already running on the GPIO serial console:
+
+  ```sh
+  $ sudo raspi-config
+  ```
+
+  Go to **Interfacing options > Serial**. Select **NO for *Would you like a login shell to be accessible over serial?*. Valid and reboot.
```
hamishwillee: I don't understand these instructions. Are they prompts I'll see somewhere in some UI?

en/middleware/micrortps.md
```diff
@@ -0,0 +1,344 @@
...
+
+3. Check UART in kernel:
+
+  ```sh
+  $ sudo vi /boot/config.txt
+  ```
+
+And enable UART setting `enable_uart=1`.
```
hamishwillee: How?

en/middleware/micrortps.md
```diff
@@ -0,0 +1,344 @@
...
+
+## Graphical example of usage
+
+This flow chart shows graphically how the bridge works. It demonstrates a bridge that sends the topic `sensor_combined` from a Pixracer to a Raspberry Pi through UART.
+
+![basic example flow](../../assets/middleware/micrortps/basic_example_flow.png)
+
+If all steps has been followed, you should see this output on the subscriber side of Fast RTPS.
+
+![Fast RTPA subscriber](../../assets/middleware/micrortps/fastrtps_subscriber_ex.png)
```
hamishwillee: Does not quite match the preceding code (received is spelled wrong, so obviously this was run before that was fixed). Is it otherwise OK?

en/middleware/micrortps.md
```diff
@@ -0,0 +1,344 @@
...
+
+## Hello world and Throughput test
+
+The [Hello world](../middleware/micrortps_hello_world.md) and [Throughput test](../middleware/micrortps_throughput_test.md) show some real-world examples of how to use the features described in this topic. 
+
+## Graphical example of usage
+
+This flow chart shows graphically how the bridge works. It demonstrates a bridge that sends the topic `sensor_combined` from a Pixracer to a Raspberry Pi through UART.
+
+![basic example flow](../../assets/middleware/micrortps/basic_example_flow.png)
```
hamishwillee: "Almost" best diagram so far. Is the arrow in the middle UART connection to RaPI? If so, great (the image on the right is supposed to be a RaPi. 
Can this have the Agent and Client overlayed?

en/SUMMARY.md
vicentemonge: *RTPS/ROS2 interface* we think is the correct one

en/SUMMARY.md
vicentemonge: Ok

en/middleware/micrortps.md
vicentemonge: We have modified this part in the last commit, update to latest version and tell me what you think.

vilhjalmur89:
I added a simple example for creating a node to subscribe to sensor_combined.

I think it makes more sense than the 'Hello World' example since it creates a new node instead of changing the automatically generated Agent.

I also think we should keep it on *micrortps.md* instead having a separate page, since it depends on the topics chosen in the beginning.

en/middleware/micrortps.md
vilhjalmur89: Yes, we need fastrtpsgen.

en/setup/fast-rtps-installation.md
vicentemonge: Already changed in the last commit.

en/middleware/micrortps.md
vicentemonge: For simplicity we are going to remove these images.

en/middleware/micrortps.md
vicentemonge: For simplicity we are going to remove these images.

en/middleware/micrortps.md
vicentemonge: I have changed "important" for "first" for pointing that is an clarification over the process.

en/middleware/micrortps.md
vicentemonge: Avoided

en/middleware/micrortps.md
vicentemonge: Yes

en/middleware/micrortps.md
vicentemonge: Not tested for now

en/middleware/micrortps.md
vicentemonge: Changed and clarified.

en/middleware/micrortps.md
vicentemonge: Only for example purposes to list here the files created.

en/middleware/micrortps.md
vicentemonge: _**PX4 Firmware side: the micro RTPS client**_  sounds good and underlines that the app for PX4 side is the client. Is correct for you?

en/middleware/micrortps.md
vicentemonge: Yes

en/middleware/micrortps.md
vicentemonge: The same as for PX4 side...

en/middleware/micrortps.md
vicentemonge: Ok, separating this may have more sense.

en/middleware/micrortps.md
vicentemonge: Compile the code, i think is pointed only here.
Yeah, this is the role of the agent

en/middleware/micrortps.md
vicentemonge: changed

en/middleware/micrortps.md
vicentemonge: Setting the value in the text file as shown.

vicentemonge:
Sounds good Vili, thanks!

en/middleware/micrortps.md
vicentemonge: Done.

vicentemonge:
@vilhjalmur89 @hamishwillee I'll be on vacation next week. Please enter any changes that you think are necessary for the integration.

Thanks!!

vilhjalmur89:
I simplified the instructions quite bit and removed most of the things that I felt were out of place. Feel free to add them back in case you think they need to be there.

I think we need to separate the description of how things work behind the scene, and what you need to do to get this running. So, now there is a short description at the beginning and the rest focuses on running the example. Maybe it makes more sense to have a 'Quick Start'-guide on a separate page and have a detailed explanation on the main page.

hamishwillee:
Guys, I think I should take a proper hack at this tomorrow. Are you still working on this - I don't want to trample on your changes.

BTW, I'm not sure we needed to separate out the "manual approach" (though it won't hurt). Moving the Listener code into the doc was excellent. Would be good to have an example of a publisher too.

Shouldn't the client always run and be installed as part of intialisation code for PX4?

hamishwillee:
@vilhjalmur89 
Just looked at your changes - excellent, that is most of what I was going to do - image up the top showing the parts interacting, clear description of their function, separating out what the toolchain does a bit more in terms of creating IDL files from .msg from the what the end result is (client and agent).

Some of the rendering is broken.

The further improvements I would make are that the top image is still poor. It doesn't really need the image of flight controller or whatever the thing on the right is. It should probably show the data flow in both directions, have labels on the channel, and possibly show arrows coming from "uORB".

"The Agent similarly acts as a bridge between Fast RTPS messages on the offboard computer and CDR-messages on either UART or UDP."?
Are they CDR messages, or are they FastRTPS messages from the client? I ask because a) RTPS uses CDR, b) you previously referred to it as a microRTPS node.
I will do minor mods here, but keep the flavour.

The client should probably be a permanent daemon that gets started in intialisation code.

We may need a publisher example as well as a listener.

I still think maybe the primary goal is to share uORB topics to ROS. After all, we haven't added any code to do anything else.

vilhjalmur89:
@hamishwillee That would be great, I won't change anything until then.

I think the automatic code generation should ideally handle most of the use-cases, in which case the 'manual approach' breaks up the flow. At the very least, I think the chapter is too long to be an optional step in the main page.

I agree about the image at the top and the need for a publisher. The Client should be started on start-up, but that is not the case right now.

I guess the messages are both FastRTPS and CDR messages so let's go with whatever is most clear.

en/setup/fast-rtps-installation.md
```diff
@@ -20,10 +28,9 @@ Java is required to make use of our built-in code generation tool *fastrtpsgen*.
 
 *eProsima Fast RTPS* requires the Visual C++ Redistributable packages for the Visual Studio version you choose during the installation or compilation. The installer gives you the option of downloading and installing them.
```
hamishwillee: @VicenteEprosima @vilhjalmur89 "eProsima Fast RTPS* requires the Visual C++ Redistributable packages for the Visual Studio version you choose during the installation or compilation. "
This is confusing, because this is a requirement - at this point I haven't chosen or installed anything, so it isn't clear where I should have got the visual studio. Can you clarify?

en/setup/fast-rtps-installation.md
```diff
@@ -1,13 +1,21 @@
-# Fast RTPS installation ![](../../assets/fastrtps/eprosima_logo.png)
+# Fast RTPS installation
```
hamishwillee: @VicenteEprosima @vilhjalmur89 Do the Linux instructions work on Mac OSX?

en/setup/fast-rtps-installation.md
vilhjalmur89: Maybe @JaimeMartin can clarify?

hamishwillee:
OK, I've updated the main doc to have a better diagram and structure. I'm interested to know what you think. For me anyway it is more clear what the interactions are, and what you have to do. You will note that some comments got pushed into troubleshooting section - I think this is a good "catchall" for most deviations from the "normal process". 
1. I will look at the other docs tomorrow. 
2. I'll add a few more questions in place.
3. I think we're getting very close to done :-)

en/middleware/micrortps.md
```diff
@@ -0,0 +1,271 @@
...
+
+set(config_rtps_receive_topics
+   vehicle_command
+   # Add new topic...
+   )
+```
+
+A uORB serialization function will be generated for each topic in both lists. An IDL file will also be created for each topic from the associated uORB **.msg** files 
+
+> **Note** FastRTSP uses IDL files to define the structure of RTPS messages (in this case, RTPS messages for sending uORB topics). The IDL files are compiled to C++ by the *fastrtpsgen* tool, which is then used by the Agent and other FastRTPS applications.
```
hamishwillee: @vilhjalmur89 I'm inclined to create another topic for this comment at the same level as the next one "uORB serialisation code". Reason being that both are "intermediate code" created as part of generation process. We don't care about them as such, as a user, but worth highlighting?

en/middleware/micrortps.md
```diff
@@ -0,0 +1,271 @@
...
+
+
+#### uORB serialization code
```
hamishwillee: @vilhjalmur89 Note that really the serialisation code isn't of that much interest to people wanting to understand the bridge. I've therefore moved it under the generation here as a "for information" rather than part of the detail of the architecture.

en/middleware/micrortps.md
```diff
@@ -0,0 +1,271 @@
...
+  -u <update_time_ms>     Time in ms for uORB subscribed topics update. Default 0
+  -l <loops>              How many iterations will this program have. -1 for infinite. Default 10000
+  -w <sleep_time_ms>      Time in ms for which each iteration sleep. Default 1ms
+  -b <baudrate>           UART device baudrate. Default 460800
+  -p <poll_ms>            Time in ms to poll over UART. Default 1ms
+  -r <reception port>     UDP port for receiving. Default 2019
+  -s <sending port>       UDP port for sending. Default 2020
+```
+
+> **Note** If the selected UART port is busy, it's possible that the MAVLink application is already being used. If it is the case, you can stop MAVLink from NuttShell by typing:
```
hamishwillee: @vilhjalmur89  Will this happen because MAVLink is being used by "accident" and should not be (e.g. it is a configuration issue and the port should be dedicated to the bridge), or will it happen because the port must be shared?
1. If the former, then I'd say that is a bug, and in "real use" this should not happen as part of the normal flow. In this case would this not be a troubleshooting bug while we start moving from MAVLInk to RTPS where needed?
2. If it is actually a shared port and this could happen intentionally then turning off MAVLink isn't really a solution - right?

en/middleware/micrortps.md
```diff
@@ -0,0 +1,271 @@
...
+For NuttX/Pixhawk flight controllers:
+```sh
+make px4fmu-v4_default upload
+```
+For Qualcomm Snapdragon Flight:
+```sh
+$ make eagle_default upload
+  ```
+
+After uploading the firmware, the *Client* application can be launched by typing its name and passing a variable number of arguments as shown below:
```
hamishwillee: @vilhjalmur89 @LorenzMeier Manually launching the RTPS client sounds very wrong. Normally a process in firmware that is needed should be started by the configuration and then run forever. What is going on here?

en/middleware/micrortps.md
```diff
@@ -0,0 +1,271 @@
...
+  -d <device>             UART device. Default /dev/ttyACM0
+  -w <sleep_time_us>      Time in us for which each iteration sleep. Default 1ms
+  -b <baudrate>           UART device baudrate. Default 460800
+  -p <poll_ms>            Time in ms to poll over UART. Default 1ms
+  -r <reception port>     UDP port for receiving. Default 2019
+  -s <sending port>       UDP port for sending. Default 2020
+```
+
+
+## Creating a Listener application
```
hamishwillee: This is great. Possible to get a publisher example as well that does something realistic?

en/middleware/micrortps.md
vilhjalmur89: I actually think this along with other details could be moved to _Manually Generate Client/Agent_, which would than explain how the code generation works, and how to run it manually.

en/middleware/micrortps.md
vilhjalmur89: Yes, I agree. It just needs to be implemented.

en/middleware/micrortps.md
vilhjalmur89: @hamishwillee @VicenteEprosima It can be either of them.
1. In this case you probably selected the wrong port for either the bridge or mavlink. So stopping mavlink is good enough for testing, but is not a real solution
2. In this case some documentation is needed for how to run the bridge and mavlink on the same port.

en/middleware/micrortps.md
hamishwillee: So who makes that happen and when? For now we could do:

----
After uploading the firmware, the *Client* application can be launched from the [System console](https://dev.px4.io/en/debug/system_console.html) by typing its name and passing a variable number of arguments as shown below:
```
XXXX - the tool invocation
````
> **Note** In the near future the Client will automatically be launched by the system initialisation code.
----
What do you think. Will this work? My concern being that if you use the MAVLink console (most likely) this will screw up the port so you can't use the bridge.

File: en/middleware/micrortps.md (hunk 38 above)
Comment (hamishwillee) [reply en/middleware/micrortps.md]: @vilhjalmur89 Who will create the docs for "2"? I have created https://github.com/PX4/Devguide/issues/233 to track. I don't think it should block release of this doc.

I will move this note down to troubleshooting.

File: en/middleware/micrortps_throughput_test.md [hunk 41]
```diff
@@ -0,0 +1,105 @@
...
+            ++sent;
+        }
+    }
+     usleep(_options.sleep_ms*1000);
+    ++loop;
+}
+...
+```
+
+Compile and launch both the [client](../middleware/micrortps.md#client-px4-firmware) and the [agent](../middleware/micrortps.md#agent-off-board-fastrtps-interface).
```
hamishwillee, line 94: @vilhjalmur89  How do we compile the througput client after manual generation - now that we have disabled automatic generation, will the code for the Client still be built and uploaded if we build the firmware (as normal)? Or to put it another way, can we just point users at the normal instructions for building firmware?

en/middleware/micrortps_throughput_test.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Result
+
+The throughput can be observed in the client shell window.
+
+For PX4 running on Pixracer connected to an ordinary PC running Ubuntu 16.04, the results are:
```
hamishwillee, line 100: @vilhjalmur89 Are they connected over the USB/UART? What settings (is it important to say - IMO yes, particularly if this was just "default" settings for both client and agent)

en/middleware/micrortps_throughput_test.md
```diff
@@ -0,0 +1,105 @@
...
+
+```sh
```
hamishwillee, line 102: @vilhjalmur89 Is this output the normal/unmodified output for the client when it sends and receives messages (or do we need to add code for this output)?. 
When is it displayed - after the client exits? If so this implies that we have to start the Client for a certain number of loops.

en/middleware/micrortps_throughput_test.md
```diff
@@ -0,0 +1,105 @@
...
+SENT:     13255 messages in 13255 LOOPS, 3512575 bytes in 30.994 seconds - 113.33KB/s
+RECEIVED: 13251 messages in 10000 LOOPS, 3511515 bytes in 30.994 seconds - 113.30KB/s
```
hamishwillee, line 104: @vilhjalmur89 Where did this "received" data come from?
The example states that we simultaneously send and receive data. We modify the client to send data continuously, but we did no modification to the agent, nor did we create a sender RTPS application. So where are the received messages coming from?

hamishwillee:
@vilhjalmur89 I've updated the main bridge doc as you suggested, pulling out the "how toolchain works" docs other than information that is absolutely necessary. It is MUCH better. I this think one is good to publish.

I also updated the throughput doc - that has a few questions inline. Now to reintegrate the useful info back into the "manual" doc. 

Thanks for your advice!

hamishwillee:
@vilhjalmur89 I have now updated the "Manually generating" doc. This is nearly there. I plan to update it based on comments to the throughput doc. Specifically ad some next steps ...
1. Make it clear how you compile the end result (e.g. note that you can compile the manually generated client code in the same way as automatic generation)
1. Point to the troubleshooting and example in the main doc.

en/middleware/micrortps.md
```diff
@@ -0,0 +1,261 @@
...
+This topic describes the bridge architecture, how it is compiled, and how to write a simple FastRTSP application to subscribe to PX4 changes.
+
+
+## Architectural overview
+
+![basic example flow](../../assets/middleware/micrortps/basic_example_flow.png)
+
+The main elements of the architecture are the client and agent processes shown in the diagram above.
+
+- The *Client* is PX4 middleware daemon process that runs on the flight controller. It subscribes to uORB topics published by other PX4 components and sends any updates to the *Agent* (via a UART or UDP port). It also receives messages from the *Agent* and publishes them as uORB message on PX4.
```
ndepal, line 18: The messages on the serial link are in the uORB format?

hamishwillee: From a docs point of view it does not matter - as a user of this doc you just need to know that it is the contents of a "uORB" message. Additional detail would live in the "Manual generation" doc - but I question whether more detail is useful even there. 

That said, I am 90% sure they are a CDR serialized version of the uORB message, and that the further conversion to RTPS format happens in the agent.

hamishwillee:
@vilhjalmur89 @VicenteEprosima 
1. Did you see comments above/can you respond?

Looking at the work in https://github.com/PX4/Firmware/pull/7685 some of this will have changed - e.g. it looks like the send/receive list of uORB topics might have changed/moved. 

Also the docs were written to some extent as though FastRTPS/bridge was a prototype and this was early adopter docs. As a result you have things like "how to start the agent". I get the impression that now the bridge will be enabled by default as a daemon process, so some things should be slightly re-written with that focus. Thoughts?

en/middleware/micrortps.md
vicentemonge: The serialized information is not in uORB format. The CDR serializated format comes from the IDL type specification of the message, the same as Fast RTPS.

LorenzMeier:
@hamishwillee Can you work with @bkueng to just simplify / fix the docs as appropriate? Please do not try to retain all information or the original style - this started as describing a prototype and we need production documentation now. A lot of the internals have been automated.

@bkueng Please help to simplify this as needed.

hamishwillee:
@LorenzMeier Sure. The last iteration did a lot of work that gets this pretty close to the right level of information/style. Mostly we just need to make it clear how the client is started and any configuration for that.
@bkueng My take on what needs to happen is explained below. Can you give the docs a quick scan and see if this makes sense/what I'm missing. We could perhaps chat through this via skype/Slack?

----
* [RTPS/ROS2 Interface: PX4-FastRTPS Bridge](https://dev.px4.io/en/middleware/micrortps.html) - the entry point
   * Instructions are about the right level
   * Main problem is that they assume that the client has to be started manually. I presume that it is now started automatically as a server (please confirm). We should state that and ideally also explain how you can test link validity/that it has started (?)
   * Maybe move the following to the "manual generate code" doc - client syntax, troubleshooting section 
   * Perhaps add a publisher example (complementary to the listener example there now)
* [Manually Generate Client and Agent Code](https://dev.px4.io/en/middleware/micrortps_manual_code_generation.html) - This is the guide if you want to tweek/understand detail. I think it is very useful to have. Most people will not need. 
  * This basically has to reflect how the system now really works. I suspect it is mostly correct EXCEPT where [it explains the code lives that defines which uORB messages are published/subscribed](https://dev.px4.io/en/middleware/micrortps_manual_code_generation.html#supported-uorb-messages). 
  * Consider moving client syntax and troubleshooting docs here as in previous point - because most users won't need it.
* [Throughput test](https://dev.px4.io/en/middleware/micrortps_throughput_test.html) - example of how you can modify client code to measure throughput
  * See my comments against existing code - I THINK that this is single directional, but example tet states it is bidirectional. Need confirmation and either update description or code. I think this is a useful practical example of modifying client code. Not vital, but worth keeping.

I will also update "normal" Installation scripts. For Mac this will just be a note that FastRTPS is included in latest brew stuff. For Linux this will  probably be link to the FastRTPS instructions and also update "default" scripts I created to include FastRTPS.

@bkueng Does that make sense? If so, then what I most need is:
- Confirmation that the client starts automatically, and how that happens (parameters set by default, where it is tweeked)
- Ideas on best way for a user to verify that it started they are doing testing.
- Confirmation in the manual generation doc that this is still correct, in particular a) how the set of uorb topics is specified, b) Any initialisation setup of the client in firmware.
- Quick review/confirmation that the throughput example doesn't work as stated.

en/middleware/micrortps_throughput_test.md
vicentemonge: @hamishwillee When we generate the code manually we allow to compile and upload this code to the device in a 'normal' way but disabling the auto generation as is talled in the throughput test example.

en/middleware/micrortps_throughput_test.md
vicentemonge: Yes, is connected through USB/UART and the configuration is the default one (this include the UART communication).

en/middleware/micrortps_throughput_test.md
vicentemonge: This is the default displayed in the agent window when stop receiving messages for 2 seconds. This happens if the client ends its execution (limited number of loops) for example.

en/middleware/micrortps_throughput_test.md
vicentemonge: Since the agent is publishing this topic and subscribed to the same topic in the Fast RTPS side, is receiving the same information which then is sent back to the client.

vicentemonge:
@hamishwillee 
>  Did you see comments above/can you respond?

Done

> Looking at the work in PX4/Firmware#7685 some of this will have changed - e.g. it looks like the send/receive list of uORB topics might have changed/moved.

The send/receive list of uORB topics haven't changed.

> Also the docs were written to some extent as though FastRTPS/bridge was a prototype and this was early adopter docs. As a result you have things like "how to start the agent". I get the impression that now the bridge will be enabled by default as a daemon process, so some things should be slightly re-written with that focus. Thoughts?

Both applications the agent and the client are launched manually for now.

vicentemonge:
> - _Confirmation that the client starts automatically, and how that happens (parameters set by default, where it is tweeked)_

The client doesn't start automatically. The default values are listed [here in the doc](https://dev.px4.io/en/middleware/micrortps.html#client-px4-firmware)

> - _Ideas on best way for a user to verify that it started they are doing testing._

I think is the idea of this [part](https://dev.px4.io/en/middleware/micrortps.html#creating-a-fastrtps-listener-application)

 > - _Confirmation in the manual generation doc that this is still correct, in particular
a) how the set of uorb topics is specified,
b) Any initialisation setup of the client in firmware._

a) [this](https://dev.px4.io/en/middleware/micrortps_manual_code_generation.html#using-generatemicrortpsbridgepy) explains how to use the script and includes an example in the end of section.

I think this document mix things like this [section](https://dev.px4.io/en/middleware/micrortps_manual_code_generation.html#supported-uorb-messages) that explain where is placed the list of messages for **automatic** generation of the code and I think must be in the main doc micrortps.md.

> - _Quick review/confirmation that the throughput example doesn't work as stated._

The throughput test performs a measurement in bidireccional communications, it was planned to work in this way.

en/middleware/micrortps.md
hamishwillee: Thanks for this clarification @VicenteEprosima. I have not updated the text to state this because I believe that users don't need to know this level of detail, and what we have now is accurate: 

> "The *Agent* and *Client* are connected via a serial link (UART) or UDP network. The uORB information is [CDR serialized](https://en.wikipedia.org/wiki/Common_Data_Representation) for sending (*CDR serialization* provides a common format for exchanging serial data between different platforms)."

Let me know if you disagree, and I can talk more about the format.

en/middleware/micrortps_throughput_test.md
hamishwillee: @VicenteEprosima Interesting. Does that mean that every RTPS subscriber gets notified of its own publishing? ... and that uORB has similar dynamics so that it doesn't care about updates being reflected back?

vicentemonge: No. In this case the agent have a RTPS publisher (receiving from the link) and a RTPS subscriber (sending through the link) over the same topic so the info published is received for the subscriber.

This is a test for the bridge that includes a completelly new message, uORB doesn't know it so it's ignored. For a general case, all the info published over the same topic is received for all the subscribers of this topic. The publisher and the subscriber are different entities although can be in the same application.

en/middleware/micrortps.md
vicentemonge: Looks good for me. I think this is enough.

PR #221, opened 2017-07-17
px4tools has been renamed to apptools; updating documentation

jliphard:


hamishwillee:
@jliphard what is 'apptools' in that command line  - a package name somewhere? Apptools would seem to be a little "generic" if that is the case. Anyway, my concern with this is that the line before says "The recommended procedure is to use anaconda3. See px4tools github page for details.", but on the px4tools github page there is no mention of apptools.

hamishwillee:
@bkueng This one has been sitting around a while. Is this something you know about and could review?

bkueng:
I haven't used `conda` so far. If I go to https://anaconda.org/dronecrew/repo?type=conda&label=main, there is no package at all. However it is available via `pip`: https://pypi.python.org/pypi/px4tools/0.9.3, so I'd change the instructions to use that.

jgoppert:
Didn't know about the name change. If you want to take over px4tools maintenance,  can always transfer the repo to px4.

hamishwillee:
> Didn't know about the name change. If you want to take over px4tools maintenance, can always transfer the repo to px4.

Not a question for me. @bkueng  ?

@jliphard  Jan, can you please update in line with Beat's comment above: https://github.com/PX4/Devguide/pull/221#issuecomment-342439844 ?

hamishwillee:
@jliphard  Ping :-)

hamishwillee:
.. and again. Ping

LorenzMeier:
@jgoppert @bkueng If we depend on them would be best to maintain them upstream.

hamishwillee:
So basically the old version of the command does not work, but @jliphard version does seem to do "something" (see below). 

@bkueng not sure why the apptools package doesn't show up at your link, and I really have no way of verifying this is good. As you say the pip version does exist, and there are special instructions on how to install this on the px4 tools page.

My suggestion is that we remove the command line in this PR and just point to the instructions on px4tools github page. Those should work, because they cover install from source and using PIP.

The bigger question of owning maintenance of px4 tools etc should be managed outside this PR ... and I leave that to you guys. Thoughts?

```
ubuntu@ubuntu:~$ conda install -c https://conda.anaconda.org/dronecrew apptools
Fetching package metadata .............
Solving package specifications: .

Package plan for installation in environment /home/ubuntu/anaconda2:

The following NEW packages will be INSTALLED:

    apptools:  4.4.0-py27_0         
    configobj: 5.0.6-py27_0         
    pyface:    5.1.0-py27_0         
    traits:    4.6.0-py27_0         
    traitsui:  5.1.0-py27_0         

The following packages will be UPDATED:

    anaconda:  5.0.1-py27hd9359a7_1  --> custom-py27h4a00acb_0
    conda:     4.3.30-py27h6ae6dc7_0 --> 4.4.7-py27_0         
    pycosat:   0.6.2-py27h1cf261c_1  --> 0.6.3-py27ha4109ae_0 

Proceed ([y]/n)? Y

anaconda-custo 100% |################################| Time: 0:00:00   1.77 MB/s
pycosat-0.6.3- 100% |################################| Time: 0:00:00   1.56 MB/s
traits-4.6.0-p 100% |################################| Time: 0:00:00   1.84 MB/s
configobj-5.0. 100% |################################| Time: 0:00:00   2.80 MB/s
pyface-5.1.0-p 100% |################################| Time: 0:00:00   1.81 MB/s
traitsui-5.1.0 100% |################################| Time: 0:00:00   1.79 MB/s
apptools-4.4.0 100% |################################| Time: 0:00:00   1.90 MB/s
conda-4.4.7-py 100% |################################| Time: 0:00:01 766.96 kB/s
ubuntu@ubuntu:~$ 
```

bkueng:
> My suggestion is that we remove the command line in this PR and just point to the instructions on px4tools github page. Those should work, because they cover install from source and using PIP.

Agreed.

> The bigger question of owning maintenance of px4 tools etc should be managed outside this PR ... and I leave that to you guys. Thoughts?

I have not used it myself, but it's definitely a useful project. I see it's used [in CI mission tests](https://github.com/PX4/Firmware/blob/master/integrationtests/python_src/px4_it/mavros/mission_test.py#L51), but I'm not sure if it's actually still executed. Moving it to PX4 would increase its visibility and thus user base. @jgoppert @Stifael what do you think?

Stifael:
> Moving it to PX4 would increase its visibility and thus user base. 

agreed. Would be nice to have a standard procedure for more advanced mathematical analysis, and px4tools could be part of it.

hamishwillee:
OK, I am closing this as superceded by #432 (I subedited file as well). 

@Stifael @bkueng @jliphard Sounds like you're all in agreement to move it to PX4. Which of you owns that problem?

PR #213, opened 2017-07-06
Proposed update of toolchain docs

hamishwillee:
This attempts to improve our toolchain docs (particularly for Linux) in a few ways:
1. Some duplicated material has been separated into .txt files with _ prefer (`_somefile_whatever.txt`) so it can be imported where needed. E.g. QGroundControl and Editor instructions. This could also be at a higher level, but I think works well to be able to have in both the Mac/Linux instructions.
2. Better highlighting of the fact we selected Ubuntu LTS
3. Separation of Simulator dependencies so people can choose what to install.
4. Addition of easy to use bash scripts that people can use if they wish, at least for setting up simulators and NuttX.
5. Stop jumping around from Linux to "Advanced Linux" topic for basic setup.

I will add additional notes/questions inline.

PR #224, opened 2017-07-19
Simulation intro

hamishwillee:
This updates the Simulation introduction. The main aim is to provide context for users at the top level, explaining what simulators are supported and what sorts of testing each can be used for. Essentially helping them make a decision about what simulation to use.

Following that main introduction, there is additional information that gives users general overview about how simulation works (more intended as background reading, and to understand what can be modified). This could arguably be in a separate doc, but it is quite a useful overview.

I have also renamed sitl.md to jmavsim.md, since that is what it is about. This breaks some links, but I hope to avoid fixing them until #213 is merged.


en/simulation/README.md
```diff
@@ -0,0 +1,187 @@
...
+
+Information about available simulators and how to set them up are provided in the next section. The other sections provide general information about how the simulator works, and are not required to *use* the simulators. 
+
+
+## Supported Simulators
+
+The following simulators work with PX4 for HITL and/or SITL simulation.
+
+Simulator |Description
+---|---
```
hamishwillee, line 17: Is the "supported" list of vehicles correct for each simulator?

en/simulation/README.md
```diff
@@ -0,0 +1,187 @@
...
+mavlink stream -r 20 -s RC_CHANNELS -u 14556
+mavlink stream -r 250 -s HIGHRES_IMU -u 14556
+mavlink stream -r 10 -s OPTICAL_FLOW_RAD -u 14556
+logger start -e -t
+mavlink boot_complete
+replay trystart
+```
+
+
+## HITL Simulation Environment
```
hamishwillee, line 165: @dagar Can you check this section in particular? Is the diagram correct? The text suggests that HITL uses a normal build in which you set a configuration that doesn't start any sensors (presumably via QGC) - which indicates that configuration is a matter of changed parameters.
BUT there are specific HIL airframes - which indicates that there is some configuration that needs to be made in the build. I don't think both can be true. 
IF the build is "special", a) how do we invoke those builds, b) what is it that the config does differently from "normal" builds.

en/simulation/jmavsim.md
```diff
@@ -1,4 +1,17 @@
-# Software in the Loop (SITL) Simulation
+# jMAVSim with SITL
```
hamishwillee, line 1: FYI, the changes in this file are not supposed to be exhaustive. They are just enough to change the focus from "SITL" to "jMAVSim"

LorenzMeier:
Awesome!

en/simulation/README.md
dagar: The specific HIL airframes skip starting the real sensor drivers and then start the sensors and commander modules in HIL mode from the beginning (they can be switched into HIL later on as well).

hamishwillee: @dagar Awesome. 

 > The specific HIL airframes skip starting the real sensor drivers and then start the sensors and commander modules in HIL mode from the beginning

1.  Is that all done/caused by adding `set HIL yes` to the init file (e.g. as in [1001_rc_quad_x.hil](https://github.com/PX4/Firmware/blob/master/ROMFS/px4fmu_common/init.d/1001_rc_quad_x.hil#L15)
2. Following on from that, if I added `set HIL yes` to another (non hil) init file like [10015_tbs_discovery](https://github.com/PX4/Firmware/blob/master/ROMFS/px4fmu_common/init.d/10015_tbs_discovery) is it now a HIL file - ie can I build the custom file and upload it to firmware and expect it to work like another HIL target, or do I need to specifically enable/disable other things?

> (they can be switched into HIL later on as well).

3. How? I think what your'e saying here is that perhaps I can do something to make a running firmware work in HIL mode. Is this a parameter setting?

dagar: 1. yes that's how the startup handles it.
2. For the most part yes, but I believe there are builtin assumptions about mixer setup to actually use things like x-plane and jmavsim HIL.
3. HIL is a flag used here - http://mavlink.org/messages/common#HEARTBEAT (see MAV_MODE_FLAG_HIL_ENABLED). 

Side question, are the mavlink message definitions in one of the gitbooks now? I couldn't find them.

PR #226, opened 2017-07-20
Update for airmind products

flynxea:
Update for airmind products
1) Update MR210 summary
2) Nanomind Flight Video


airmind:
@dagar thanks

PR #234, opened 2017-07-31
Update tuning_the_ecl_ekf.md

elton-choi:
Modify Drag Specific Forces description.
bit position 4 in the EKF2_AID_MASK should be modified to bit position 5

### DIFF
--- a/en/tutorials/tuning_the_ecl_ekf.md
+++ b/en/tutorials/tuning_the_ecl_ekf.md
@@ -74,7 +74,7 @@ Fixed wing platforms can take advantage of an assumed sideslip observation of ze
 
 ### Drag Specific Forces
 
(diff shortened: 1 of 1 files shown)


hamishwillee:
Bit position 5 in the [EKF2_AID_MASK](https://dev.px4.io/en/advanced/parameter_reference.html#EKF2_AID_MASK) is undocumented - so if this is correct we should also [ekf2/ekf2_params.c](https://github.com/PX4/Firmware/blob/master/src/modules/ekf2/ekf2_params.c#L554) and regenerate the parameters. 
@priseborough - thoughts?

priseborough:
This is included in https://github.com/PX4/Firmware/pull/7264

hamishwillee:
Thanks. Then we can merge as soon as that goes in.

LorenzMeier:
Awesome, thanks all!

PR #236, opened 2017-08-01
Fixed the display problem of the third flow chart

Boyang--Li:
Change the original Tab retraction to Space so that the flow chart can be displayed well on the web.


mhkabir:
Can you send a screenshot of what this renders as now?

Boyang--Li:
The previous webpage looks like this:
ori
I think it should be like this:
after1

Can you help me to delete two more spaces in line 65 and 70 of my pulled request like the figure below then it will be right. Thanks!

after

PR #238, opened 2017-08-03
Micrortps doc update

vicentemonge:


vicentemonge:
@hamishwillee please review that.

en/middleware/micrortps.md
```diff
@@ -34,6 +33,25 @@ The *Client* application is also compiled and built into the firmware as part of
 > **Tip** The bridge code can also be [manually generated](micrortps_manual_code_generation.md). Most users will not need to do so, but the linked topic provides a more detailed overview of the build process and can be useful for troubleshooting.
 
 
+## Supported uORB messages
```
hamishwillee, line 36: I don't think we should move this section. A normal developer who wants to use RTPS does not need this information to build and use the bridge - because the standard set is already defined. This information would be useful as "additional detail", which is why it was in the other document

en/middleware/micrortps_manual_code_generation.md
```diff
@@ -83,15 +64,15 @@ $ python Tools/generate_microRTPS_bridge.py -s msg/sensor_baro.msg -r msg/sensor
 
 Code is generated for the *Client*, *Agent*, *CDR serialization/deserialization* of uORB messages, and the definition of the associated RTPS messages (IDL files). 
 
-Code for the bridge can be found here:
+Code for the bridge by default in the manual generation can be found here:
```
hamishwillee, line 67: Is this true for just manual generation, or for manual and auto generation?

en/middleware/micrortps.md
vicentemonge: The standard set is not defined yet and is needed to be specified.

en/middleware/micrortps_manual_code_generation.md
```diff
@@ -83,15 +64,15 @@ $ python Tools/generate_microRTPS_bridge.py -s msg/sensor_baro.msg -r msg/sensor
...
-- *Client*: **build\__BUILDPLATFORM_/src/modules/micrortps_client/**
-- *Agent*: **build\__BUILDPLATFORM_/src/modules/micrortps_bridge/**
+- *Client*: **src/modules/micrortps_bridge/micrortps_client/**
+- *Agent*: **src/modules/micrortps_bridge/micrortps_agent/**
 
 
 ### uORB serialization code
 
-Serialization functions are generated for the [supported uORB topics](#supported-uorb-messages). For example, the following functions would be generated for the *sensor_combined.msg*:
+Serialization functions are generated in the PX4 normal compilation process for all the uORB topics. For example, the following functions would be generated for the *sensor_combined.msg*:
```
hamishwillee: So just to clarify, the serialisation functions are build for all uORB topics in all cases (manual and auto)? But only some are then called/included by the client/agent?

en/middleware/micrortps_manual_code_generation.md
```diff
@@ -100,7 +81,7 @@ void deserialize_sensor_combined(struct sensor_combined_s *output, char *input,
 
 ### RTPS message IDL files
 
-IDL files are generated from the uORB **.msg** files ([for supported uORB topics](#supported-uorb-messages)). These can be found in: **build\__BUILDPLATFORM_/src/modules/micrortps_bridge/idl/**
+IDL files are generated from the uORB **.msg** files for selected uORB topics in the generation of the bridge. These can be found in: **src/modules/micrortps_bridge/micrortps_agent/idl/**
```
hamishwillee: So why not link to the "supported uorb messages"? Is there some other selection process for selecting the topics?

en/middleware/micrortps_manual_code_generation.md
```diff
@@ -120,22 +101,22 @@ src/modules/micrortps_bridge/micrortps_agent
 ├── idl
 │   ├── sensor_baro_.idl
 │   └── sensor_combined_.idl
-├── microRTPS_agent.cxx
-├── microRTPS_transport.cxx
+├── microRTPS_agent.cpp
```
hamishwillee, line 104: All these look good. For my info only, why was the text saying .cxx?

en/middleware/micrortps_manual_code_generation.md
vicentemonge: This path is the default for manualcreation. In the automatic generation the generated client code ends on build_PLATFORM/src/modules/micrortps_bridge/modules_client/

hamishwillee:
I've added some comments. Main issue is location of the list of uorb topics. I'd have said that was a fixed set for most users, so is not something in the main doc. This is arguable of course, but I've tried to separate into "essential for most users" and "extra stuff you might need to do to customise"

I might have more to say based on your responses :-)

en/middleware/micrortps_manual_code_generation.md
vicentemonge: These are created in the PX4 compilation process with or without generation of the other code for all the topics. The client/agent only uses what is needed.

en/middleware/micrortps_manual_code_generation.md
vicentemonge: The extension was changed a couple of days ago.

en/middleware/micrortps.md
hamishwillee: If this is going to happen in the coming weeks OR if this is 90% complete and we're only deciding on some outliers then I'd leave it as I had it - and just add a link to the existing location. The reason being that it is good to keep the docs as close to how it "should" eventually look. 

If everyone is going to have to edit this list, then you're quite right to move it. Which is true?

en/middleware/micrortps_manual_code_generation.md
vicentemonge: There are two ways that depend on you generate manually or automatically. Manually by the script and --send and --receive options, automaticlly through the .cmake files.

en/middleware/micrortps_manual_code_generation.md
hamishwillee: Thanks. 
IF the default location is  NOT editable/changeable this text should say:
> Manually generated code for the bridge can be found here:

IF the default is really a default (i.e. it is something that has a value you can change using the tool), then this text should say. 
> Manually generated code for the bridge can be found here (by default):

en/middleware/micrortps_manual_code_generation.md
hamishwillee: THanks for that clarification. But please confirm the functions also generate when we don't use the normal compilation process - ie we use manual process?

en/middleware/micrortps_manual_code_generation.md
vicentemonge: 2nd one

en/middleware/micrortps.md
```diff
@@ -34,6 +33,25 @@ The *Client* application is also compiled and built into the firmware as part of
...
+   # Add new topic...
+   )
+
+set(config_rtps_receive_topics
+   vehicle_command
+   # Add new topic...
+   )
+```
+
+> **Note** You can further refine which messages are supported when you call **generate_microRTPS_bridge.py** using `-s` (send) and `-r` (receive) flags. Please see [Manual Generation of the Code](../middleware/micrortps_manual_code_generation.md).
```
hamishwillee: OK, so based on what you say below this is wrong. We need to convey that the lists are used to specify the set for auto generation and that the arguments are used for manual generation (what it says now is that the arguments "filter down" the list defined in the txt file). 

I would do this as a single sentence at the very top (before what is there now).

en/middleware/micrortps_manual_code_generation.md
vicentemonge: The manual part of the process is only for code generation the other part is the same and for that reason are created at the same way.

en/middleware/micrortps_manual_code_generation.md
hamishwillee: I would leave this as a link, but make it say "selected" rather than supported (as you have done). To make this work the linked topic will have to make it clear that there are 2 ways to define the list of topics (manual via parameter and auto via the lists)

en/middleware/micrortps_manual_code_generation.md
hamishwillee: Thank you.

en/middleware/micrortps.md
vicentemonge: I'm not sure of this. I think should point to the current state.

en/middleware/micrortps_manual_code_generation.md
hamishwillee: OK, in that case I might change this as below (essentially clarifying that it happens for all forms of generation). 

> Serialization functions are generated for **all** the uORB topics as part of the normal PX4 compilation process (and also for manual generation).

en/middleware/micrortps.md
hamishwillee: We should certainly explain the current state - and we should add a note saying that the list is in flux. But where the information is located should be based on whether users of the docs will rarely or always need the information. It sounds like you think most users will want to customise the list at this time?

hamishwillee:
I think I've responded to all your points. I'm off to bed now, but I'll check this again in the morning. Thanks for all the clarifications - nice to work with you in real time!

One last point, do we know when the work to make the client be auto-started will be done?

vicentemonge:
Thanks for your comments!!

I auto-start client is not planned for now.

en/middleware/micrortps.md
vicentemonge: Yes, now is very needed.

hamishwillee:
@VicenteEprosima Awesome. I've merged this. I'll do my own updates as a post-process.

PR #241, opened 2017-08-04
RTPS Docs updates

hamishwillee:


en/middleware/micrortps.md
```diff
@@ -35,7 +35,9 @@ The *Client* application is also compiled and built into the firmware as part of
 
 ## Supported uORB messages
 
-The set of uORB topics that can be supported by the bridge are listed in the **msg/CMakeLists.txt**. You can modify the list to change what uORB messages can be published/subscribed using RTPS in the **.cmake** file (**cmake/configs**) for each target platform.
```
hamishwillee, line 38: @VicenteEprosima I modified this because **msg/CMakeLists.txt** does not include the set of topics - those are in the cmake files. 
From what I can see the only people who need to know about CMakeLists.txt are those that are changing  our support for RTPS (so for example, this might be where we would define a default list of uORB topics for all targets). Do you think we need to still say something about what this file is for?

en/middleware/micrortps.md
```diff
@@ -44,12 +46,14 @@ set(config_rtps_send_topics
...
 
 set(config_rtps_receive_topics
-   vehicle_command
+   sensor_baro
    # Add new topic...
    )
 ```
 
-> **Note** There is another way to specify which messages are intended to be used calling **generate_microRTPS_bridge.py** using `-s` (send) and `-r` (receive) flags. Please see [Manual Generation of the Code](../middleware/micrortps_manual_code_generation.md).
+> **Caution** At time of writing (August 2017), only the small set of uORB topics listed above are included in our cmake files: **posix_sitl_default.cmake**, **nuttx_px4fmu-v4_default.cmake**, **posix_sdflight_default.cmake**. It is likely you will need to edit your *cmake* file and add additional uORB topics. In future we hope to define a larger standard set.
```
hamishwillee, line 54: @VicenteEprosima This is the note we discussed making it clear that they almost certainly WILL need to edit topics. Can you confirm that adding these defaults means that **all** nuttx, sitl and sdflight targets will inherit these defaults?

en/middleware/micrortps.md
```diff
@@ -89,6 +92,8 @@ By default the *Client* runs for 10000 loops and then stops. To run the *Client*
 micrortps_client start -l -1
 ```
 
+> **Note** The PX4 Firmware initialisation code may in future automatically start the *Client* as a permanent daemon process. In the meantime you will need to start the client manually.<!-- at that point, most of this section would move into the "manual generation" doc: https://github.com/PX4/Firmware/pull/7663#issuecomment-317928506 -->
```
hamishwillee, line 95: @VicenteEprosima FYI, you stated that there is no plan to make starting the client automatically. This modification changes the statement that "in future we will do this" to "this is something we may do"

en/middleware/micrortps_throughput_test.md
```diff
@@ -93,11 +93,13 @@ while (!_should_exit_task)
...
 
+> **Note** You may recall this is a *bidirectional* throughput test, where messages must also be sent from the *Agent* to the *Client*. The good news is that you do not need to modify the Agent code to make this happen. As the *Agent* is an RTPS publisher and subscriber, it will automatically get notified of the RTPS messages it sends, and will then transfer these back to the client.
+
 ## Result
 
-The throughput can be observed in the client shell window.
+The test was executed with PX4 running on Pixracer, connected via a UART to an ordinary PC running Ubuntu 16.04. The default configuration was used for both the Client/Agent configuration.
 
-For PX4 running on Pixracer connected to an ordinary PC running Ubuntu 16.04, the results are:
+The throughput that was observed in the client shell window is shown below:
```
hamishwillee: @VicenteEprosima  - you said in #219 "This is the default displayed in the *agent* window when stop receiving messages for 2 seconds. This happens if the client ends its execution (limited number of loops) for example."

So just to be very clear, "in the client shell window" as per the current text is incorrect right?

en/middleware/micrortps_manual_code_generation.md
```diff
@@ -133,3 +133,7 @@ src/modules/micrortps_bridge/micrortps_client
 └── microRTPS_transport.h
  0 directories, 4 files
 ```
+
+## Build and use the code
```
hamishwillee, line 137: @VicenteEprosima Can we expand on this a bit. In #210 you said that once automatic code generation is disabled you can just use "normal build processes" for the client and agent, BUT in the normal build process we don't build the client (it is done for us). I don't think just running the normal build process will do the compile set for us because the  manually generated source code is not located at the same location as the automatically generated source code. 

1. How do we build the client manually? (command line we enter please - and does it differ for SITL vs nuttx vs snapdragon?)

It does seem likely that the firmware build process will automatically generate the firmware and include the client executables IFF they are present in the normal place - which they should be after manually building.

2. Can you confirm?

3. Do we have to do anything different for building the agent other than use the source from a different location?

hamishwillee:
@VicenteEprosima 
This update includes corrections and questions related to our previous discussions.

@LorenzMeier @bkueng I have also updated the installation sections for FastRTPS
- scripts for Ubuntu and *Bash for Windows*  include FastRTPS 1.5 from binaries. 
- ubuntu "manual" instructions show the change, and link to the main install for RTPS topic
- The FastRTPS install guide was updated so that it now has a topic (first one you get to) explaining that FastRTPS is already installed on Mac, Ubuntu Linux, Bash for Windows. The instructions are now billed as being for helping you on other platforms. This is good as the scripts are far more turnkey than this guide.

@bkueng Can you please sanity check the whole FastRTPS usage story after these changes are in? There is certainly more that *could* be done, but I think this is enough for people to easily onboard and understand what is going on.

en/middleware/micrortps.md
vicentemonge: It seems right, now is simpler.

en/middleware/micrortps.md
vicentemonge: This is not decided for now. There is an open discussion over the set of messages and I think this will be included and resolved in the near future.

vicentemonge (approved):

en/middleware/micrortps_throughput_test.md
vicentemonge: I was wrong, the info is from the client.

en/middleware/micrortps_throughput_test.md
```diff
@@ -93,11 +93,13 @@ while (!_should_exit_task)
...
 Compile and launch both the [client](../middleware/micrortps.md#client-px4-firmware) and the [agent](../middleware/micrortps.md#agent-off-board-fastrtps-interface).
 
+> **Note** You may recall this is a *bidirectional* throughput test, where messages must also be sent from the *Agent* to the *Client*. The good news is that you do not need to modify the Agent code to make this happen. As the *Agent* is an RTPS publisher and subscriber, it will automatically get notified of the RTPS messages it sends, and will then transfer these back to the client.
+
 ## Result
 
-The throughput can be observed in the client shell window.
+The test was executed with PX4 running on Pixracer, connected via a UART to an ordinary PC running Ubuntu 16.04. The default configuration was used for both the Client/Agent configuration.
 
-For PX4 running on Pixracer connected to an ordinary PC running Ubuntu 16.04, the results are:
```
vicentemonge: For throughput measurement is very important talk about the environment since there are significant variations. I would include a note below with the hardware info.

vicentemonge (changes requested):
Good job. Only I think we need to add a comment over throughput results.

en/middleware/micrortps_throughput_test.md
hamishwillee: I don't know what the hardware is - it was you guys testing :-) If you can provide that would be great. (though I agree your point).

en/middleware/micrortps.md
hamishwillee: Thanks. My point is that I *THINK* that the settings you put in default files are inherited by all other configurations. Perhaps @dagar could confirm?

en/middleware/micrortps_manual_code_generation.md
hamishwillee: This is the most important bit to answer :-)

en/middleware/micrortps_throughput_test.md
vicentemonge: _PX4 running on Pixracer connected to an ordinary PC running Ubuntu 16.04_ I think is enough

en/middleware/micrortps_manual_code_generation.md
vicentemonge: 1. After a manual generation you can run _make PLATFORM_ for compile as usual and the client is compiled.

vicentemonge: 2. The client as binary is ever generated in the same place.

vicentemonge: 3. I don't understand you very well. Can you be more specific?

hamishwillee: I have updated this section completely for your review (rather than clarify, you can just check what is now written makes sense).

hamishwillee:
@VicenteEprosima I'm merging this now as it is a net improvement. What  could still be usefully provided by you (as per discussions above) is detail on the test setup you used to create the througput results. 

@bkueng Can you please sanity check the whole FastRTPS usage story now? There is certainly more that *could* be done, but I think this is enough for people to easily onboard and understand what is going on.

bkueng:
This looks good. One thing for the manual installation from source: in addition to java, I had to install `gradle` to make it compile.

I would add a section FastRTPS vs. mavlink: describing that FastRTPS is not a replacement for mavlink, mavlink will still be used e.g. for communication with ground stations (and gimbals, cameras, ...), it's a long-term stable API. FastRTPS is not intended to be used for slow links, such as radio telemetry...

hamishwillee:
@bkueng Thank you. What platform did you need gradle on? If it was Windows, how do you set up your Windows env - the standard build tool package didn't work for me. 

Agree re the FastRTPS vs MAVLink. I'll add something for that.

bkueng:
> What platform did you need gradle on?

Linux, Fedora

hamishwillee:
>> What platform did you need gradle on?

> Linux, Fedora

Annoying. To know for sure where it needs to be installed I'd have to try a bunch of platforms. I checked the source was created, but not that it built.

@bkueng I didn't get any errors in build on Ubuntu. How can I verify compilation was successful all the way to firmware - is there a list of what was built into the firmware, or where can I find the binaries (what is filename). I did a search on micrortps_client but nothing I could be sure was an executable popped up.

@VicenteEprosima Is this a general dependency that should be added to the list. If not, could we get away by adding a note e [in installation run dependencies section](https://dev.px4.io/en/setup/fast-rtps-installation.html#run-dependencies) as below 

> **Note** You may also need to explicitly [install gradle](https://gradle.org/install/) on some platforms (e.g. Fedora Linux).

EDIT. I did add this comment. See #252 change below.

en/middleware/micrortps_manual_code_generation.md
vicentemonge: Looks good, thanks for your work!

vicentemonge:
@hamishwillee Thanks for add the information to the docs. I think this enough.

en/middleware/micrortps.md
dagar: No these aren't really inherited except for posix_sitl_default by shared by posix_sitl_{broadcast, ekf2, inav, lpe, shell, test} and sdflight shared by eagle and excelsior. I suppose that sounds like the opposite of what I said, but the posix_sitl_default setup is more of a hack until there's a better init system.

hamishwillee: Thanks Daniel. That helps me understand things. Nothing changes here as a result (I didn't note that there was inheritance). This section will need to change if/when uORB "standard set" is agreed.

PR #243, opened 2017-08-04
Add instructions to connect QGC to docker

martinlowinski:


LorenzMeier (approved):

hamishwillee:
This is excellent. 

@martinlowinski I've got a Windows PC, so I run a Ubuntu VM and run docker in that (at least for testing). Do you happen to know how/if you can share from docker all the way out to the "real" (Windows) host?

martinlowinski:
Although I haven't tested this yet, there should be two options (e.g. for VirtualBox):

- Bridged Networking: When in bridged mode, the virtual machine will join the same network as the host and have its own separate IP address. Then you can access the container via this IP address and the exposed port of the container.
- Port forwarding: When the VM should stay in NAT mode, you can add a port forwarding rule. It's important here to leave the IP address fields empty. Then access the container via localhost and the exposed port.

hamishwillee:
Thanks @martinlowinski . I think I have some experimentation to do.

martinlowinski:
I can confirm that it is working with bridged networking. The QGC von my Android phone could connect to the simulation running inside the docker container of the VM. However, the container has to expose the port for that: `docker run ... -p 14557:14557`

What I don't understand: Once you expose the port, you cannot connect to the container from within the VM (address already in use). For that you have to start the container without exposing the port but then you cannot connect from outside of the VM.

hamishwillee:
I wonder if it might be possible to expose multiple ports? @dagar Any ideas? The context is "how to connect to docker hosted in VM hosted on windows, from Windows"

dagar:
Forward by default here? https://github.com/PX4/containers/blob/master/docker/px4-dev/Dockerfile_base#L69

martinlowinski:
If you want to connect to docker only from your Windows host, then these two ports (14556, 14557) are exposed by default but have to be mapped.

```
xhost +
docker run -it --privileged \
    -v <local_src>:<container_src>:rw \
    -v /tmp/.X11-unix:/tmp/.X11-unix:ro \
    -e DISPLAY=:0 \
    -p 14556:14556/udp \
    -p 14557:14557/udp \
    --name=px4-dev-simulation px4io/px4-dev-simulation /bin/bash
```

This should map the two exposed ports to the same ports on your VM. Then you can connect to it from your Windows host either by using the VM's IP address (bridged networking) or localhost (NAT port forwarding).

For more ports you have to change the Dockerfile, rebuild the image and map the ports accordingly.

PR #244, opened 2017-08-05
update the standard plane output channel

oneWayOut:
the latest firmware has deleted  SYS_AUTOSTART id 2101, and changed the throttle channel from MAIN3 to MAIN4.
this guide need to be updated.

### DIFF
--- a/en/airframes/airframe_reference.md
+++ b/en/airframes/airframe_reference.md
@@ -722,14 +722,10 @@ This page lists all supported airframes and types including
    <tr><th>Name</th><th></th></tr>
  </thead>
 <tbody>
-<tr id="plane_standard_plane_multiplex_easystar">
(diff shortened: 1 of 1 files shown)


LorenzMeier:
Thanks!

hamishwillee:
@oneWayOut Thanks very much for this. FYI, this document is build from source, so the "best" way to update it (if you can) is to re-run the script. That way you catch not just the change you spotted, but any other changes that might have been added (see #245). 
It is pretty easy to do this. You just get the latest github px4/Firmware, and in /Tools, run:
```
python px_process_airframes.py -m
```

@dagar Any further progress/ETA on automating update of these docs?

dagar:
Sorry, haven't had a chance to get to it.

PR #245, opened 2017-08-06
Update airframes ref

hamishwillee:
Generated ref again from master source.

### DIFF
--- a/en/airframes/airframe_reference.md
+++ b/en/airframes/airframe_reference.md
@@ -39,6 +39,34 @@ This page lists all supported airframes and types including
 </tr>
 </tbody></table>
 
+### Dodecarotor cox
+
+<div>
+<table style="float: right; width: 70%; font-size:1.5rem;">
+ <colgroup><col></colgroup>
+ <thead>
+   <tr><th>Common Outputs</th></tr>
+ </thead>
+<tbody>
+<tr>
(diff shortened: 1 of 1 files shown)


en/airframes/airframe_reference.md
```diff
@@ -39,6 +39,34 @@ This page lists all supported airframes and types including
 </tr>
 </tbody></table>
 
+### Dodecarotor cox
```
hamishwillee, line 42: @bkueng There doesn't appear to be any image for this vehicle. Do you think we can get a build guide or other information?

hamishwillee:
@bkueng Also, is it OK to build this off "master"? Ie obviously master leads the actual released code, so might differ from what people have on their vehicle.

en/airframes/airframe_reference.md
bkueng: It was recently added in https://github.com/PX4/Firmware/pull/7532.
@imcnanie if you could add a build log of the vehicle to the dev-guide here that would be great!

bkueng:
> @bkueng Also, is it OK to build this off "master"? Ie obviously master leads the actual released code, so might differ from what people have on their vehicle.

Yes the information in the dev-guide is generally in sync with master. At some point we might want to introduce versioning.

en/airframes/airframe_reference.md
hamishwillee: @bkueng I mean a rotor-order diagram as per all the other copters in the airframe docs. As far as I can tell, not added in that PR. It is possible I am being dim!

bkueng: You're right, that does not exist yet. I'll see if I can get it done, I'll need to check the motor assignments & directions.

PR #247, opened 2017-08-07
Added QAV-R 5" KISS ESC build

MaEtUgR:


MaEtUgR:
Can I look at the final GitBook that would result from this branch such that I can check all internal links and so on before merging?

MaEtUgR:
I managed to make a build of the branch but only using a fork...
https://maetugr.gitbooks.io/px4-devguide-my-draft/content/en/airframes_multicopter/qav-r-5-kiss-esc-racer.html
There are still errors. I'll push the changes as soon as I fixed them.

hamishwillee:
Hi @MaEtUgR,

It is possible to set up gitbook locally and build (you need node and a few other things): https://toolchain.gitbook.com/setup.html
This can be fairly slow for the Devguide, and is painful on Windows. I do like what you've done, because we can see it :-)

A few hints:
1. If you want to have multiple tips/notes in a row you need to separate them with spans (bug I think)
```
> **Tip** This is tip 1

<span></span>
> **Note** This is a note
```

For config instructions I like to link to https://docs.px4.io/en/config/. Then I state anything different that I did.  ON that note, any reason you chose to use Master? Also include any special tuning required for vehicle. I'll check it out properly when you say it is ready.

MaEtUgR:
Created a follow up https://github.com/PX4/Devguide/pull/250 because GitBook only created a preview from my fork... it went into infinite loop for this upstream branch.

PR #248, opened 2017-08-08
Add section explaining when to use FastRTPS

hamishwillee:
@bkueng This attempts to address your comment: https://github.com/PX4/Devguide/pull/241#issuecomment-320844545

I decided to add a section that was "RTPS" centric, (rather than comparative with MAVLink) because  that seemed to fit better. I did still add the note about MAVLink, but as a note.

I was fairly "hand wavy" about when you'd use FastRTPS. Are there any other specific cases we have in mind?


bkueng:
Looks great. I see tasks like vision and path planning as the primary use-cases too, but I'm sure people will be creative and come up with more.

PR #255, opened 2017-08-13
Add pyserial to common dependencies to support firmware upload

hamishwillee:


hamishwillee:
Instructions didn't work. Grrr. Will resubmit.

PR #257, opened 2017-08-15
Add document of linux_sbus driver

crossa:
Yesterday I've make a pull request of linux_sbus driver. https://github.com/PX4/Firmware/pull/7785
This driver require some  electronic component so I  write this document to help users to use this driver

PR #263, opened 2017-08-21
Linux S.Bus driver updates

hamishwillee:
This makes a number of updates for English and rendering. The two "important" ones are:
1. Images are now local. This means that if @crossa deletes his branch, we can still load them. They have sensible names in the tree.
2. The signal inverter circuit is now a clear "subsection" - ie all bits related to that are under a single section.

@crossa @TSC21, can you please sanity check. This is a subedit against https://github.com/PX4/Devguide/pull/257.


hamishwillee:
P.S. Note, we still need to have an updated circuit diagram with labels.

crossa:
@hamishwillee  That's ok and I think a mybe a PCB board without any compnent  is necessary for users to use  build thire Quadocopter with raspberry pi  just like building blocks. If there is a PCB board, then users  can use the pi zero to instead pi3.  I used to use Pi0 to build the autopilot but it’s too small that it is very difficult to make the circuit manually.

hamishwillee:
PCB board would be excellent.

TSC21:
@hamishwillee the changes look good and clear. Thanks.

> PCB board would be excellent.

I suppose I can help out with that on Altium. But should we merge this as is and then add the PCB files when able?

TSC21:
Also @crossa should make the the Chinese version match the English version.

hamishwillee:
Thanks guys. Merged. I've created https://github.com/PX4/Devguide/issues/268 so you can track further needed work.

PR #265, opened 2017-08-21
setup/simulation: opencv moved to brew core

julianoes:
Opencv is no longer in its own tap but moved to brew core, therefore we
can install it directly from there.


hamishwillee:
Thanks very much.

PR #258, opened 2017-08-17
Including FlytOS in Devguide

peaboard:
Request to include FlytOS under companion computers of the Devguide.
For reference please visit [flytbase](https://flytbase.com/).

PR #270, opened 2017-08-22
Request to add FlytBase to PX4Devguide under Companion Computers

peaboard:
Due to issues with the last [PR](https://github.com/PX4/Devguide/pull/258) have closed it. Please review this request instead. 
Thanks and sorry for the inconvenience.


LorenzMeier:
I'm closing this as this essentially describes a company software product which is not part of Dronecode. The difference to airframes or autopilots (which are also products) is that we need those to run our software on. The same applies for data sources like the Airmap elevation API. In contrary your product seems to just use Dronecode and not enable it.

You are using the IRIS airframe model in FlytSim, which is originating from this repository: PX4/sitl_gazebo

Where are you attributing the usage of this model in your documentation as required by the Apache 2.0 license?
http://www.apache.org/licenses/LICENSE-2.0

PR #272, opened 2017-08-24
Remove airframes migrated to userguide

hamishwillee:
This removes the migrated airframes. It also updates the entry point so that it makes sense as the remaining airframes disappear ....


hamishwillee:
@LorenzMeier I am merging this - the files being deleted have already been moved, and the "high level" changes make it easier to make atomic/more reviewable changes going forward.

PR #275, opened 2017-08-25
Include micrortps client as daemon

vicentemonge:
@hamishwillee we have modified the client as a daemon but still is needed be launched manually. Also we have erased the raspberry pi logo from the architecture image.

Thanks!


hamishwillee:
Looks good thanks. I fixed a minor typo and merged.

vicentemonge:
Great, thanks!

PR #276, opened 2017-08-28
uLanding: add usage and support for posix platform

jgs2185:


TSC21:
It needs to be clarified if the uLanding is capable of running on other POSIX targets.

en/sensor/uart_ulanding_radar.md
```diff
@@ -20,8 +20,9 @@ You can simply add the following line to an [extras.txt](../advanced/system_star
 ulanding_radar start /dev/serial_port
 ```
 
-In the upper command you will have to replace the last argument with the serial port you have connected the hardware to.
-If you don't specify any port the driver will use /dev/ttyS2 which is the TELEM2 port on Pixhawk.
+In the above command you will have to replace the last argument with the serial port you have connected the hardware to.
+If you don't specify any port the driver will use /dev/ttyS2 which is the TELEM2 port on Pixhawk. If either of the OcPoc config
+files are used (posix_ocpoc_cross.cmake or posix_ocpoc_ubuntu.cmake) the default port will be /dev/ttyS6.
 
 **Warning**
```
hamishwillee: Best way to do warnings is to use the syntax like below (this wraps the text in a very visible box).

```
> **Warning** This is the text in the warning ...
```

TSC21:
@jgs2185 can you please replace https://github.com/PX4/Devguide/pull/276/files#diff-89fd8d228a25ab8de4178a5d6bdf69fbR30 with @hamishwillee suggestion: `> **Warning** This is the text in the warning ...`

jgs2185:
@TSC21 I did this with the last commit

TSC21:
You are missing `>` I believe @jgs2185

hamishwillee:
Hi @TSC21 - for some reason Github doesn't render it, but you can see the '>' is present if you view the raw source. So @jgs2185 is correct :-)

TSC21:
Where? https://raw.githubusercontent.com/jgs2185/Devguide/cb434bf51371ee3f4d904cddcd77b0ab2e55d454/en/sensor/uart_ulanding_radar.md

TSC21:
Shouldn't it be:
```

> **Warning** If you are connecting the radar device to TELEM2 then make sure to set the parameter SYS_COMPANION to 0. Otherwise the
serial port will be used by a another application and you will get unexpected behaviour.
```

hamishwillee:
@TSC21 YES - sorry, I must be remembering another example of this I checked. Interestingly though, the rendered email with your comment above did not include the leading > , so it is quite possible that @jgs2185 fixed it based on the email. Thanks for checking!

jgs2185:
Sorry all, I didn't see the > initially. It is included now.

TSC21 (approved):
lgtm

hamishwillee:
Merged. Sorry about the delay!

PR #280, opened 2017-09-05
Adds build instructions to each flight controller page

ksschwabe:
This commit adds the build instructions for each of the Pixhawk series flight controller boards, so that the pages are consistent with the Pixhawk 2 page, which had the build instructions.

This will help the users to know what build instruction needs to be run to compile the firmware for a particular flight controller.

### DIFF
--- a/en/flight_controller/pixfalcon.md
+++ b/en/flight_controller/pixfalcon.md
@@ -30,3 +30,7 @@ Pixfalcon is binary-compatible derivative of the [Pixhawk](../flight_controller/
   * 2x UART (one for Telemetry / OSD, no flow control)
   * 8x PWM with manual override
   * S.BUS / PPM input
+
+## Build Instructions
+
+`make px4fmu-v2_default upload`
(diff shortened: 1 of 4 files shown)


hamishwillee:
I'm actually moving all of these boards out to the PX4 User guide (as soon as @LorenzMeier does his review). I wouldn't want those build instructions in the user guide. Instead we'd have one page with the list of controllers vs FMU/build target. 

So this is good idea, but recommend we don't merge, or we do merge and understand that the info will be lost from here.

ksschwabe:
Understood. I just wanted to be able to find what build target related to what physical hardware controller board. I couldn't find a table anywhere.

There is the following page in the Px4 User Guide, https://docs.px4.io/en/flight_controller/pixhawk_series.html, but that still doesn't tell you that you need to run `make px4fmu-v4pro_default` to build for the Pixhawk 3 Pro.

hamishwillee:
@ksschwabe  Completely fair enough - and it is great that you also took the effort to add the information :-) 
I am [updating the pixhawk series](https://github.com/hamishwillee/px4_user_guide/blob/ef8a1d4a9323deed2711106124bacff98ba37ddd/en/flight_controller/pixhawk_series.md#fmu-versions) guide, and in that doc you could INFER the information. - but it isn't really the right place.

I'm going to merge this PR for now, and fix it properly with a table once the autopilots have moved to the user guide.: https://github.com/PX4/Devguide/issues/282

ksschwabe:
@hamishwillee: Sounds good. Thanks.

PR #281, opened 2017-09-05
Fixes path to Pixhawk 2 image

ksschwabe:
The path to the Pixhawk 2 image was incorrect and hence wasn't being displayed.


hamishwillee:
Thank you @ksschwabe !

PR #285, opened 2017-09-07
simulation: add details about MAVLink UDP ports.

JonasVautherin:
This PR aims at giving more details about the UDP ports used by the MAVLink interfaces. Another section somewhere in the guide should probably go into more details about that, but the scope of this PR is to give more details when explaining the config file.

Once the other, more specific section is created, the "init startup file" section could link to it.

Lastly, I believe that the PX4 SITL overview image could be updated to show the ports on both sides (e.g. 14556 on the PX4 side goes to 14550 on QGC).


hamishwillee:
Thanks. I've merged this, and will make my changes as a follow on PR.  I don't have access to modify in your repo.

hamishwillee:
OK, update done in #288. I've updated the image. Kept all your information, but restructured a little.

PR #289, opened 2017-09-13
Correct body frame's Y direction.

shivangag:
In ENU reference frame the body fixed frame's Y direction should be towards the left direction in accordance with the right-hand rule.


TSC21 (approved):

TSC21:
Correct and can be confirmed in [ROS REP 103](http://www.ros.org/reps/rep-0103.html).

PR #292, opened 2017-09-14
Add description for OcPoC-Zynq Mini

davidaroyer:
Adds basic documentation for Aerotenna OcPoC-Zynq Mini flight control hardware with links to detailed setup guide.


hamishwillee:
FYI, I am moving autopilot docs to the PX4 User Guide - just waiting on @LorenzMeier to approve this PR: https://github.com/PX4/px4_user_guide/pull/48

The audience is slightly different there - it will include non-developer "makers". The things that occurred to me are:
1. It might be worth including a sentence in the introduction about who it is for - ie primarily for commercial and researchers interested in seeing what can be done with high performance flight controller hardware.
2. Also worth having a note that firmware has to be built from source - no pre-built binaries supplied through QGroundControl. 

Other than that looks great - we can live with the missing full stop at end of para: "Pinouts for the default OcPo. " :-)

@bkueng Anything else to add?

I am

bkueng:
Maybe add a link to https://aerotenna.readme.io/docs/px4-setup in the building instructions?

davidaroyer:
Comments addressed, thanks!

hamishwillee:
Thanks @davidaroyer !

hamishwillee:
@davidaroyer  FYI, I have just recreated this doc in the [user guide](https://docs.px4.io/en/flight_controller/ocpoc_zynq.html) (and will shortly remove from the devguide). 

The user guide divides boards into pixhawk, boards with a separate compute module (e.g. aero, snapdragon), RTF UAVs that can run PX4. Not sure what category best fits your board - any thoughts on that?

davidaroyer:
Of those three categories, I think the only thing that makes sense is to lump it in with aero, snapdragon, and raspi.

hamishwillee:
@davidaroyer Yes, that's what I did. 
Would I be correct in thinking you're running PX4 in a Linux environment? - but not such that you can run other apps on the linux environment and communicate with the running PX4?
I've kind of thought of this as "flight controllers with separate comute platforms" if you can run other apps in the linux env.

davidaroyer:
Correct, we're running PX4 in Linux, but we are able to run other apps. So we technically don't have a separate compute platform, but we can work in a similar manner.

hamishwillee:
Thank you. So the category would be *PX4 on Linux*, or possibly "Flight Controllers with Off-Flight-Stack Computing" (bit of a mouthful). I'm more in favour of something like the second, because the options for additional computing are all  trying to solve the same problem. 
Are there any issues with doing things the way you have? Ie Linux is not an RTOS, and presumably you can't guarantee that the flight stack is getting enough resources?

PR #294, opened 2017-09-15
AT command correction

AlexisTM:
As stupid as it might sound, I had to check the SiK Firmware to check how to save the configuration. 

```
AT
&
W
```

to  

```
AT&W
```


hamishwillee:
Thanks @AlexisTM ! Matches http://files.rfdesign.com.au/Files/documents/RFD900u%20DataSheet.pdf

PR #301, opened 2017-10-04
Require installation of xxd (via vim) on archlinux

jlecoeur:
Since https://github.com/PX4/Firmware/pull/7873 the build system requires xxd.
On archlinux vim has to be installed, otherwise xxd is absent and the build fails.


dagar:
Thanks!

PR #304, opened 2017-10-07
Setup: add python-numpy and python-toml

jlecoeur:
Required by https://github.com/PX4/Firmware/pull/8063
Should be merged after https://github.com/PX4/containers/pull/77 is accepted.


hamishwillee:
Changes look good, but I can't assess if they are "necessary". I've nominated @dagar as approver for that purpose - @jlecoeur please let me know if there is someone else who should do that.

dagar:
Holding off until the Firmware side is ready.

PR #315, opened 2017-10-10
Make ubuntu_sim_nuttx.sh non-interactive

ShikherVerma:
The `ubuntu_sim_nuttx.sh` script is non-interactive in most places. This PR makes the remaining apt commands non-interactive.

### DIFF
--- a/build_scripts/ubuntu_sim.sh
+++ b/build_scripts/ubuntu_sim.sh
@@ -10,6 +10,10 @@
 ## - Gazebo8 simulator
 ## - PX4/Firmware source (to ~/src/Firmware/)
 
+# Preventing sudo timeout https://serverfault.com/a/833888
+trap "exit" INT TERM; trap "kill 0" EXIT; sudo -v || exit $?; sleep 1; while true; do sleep 60; sudo -nv; done 2>/dev/null &
+
+
 # Ubuntu Config
 sudo apt-get remove modemmanager -y
 
@@ -82,7 +86,7 @@ sudo apt-get update -y
 ## Install Gazebo8
 sudo apt-get install gazebo8 -y
 ## For developers (who work on top of Gazebo) one extra package
-sudo apt-get install libgazebo8-dev
+sudo apt-get install libgazebo8-dev -y
 
 
 # Clone PX4/Firmware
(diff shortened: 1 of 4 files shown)


ShikherVerma:
First of all. Thanks for this script :+1: saved me a lot of time.
However, I was expecting to run it once and let it execute in the background.
But the script stopped at some points due to missing `--yes` for `apt` commands.
Also a few times `sudo` timed out, due to which I had to regularly check if the script is still executing.
This PR only address the `apt` commands.
For fixing `sudo` timeout issue, [this one liner](https://serverfault.com/a/833888) can be added at the start of the script.
Should I add the sudo timeout fix too in this PR?

hamishwillee:
@ShikherVerma Thanks - great to see the script getting a workout from real users. You are quite right that it should be fully non-interactive ... unfortunately I am not a bash expert, so I missed these cases in my testing.

1. Could you also apply this change to the other scripts in the same directory?

FYI, there is a plan/hope that we will start using at least some of these scripts in our docker provisioning. This will allow us to better test them and keep them maintained ongoing.

2. 
> For fixing sudo timeout issue, this [one liner](https://serverfault.com/a/833888) can be added at the start of the script.
Should I add the sudo timeout fix too in this PR?

@dagar 
- Can you assess the proposed one liner change above?
- Any progress on the thinking to get these scripts used by our docker setup?

ShikherVerma:
> Could you also apply this change to the other scripts in the same directory?

Yeah sure

> FYI, there is a plan/hope that we will start using at least some of these scripts in our docker provisioning. This will allow us to better test them and keep them maintained ongoing.

That would be nice.

dagar (approved):

dagar:
Looks good to me.

What do you think about moving the scripts to the Firmware repo?

For docker we simply have this part download and execute the script. https://github.com/PX4/containers/blob/master/docker/px4-dev/Dockerfile_base#L11
It may need some tweaks.

hamishwillee:
@ShikherVerma Please make the updates to the other script then we can merge.

@dagar 

> What do you think about moving the scripts to the Firmware repo?

That would be best.

> For docker we simply have this part download and execute the script. https://github.com/PX4/containers/blob/master/docker/px4-dev/Dockerfile_base#L11
It may need some tweaks.

Great. As long as it still works stand-alone for end users I'll be happy. We should perhaps add notes at the top to make it clear that this is used in both docker and in "ordinary" installation?

hamishwillee:
Merged (finally!). Thanks for this.

PR #318, opened 2017-10-11
Add vtol-specific information to the airframe doc

hamishwillee:
The VTOL information was in an airframe overview doc. It probably belongs here.

### DIFF
--- a/en/airframes/adding_a_new_frame.md
+++ b/en/airframes/adding_a_new_frame.md
@@ -20,6 +20,7 @@ All configurations are stored in the [ROMFS/px4fmu_common/init.d](https://github
 
 A typical configuration file is shown below ([original file here](https://github.com/PX4/Firmware/blob/master/ROMFS/px4fmu_common/init.d/3033_wingwing)) .
 
+The first section is the airframe documentation. This is used in the [Airframes Reference](../airframes/airframe_reference.md) and *QGroundControl*.
 ```bash
 #!nsh
 #
@@ -40,7 +41,10 @@ A typical configuration file is shown below ([original file here](https://github
 #
 # @maintainer Lorenz Meier <lorenz@px4.io>
 #
+```
 
+The next section specifies tuning gains and other parameters:
+```bash
 sh /etc/init.d/rc.fw_defaults
 
 if [ $AUTOCNF == yes ]
@@ -63,11 +67,21 @@ then
 	param set FW_RR_FF 0.6
 	param set FW_RR_P 0.04
 fi
+```
 
+Set frame type:
+```bash
 # Configure this as plane
 set MAV_TYPE 1
+```
+
+Set the mixer:
+```bash
 # Set mixer
 set MIXER wingwing
+```
+
+```bash
 # Provide ESC a constant 1000 us pulse
 set PWM_OUT 4
 set PWM_DISARMED 1000
@@ -75,9 +89,10 @@ set PWM_DISARMED 1000
 
 > **Warning** If you want to reverse a channel, never do this on your RC transmitter or with e.g `RC1_REV`. The channels are only reversed when flying in manual mode, when you switch in an autopilot flight mode, the channels output will still be wrong (it only inverts your RC signal). Thus for a correct channel assignment change either your PWM signals with `PWM_MAIN_REV1` (e.g. for channel one) or change the signs of the output scaling in the corresponding mixer (see below).
 
+
 ### Mixer File
 
-A typical configuration file is below. Note that the mixer file contains several blocks of code, each of which refers to one actuator or ESC. So if you have e.g. two servos and one ESC, the mixer file will contain three blocks of code. 
+A typical mixer file is below. Note that the mixer file contains several blocks of code, each of which refers to one actuator or ESC. So if you have e.g. two servos and one ESC, the mixer file will contain three blocks of code. 
 
 > **Note** The plugs of the servos / motors go in the order of the mixers in this file.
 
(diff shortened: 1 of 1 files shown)


en/airframes/adding_a_new_frame.md
```diff
@@ -162,8 +162,21 @@ S: 0 3      0  20000 -10000 -10000  10000
...
+* `VT_ARSP_TRANS` is the airspeed in m/s at which the plane transitions into forward flight. Setting it too low can cause a stall during the transition.
+* `RC_MAP_TRANS_SW` should be assigned to a RC switch before flight. This allows you to check if the multicopter- and fixed wing mode work properly. (Can also be used to switch manually between those two control modes during flight)
+
+
+## Getting a New Airframe to Show in QGroundControl
 
 The airframe meta data is bundled in the .px4 firmware file (which is a zipped JSON file).
 
-> **Note** Ensure to flash the resulting .px4 file in QGroundControl (custom file option) to load the meta data into the application. The new airframe will then be available in the user interface once you restart QGroundControl.
+> **Note** Flash the resulting **.px4** file in *QGroundControl* (custom file option) to load the meta data into the application. The new airframe will then be available in the user interface once you restart *QGroundControl*.
```
hamishwillee, line 199: @bkueng I think this means "As soon as you have loaded the custom firmware onto any target vehicle the metadata will be available in the QGC airframe section. Can you confirm?

en/airframes/adding_a_new_frame.md
```diff
@@ -1,10 +1,10 @@
...
+# Adding a New Airframe Configuration
 
 PX4 uses canned configurations as starting point for airframes. Adding a configuration is straightforward: Create a new file which is prepended with a free autostart ID in the [init.d folder](https://github.com/PX4/Firmware/tree/master/ROMFS/px4fmu_common/init.d) and [build and upload](../setup/building_px4.md) the software.
 
 Developers not wanting to create their own configuration can instead customize existing configurations using text files on the microSD card, as detailed on the [custom system startup](../advanced/system_startup.md) page.
 
-## Airframe configurations
+## Airframe Configurations
 
 An airframe configuration consists of three main blocks:
```
hamishwillee: @bkueng The third block is "tuning gains" (not shown here). We don't have any comment/section on that block (while we do on config file and mixer file blocks) - where is that set/specified. Are there any other documents we  should link to explaining how to set these gains?

en/airframes/adding_a_new_frame.md
```diff
@@ -162,8 +162,21 @@ S: 0 3      0  20000 -10000 -10000  10000
 
 ```
 
-### Getting the new airframe to show in QGroundControl
+
+
+## VTOL-specific Configuration
```
hamishwillee: Does this section make sense here? Is there a better place?

en/airframes/adding_a_new_frame.md
bkueng: It's this block:
```
	param set BAT_N_CELLS 2
	param set FW_AIRSPD_MAX 15
	param set FW_AIRSPD_MIN 10
	param set FW_AIRSPD_TRIM 13
	param set FW_R_TC 0.3
	param set FW_P_TC 0.3
	param set FW_L1_DAMPING 0.74
	param set FW_L1_PERIOD 16
	param set FW_LND_ANG 15
	param set FW_LND_FLALT 5
	param set FW_LND_HHDIST 15
	param set FW_LND_HVIRT 13
	param set FW_LND_TLALT 5
	param set FW_THR_LND_MAX 0
	param set FW_PR_FF 0.35
	param set FW_RR_FF 0.6
	param set FW_RR_P 0.04
```
And it's more than just tuning gains. It includes vehicle-specific parameter settings, for example the number of battery cells, or things like the gps position offset relative to the board. I think it makes sense to add a short section about that here and include a link to the tuning gains page.

For the mixer we can link to https://dev.px4.io/en/concept/mixing.html.

en/airframes/adding_a_new_frame.md
bkueng: I think so, but I never load FW that way, so I don't really know how that works :)

en/airframes/adding_a_new_frame.md
bkueng: Yes, looks good. Unless there is a VTOL-specific tuning page?

hamishwillee: Good point thanks. I'll add a link to the tuning docs.

en/airframes/adding_a_new_frame.md
hamishwillee: Absolutely.

en/airframes/adding_a_new_frame.md
```diff
@@ -63,21 +67,34 @@ then
...
 set MAV_TYPE 1
+```
+
+Set the mixer:
+```bash
 # Set mixer
 set MIXER wingwing
+```
+
+```bash
```
hamishwillee: @bkueng Is there anything we can say about this section below? (ie what is happening here)

en/airframes/adding_a_new_frame.md
```diff
@@ -1,25 +1,26 @@
...
 PX4 uses canned configurations as starting point for airframes. Adding a configuration is straightforward: Create a new file which is prepended with a free autostart ID in the [init.d folder](https://github.com/PX4/Firmware/tree/master/ROMFS/px4fmu_common/init.d) and [build and upload](../setup/building_px4.md) the software.
 
 Developers not wanting to create their own configuration can instead customize existing configurations using text files on the microSD card, as detailed on the [custom system startup](../advanced/system_startup.md) page.
 
-## Airframe configurations
+## Airframe Configurations
 
 An airframe configuration consists of three main blocks:
 
   * The apps it should start, e.g. multicopter or fixed wing controllers
```
hamishwillee: @bkueng This is just the bit ` set MAV_TYPE 1` right? Any other "apps" are automatically set from whatever number you enter? 
- Is there a list of numbers
- should we call this "The controllers that should be started, e.g.. ..." instead?
- Perhaps move this below to match order of declaration in config file?

Also, should we add "documentation" in this list - it is important!

en/airframes/adding_a_new_frame.md
```diff
@@ -1,25 +1,26 @@
...
-  * The physical configuration of the system (e.g. a plane, wing or multicopter). This is called mixer.
```
hamishwillee: @bkueng Just for my info, this is the mixer file - which you specify in the config file using `set MIXER wingwing` right?

hamishwillee:
Hi @bkueng 

Thanks for your review. As you suggested I added link to the mixer concepts doc and to tuning section (and clarified that parameters are not just for tuning). The tuning section removed the "specific content" I added previously, and instead now links to our tuning guides for each frame. 

In addition, I split the example file out so that the content of each section is separately described - docs, setting mixer etc. I think that makes things more clear. This is worth your re-review please.

There are a couple of questions inline too.

en/airframes/adding_a_new_frame.md
bkueng: For this example it's just `MAV_TYPE`, yes.
- List of numbers: it's the `MAV_TYPE` param and @dagar added the full param description for the numbers in a PR: https://github.com/PX4/Firmware/pull/8001/files#diff-86d388773b907f1fd874d2990d901c8dR79 (It matches the definition in mavlink: http://mavlink.org/messages/common).
- It's more than controllers. For example it's also the land detector that depends on the type. And the Intel Aero starts a specific ESC driver: https://github.com/PX4/Firmware/blob/master/ROMFS/px4fmu_common/init.d/4070_aerofc#L71
- Makes sense

en/airframes/adding_a_new_frame.md
bkueng: Right

en/airframes/adding_a_new_frame.md
bkueng: We can say that this configures the PWM outputs: which outputs we want to drive/activate, and the levels.

hamishwillee:
Thanks @bkueng - I integrated your information, improved the structure, merged.

PR #320, opened 2017-10-17
update airframe reference and add Quad H

bkueng:
Firmware PR: https://github.com/PX4/Firmware/pull/8143
@hamishwillee


hamishwillee:
Thanks!

PR #322, opened 2017-10-19
Update micrortps.md

JaimeMartin:
Removed this line: RTPS is not suitable for use over slow links (e.g. like radio telemetry), because

RTPS is suitable for use over slow links (e.g. like radio telemetry), as a matter of fact, companies such as 3D Robotics (Dronecode Foundation) have been testing successfully Fast RTPS over radio links, and we have some experience using RTPS in tactical radios with bandwidths as low as 2400bps shared.

### DIFF
--- a/en/middleware/micrortps.md
+++ b/en/middleware/micrortps.md
@@ -11,7 +11,7 @@ This topic describes the bridge architecture, how it is compiled, and how to wri
 
 ## When should RTPS be used?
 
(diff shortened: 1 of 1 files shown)


hamishwillee:
Sounds fair, but there is still a risk - even with MAVLink you need to be careful of your telemetry rates over radio. Perhaps a weaker "Note" would still be useful. Something like:

> **Note** RTPS can be used over slower links (e.g. like radio telemetry) but care should be taken not to   overload the channel. 

?

bkueng:
Sounds good to me. @JaimeMartin what do you think?

JaimeMartin:
Sounds good.

hamishwillee:
Great. FYI, added note in https://github.com/PX4/Devguide/pull/325

PR #329, opened 2017-10-26
fix the http link and in-correct build info for mindpx

flynxea:
fix the http link and in-correct build info for mindpx


hamishwillee:
Thank you! I'll merge now.

PR #331, opened 2017-10-27
Fix astyle path

ndepal:
Also: With the recommended version of astyle 2.06 from sourceforge, I get a huge diff with `./Tools/astyle/check_code_style_all.sh`. Its so long it overflows the buffer of my terminal and I can only see the diff of `src/lib/micro-CDR/test/AllTest.cpp`, there might be more.

### DIFF
--- a/en/contribute/README.md
+++ b/en/contribute/README.md
@@ -19,7 +19,7 @@ PX4 uses [astyle](http://astyle.sourceforge.net/) for code formatting. Valid ver
 * [astyle 2.06](https://sourceforge.net/projects/astyle/files/astyle/astyle%202.06/) (recommended)
 * [astyle 3.0](https://sourceforge.net/projects/astyle/files/astyle/astyle%203.0/)
 
-Once installed, formatting can be checked with `./Tools/check_code_style_all.sh`. The output should be `Format checks passed` on a clean master. If that worked, `make format` can be used in the future to check and format all files automatically.
+Once installed, formatting can be checked with `./Tools/astyle/check_code_style_all.sh`. The output should be `Format checks passed` on a clean master. If that worked, `make format` can be used in the future to check and format all files automatically.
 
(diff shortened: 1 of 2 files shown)


korigod:
The Korean version have same astyle script path issue.

korigod (changes requested):
Commit message should describe commit itself, not some unrelated problem.

hamishwillee:
@ndepal  - what @korigod  is suggesting is that it would be good to also update https://github.com/PX4/Devguide/blob/master/kr/contribute/README.md

@FantasyJXF The whole chinese page probably needs to be re-translated: https://github.com/PX4/Devguide/blob/master/zh/contribute/README.md

ndepal:
@korigod addressed both comments

hamishwillee:
@ndepal Thanks for this, merged.

Regarding:

> Also: With the recommended version of astyle 2.06 from sourceforge, I get a huge diff with ./Tools/astyle/check_code_style_all.sh. Its so long it overflows the buffer of my terminal and I can only see the diff of src/lib/micro-CDR/test/AllTest.cpp, there might be more.

Have you tried the other versions of asytle, and do they give the same problems? If not, @bkueng  any reason we can might perhaps recommend a later version?

bkueng:
Well, I have 2.06 and it works for me. I don't have an explanation why the same version should give different results.
We have always had some troubles with different versions, so personally I would prefer to switch to `clang-format`.

hamishwillee:
@bkueng Seems reasonable, but won't that potentially result in lots of spurious changes too? In any case, what would have to happen, to make that happen? Not really for me to drive this :-)

bkueng:
Yes it will definitely create a large diff, which is probably why it hasn't happened yet.

hamishwillee:
Thanks @bkueng , https://github.com/PX4/Firmware/issues/8277 created for tracking if anything happens with that.

PR #338, opened 2017-11-03
Update rtps documentation

avinash-palleti:
Updated couple of paths as per current implementation.


avinash-palleti:
@bkueng Please check.

bkueng:
Looks good, thanks!

PR #346, opened 2017-11-14
Remove LNDMC_POS_UPTHR and LNDMC_MAN_DWNTHR param from doc

hamishwillee:
This updates dev guide to catch up with this change https://github.com/PX4/Firmware/commit/7229ec2a378758350a9a4121c511c480550f7e15#diff-b2da8c756b49a745c8f649f0d57567c7 which removed LNDMC_POS_UPTHR and LNDMC_MAN_DWNTHR params.

The change updates the chinese and korean trees to use latest paramater_reference, and removes mention of the params from land detector docs. I also took time to add param links to other parameters in the land detector doc.

### DIFF
--- a/en/tutorials/land_detector.md
+++ b/en/tutorials/land_detector.md
@@ -10,12 +10,10 @@ By default the land detector does detect landing, but does not auto-disarm. If t
 
 The complete set of parameters is available in the *QGroundControl* parameter editor under the `LNDMC` prefix. The key parameters that might differ per airframe are these:
 
-
 * [MPC_THR_HOVER](../advanced/parameter_reference.md#MPC_THR_HOVER) - the hover throttle of the system \(in percent, default is 50%\). It is important to set this correctly as it does not only make the altitude control more accurate, but also ensures correct land detection. A racer or a big camera drone without payload mounted might need a much lower setting \(e.g. 35%\).
-* [MPC_THR_MIN](../advanced/parameter_reference.md#MPC_THR_MIN) - the overall minimum throttle of the system. This should be set to enable a controlled descend.
+* [MPC_THR_MIN](../advanced/parameter_reference.md#MPC_THR_MIN) - the overall minimum throttle of the system. This should be set to enable a controlled descent.
 * [LNDMC_THR_RANGE](../advanced/parameter_reference.md#LNDMC_THR_RANGE) - this is a scaling factor to define the range between min and hover throttle that gets accepted as landed. Example: If the minimum throttle is 0.1, the hover throttle is 0.5 and the range is 0.2 \(20%\), then the highest throttle value that counts as landed is: `0.1 + (0.5 - 0.1) * 0.2 = 0.18`.
 
-The throttle level to trigger takeoff is hard-coded to 62.5%. If the pilot raises above this threshold the system will attempt to take off. This value should be greater than the hover throttle.
 
 ### Landdetector states
 
--- a/en/tutorials/land_detector.md
+++ b/en/tutorials/land_detector.md
@@ -13,7 +13,7 @@ The complete set of parameters is available in the *QGroundControl* parameter ed
 
 * [MPC_THR_HOVER](../advanced/parameter_reference.md#MPC_THR_HOVER) - the hover throttle of the system \(in percent, default is 50%\). It is important to set this correctly as it does not only make the altitude control more accurate, but also ensures correct land detection. A racer or a big camera drone without payload mounted might need a much lower setting \(e.g. 35%\).
 * [MPC_THR_MIN](../advanced/parameter_reference.md#MPC_THR_MIN) - the overall minimum throttle of the system. This should be set to enable a controlled descend.
-* [LNDMC_THR_RANGE](../advanced/parameter_reference.md#LNDMC_THR_RANGE) - this is a scaling factor to define the range between min and hover throttle that gets accepted as landed. Example: If the minimum throttle is 0.1, the hover throttle is 0.5 and the range is 0.2 \(20%\), then the highest throttle value that counts as landed is the min throttle plus \(0.5 - 0.1\) \* 0.2 = 0.08, so a total throttle of 0.18.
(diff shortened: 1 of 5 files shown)


en/tutorials/land_detector.md
```diff
@@ -4,45 +4,49 @@ The land detector is a dynamic vehicle model representing key vehicle states suc
...
 
-By default the land detector does detect landing, but does not auto-disarm. If the hysteresis parameter `COM_DISARM_LAND` is set to a non-zero value the system will auto-disarm after N seconds \(the value it is set to\).
+By default the land detector does detect landing, but does not auto-disarm. If the hysteresis parameter [COM_DISARM_LAND](../advanced/parameter_reference.md#COM_DISARM_LAND) is set to a non-zero value the system will auto-disarm after N seconds \(the value it is set to\).
 
 ## Multicopter Land Detector Configuration
 
-The complete set of parameters is available in the QGroundControl parameter editor under the `LNDMC` prefix. The key parameters that might differ per airframe are these:
+The complete set of parameters is available in the *QGroundControl* parameter editor under the `LNDMC` prefix. The key parameters that might differ per airframe are these:
 
-* `LNDMC_MAN_DWNTHR` - the threshold \(in percent, default is 15%\) for how much manual throttle is allowed to be considering the state as landed. Systems with very high thrust-to-weight ratios like racers might need a lower setting here \(e.g. 8%\).
```
hamishwillee: @julianoes 
1. LNDMC_MAN_DWNTHR was removed, which is fine, but how has the land detector changed correspondingly? Is the value now hard coded or is throttle no longer a relevant parameter (ie can I ignore the functionality this used to have. 
2. Is there anything else that needs to change in this doc, either because of this removal, or because you "just happen to know" there are other issues with it? (i.e. are there any other known changes in land detection we need to record).

en/tutorials/land_detector.md
```diff
@@ -4,45 +4,49 @@ The land detector is a dynamic vehicle model representing key vehicle states suc
...
-* `LNDMC_POS_UPTHR` - throttle level to trigger takeoff \(in percent, default is 65%\). If the pilot raises above this threshold the system will attempt to take off. This value should be greater than the hover throttle.
+
+* [MPC_THR_HOVER](../advanced/parameter_reference.md#MPC_THR_HOVER) - the hover throttle of the system \(in percent, default is 50%\). It is important to set this correctly as it does not only make the altitude control more accurate, but also ensures correct land detection. A racer or a big camera drone without payload mounted might need a much lower setting \(e.g. 35%\).
+* [MPC_THR_MIN](../advanced/parameter_reference.md#MPC_THR_MIN) - the overall minimum throttle of the system. This should be set to enable a controlled descend.
+* [LNDMC_THR_RANGE](../advanced/parameter_reference.md#LNDMC_THR_RANGE) - this is a scaling factor to define the range between min and hover throttle that gets accepted as landed. Example: If the minimum throttle is 0.1, the hover throttle is 0.5 and the range is 0.2 \(20%\), then the highest throttle value that counts as landed is the min throttle plus \(0.5 - 0.1\) \* 0.2 = 0.08, so a total throttle of 0.18.
```
julianoes: I would write the example equation as `0.1 + \(0.5 - 0.1\) \* 0.2 = 0.18`

en/tutorials/land_detector.md
```diff
@@ -4,45 +4,49 @@ The land detector is a dynamic vehicle model representing key vehicle states suc
...
+
+The throttle level to trigger takeoff is hard-coded to 62.5%. If the pilot raises above this threshold the system will attempt to take off. This value should be greater than the hover throttle.
```
julianoes: This value of 62.5% is not actually connected to the throttle that the vehicle is generating, so it has nothing to do with hover throttle, I think.

It just means that if the stick is raised above 62.5% percent in altitude or position control, the controller will "do a takeoff". 

The logic is implemented in the position controller and not in the land detector, so that was the main change. It is now not the land detector detecting that the motors are spinning up and therefore saying "we're in air" which would tell the position controller to do a takeoff but instead the land detector will detect it based on the throttle output while the position controller looks at the manual input (stick).

julianoes (changes requested):

en/tutorials/land_detector.md
hamishwillee: Thanks. Will replace this with:

 Example: If the minimum throttle is 0.1, the hover throttle is 0.5 and the range is 0.2 \(20%\), then the highest throttle value that counts as landed is: `0.1 + (0.5 - 0.1) * 0.2 = 0.18`.

en/tutorials/land_detector.md
hamishwillee: @julianoes OK, so I'll remove this comment from here altogether because it doesn't make sense in the context of land detection. Reasonable?
OK to merge?

This still looks useful info though. Would I be correct if in [Altitude mode](https://docs.px4.io/en/flight_modes/altitude.html) I added a line: 

> The vehicle will take off if the stick is raised above 62.5% percent.

When you say 62.5% you mean "off the full range from the throttle stick bottom position" right? So in effect, to just above the large dead zone in the middle?

LorenzMeier (approved):

LorenzMeier:
@hamishwillee CI failure still needs fixing.

en/tutorials/land_detector.md
julianoes: I think so but I'm not 100% sure.

julianoes (approved):

en/tutorials/land_detector.md
hamishwillee: Thanks, I'll track that in https://github.com/PX4/px4_user_guide/issues/96

hamishwillee:
@LorenzMeier 

> @hamishwillee CI failure still needs fixing.

It fails because it checks for broken links in the wider internet - sometimes sites fail. I've changed things (I hope) so CI always succeeds, but hopefully I should still get fail notifications on broken links.

PR #348, opened 2017-11-15
added precision land tutorial and description

ndepal:


en/debug/precland.md
```diff
@@ -0,0 +1,47 @@
...
+
+The measurements in `irlock_report` contain the tangent of the angles from the image center to the beacon. In other words, the measurements are the x and y components of the vector pointing towards the beacon, where the z component has length 1.
+This means that scaling the measurement by the distance from the camera to the beacon results in the vector from the camera to the beacon.
+This relative position is then rotated into the north-aligned, level body frame using the vehicle's attitude estimate.
+Both x and y components of the relative position measurement are filtered in separate Kalman Filters, which act as simple low-pass filters that also produce a velocity estimate and allow for outlier rejection.
+
+The `landing_target_estimator` publishes the estimated relative position and velocity, whenever a new `irlock_report` is fused into the estimate. Nothing is published if the beacon is not seen or beacon measurements are rejected.
+The landing target estimate is published in the `landing_target_pose` message.
```
hamishwillee, line 14: is that a uORB message?

en/debug/precland.md
```diff
@@ -0,0 +1,47 @@
...
+
+
+## Vehicle state estimate aiding
```
hamishwillee, line 17: This sounds wrong in English. Could this be: "Enhanced Vehicle Position Estimate" ?

en/debug/precland.md
```diff
@@ -0,0 +1,47 @@
...
+This is done by fusing the beacon's velocity as a measurement of the negative velocity of the vehicle.
+
+## Precision Land controller
+A precision landing can be initiated by switching to the Precision land flight mode:
+```
+commander mode auto:precland
+```
+In this case, the precision landing is always considered "required".
+
+Alternatively, a precision landing can be initiated as part of a [mission](../tutorials/precland.html#in-a-mission).
```
hamishwillee, line 30: should be '.md' not '.html'

en/debug/precland.md
```diff
@@ -0,0 +1,47 @@
...
+
+
+## Vehicle state estimate aiding
+> **Note** Vehicle state estimate aiding is currently only supported with LPE.
+
+If the beacon is specified to be stationary using the parameter `LTEST_MODE`, the vehicle's position/velocity estimate can be improved with the help of the beacon measurements.
+This is done by fusing the beacon's velocity as a measurement of the negative velocity of the vehicle.
+
+## Precision Land controller
+A precision landing can be initiated by switching to the Precision land flight mode:
```
hamishwillee, line 24: Can QGC be configured to assign this mode to a switch yet?

en/tutorials/precland.md
```diff
@@ -0,0 +1,67 @@
+# Precision Landing
+
+PX4 supports precision landing for multicopters with the [IRLock](https://irlock.com/products/ir-lock-sensor-precision-landing-kit) sensor and a downward facing range sensor.
+
+A precision landing can be initiated by entering the precision land flight mode, or as part of a mission uploaded via Mavlink.
```
hamishwillee, line 5: MAVLink

en/tutorials/precland.md
```diff
@@ -0,0 +1,67 @@
...
+
+Besides using the beacon (landing target) to control the vehicle to land on top of it, it can be used to improve the vehicle's localization, if the beacon is stationary.
+This is controlled by a parameter (see below).
+Note: This is currently only possible with LPE.
```
hamishwillee, line 9: Notes are done like this: `> **Note** Some text here`

en/tutorials/precland.md
```diff
@@ -0,0 +1,67 @@
...
+Measurement scaling may be necessary due to lens distortions of the IRLock sensor.
+Note that `LTEST_SCALE_X` and `LTEST_SCALE_Y` are considered in the sensor frame, not the vehicle frame.
+
+To calibrate these scale parameters, set `LTEST_MODE` to moving, fly your multicopter above the beacon and perform forward-backward and left-right motions with the vehicle, while [logging](../log/logging.html#configuration) `landing_target_pose` and `vehicle_local_position`.
+Then, compare `landing_target_pose.vx_rel` and `landing_target_pose.vy_rel` to `vehicle_local_position.vx` and `vehicle_local_position.vy`, respectively (both measurements are in NED frame). If the estimated beacon velocities are consistently smaller or larger than the vehicle velocities, adjust the scale parameters to compensate.
+
+If you are observing slo sideways oscillations of the vehicle while doing a precision landing with `LTEST_MODE` set to stationary, the beacon measurements are likely scaled too high and you should reduce the scale parameter in the relevant direction.
```
hamishwillee, line 52: slow, not "slo"

en/debug/precland.md
```diff
@@ -0,0 +1,47 @@
+# Precision Landing
```
hamishwillee, line 1: I think probably should be in /advanced/ not debug (if in Devguide at all). The two precland topics should have slightly different headings/titles to differentiate them in search.

hamishwillee:
This is good material, but I have a few concerns:
1. There is a lot of duplication across the two docs. Where possible this should be removed. It would be great if they could be combined. 
2. Are users expected to use precision landing? In this case there should be a topic [like this one](https://docs.px4.io/en/flight_modes/land.html) in the User guide. It would obviously be more complicated, because hardware setup is required. Again, though we'd want to avoid duplication across the two sets. 
3. Generally we don't have a really firm split yet on what is in DevGuide vs User guide. IMO though information needs to go in the User guide if you want people to *use* a feature when flying, OR you want them to be able to build an airframe with the "thing". Stuff goes in the devguide if you want people to be able to modify the component to behave differently. This doesn't always "work" for a logical split of course, but something to think about.
4. Lets see what @bkueng suggests. It may be we accept this as is, but rework it as part of our planned DevGuide structural updates.

en/debug/precland.md
ndepal: It is. I'll mention it.

LorenzMeier:
Hamish, could we get these instructions into the dev guide so we can point people who want to write articles in the future to it?

en/debug/precland.md
ndepal: It does sound cumbersome, changing to "Enhanced Vehicle Position Estimation"

en/debug/precland.md
ndepal: No, this is currently not possible.

ndepal:
Thanks for the review, @hamishwillee 

> There is a lot of duplication across the two docs. Where possible this should be removed. It would be great if they could be combined.

I agree. I'm happy to combine them into one document. In which section should it be then?

>Are users expected to use precision landing? In this case there should be a topic like this one in the User guide. It would obviously be more complicated, because hardware setup is required. Again, though we'd want to avoid duplication across the two sets.

Yes this is something a user can use. The reason for putting it in the DevGuide for me was that currently the required modules are not compiled in any standard configuration, so the "user" does have to have their build environment set up etc. We could of course also discuss including the modules in the default configurations, but my guess is that there are other modules which are used more that will take preference over these, especially on fmu v2.

hamishwillee:
@ndepal  [EDITED ... twice]

>> There is a lot of duplication across the two docs. Where possible this should be removed. It would be great if they could be combined.

> I agree. I'm happy to combine them into one document. In which section should it be then?

I think you should delete this branch and combined them in a topic under [PX4 User Guide > Advanced Features](https://docs.px4.io/en/advanced_features/) (or possible advanced configuration ....)

>The reason for putting it in the DevGuide ...

All very reasonable and understandable but the user guide is "right" for the reasons provided. Advanced configuration is good because this is "non standard setup". 

I will also split out some "how to use when flying" instructions in a flight mode document in the user guide, but that can come a bit later. 

> So what can go in the devguide?

Stuff of interest to developers - e.g. 

1. Do we need to update this diagram? https://dev.px4.io/en/concept/flight_modes.html#flight-mode-evaluation-diagram - if so, feel free to hand draw over the top and send me a picture and I'll get it updated. 
2. Any special integration needed to work with a new rangefinder that requires code changes
3. Simulating  precision landing - this SHOULD be user guide, but our simulation isn't robust enough. - is this possible?

> ... A few more questions ..

1. What triggers QGC to offer precision landing as an option for switch selection?
2. What does QGC use for the string in the selection box for mode?

ndepal:
> I think you should delete this branch and combined them in a topic under PX4 User Guide > Advanced Features (or possible advanced configuration ....)

Ok, will do.

> Do we need to update this diagram?

Potentially. But only once switching via RC is supported. I do not have experience with this and would want support from somebody who does.

> Simulating precision landing

Simulation is indeed possible, I forgot to add it to the guide, thanks for the reminder. I'll just put it in the one user guide page I will make now, ok?

>What triggers QGC to offer precision landing as an option for switch selection?
What does QGC use for the string in the selection box for mode?

I would assume that this requires changes on the QGC side. The name string is probably hardcoded in QGC.

hamishwillee:
Closing - - moved to user guide: https://github.com/PX4/px4_user_guide/pull/98

hamishwillee:
Closed. I'll continue on the other one ....

PR #345, opened 2017-11-14
Updates to support collapsible menu

hamishwillee:
HI @bkueng 

The flat structure of the DevGuide started to annoy me. This update makes the menu collapsible, and does some other minor tweaks. It can be previewed here: https://hamishwillee.gitbooks.io/havdevguide/content/v/collapsible_menu/en/

ANyway, I think this is good. It is now immediately obvious that perhaps AutoPilot Hardware should move somewhere (maybe a concept?) and that the robotics topics might reasonably live under a parent grouping etc.

Let me know what you think.

PR #350, opened 2017-11-16
Remove fakegps

hamishwillee:
As discussed in https://github.com/PX4/Devguide/pull/345#issuecomment-344528298


TSC21:
IMO this could be changed instead of deleted, as `fake_gps` is still a possiblility with MAVROS, though the name changed from `mocap_fake_gps.cpp` to `fake_gps.cpp` as it allows other sources of estimated pose besides motion capture

hamishwillee:
@TSC21  So the document (before deleted) said "don't bother with me" see https://dev.px4.io/en/ros/external_position_estimation.html

Happy for it to be recovered - if you think it is useful and can do the work ... or point to someone who knows enough to do so?

TSC21:
We can do some adaptations to https://dev.px4.io/en/ros/external_position_estimation.html to include this. I can do it, though other people are able. @julianoes was the first to write the plugin so he is able to if he has the bandwidth.

hamishwillee:
Thanks. I've added https://github.com/PX4/Devguide/issues/351 to track this. Initially assigned to Julian, though he may ping it back to you!

PR #354, opened 2017-11-19
Improved ubuntu build scripts

cyberang3l:
Pull request related to issue #353.
Detailed in information about the pull request in the issue #353, as well as the commit messages.

PR #358, opened 2017-11-21
Ubuntu scripts now all depend in a new script that installs the common dependencies

cyberang3l:
This pull request is following up a discussion we had in the thread of PR #354.

All the scripts are installing at least the common dependencies, but the shell code that installs the common  dependencies was copied in each one of the scripts. This is error prone. If the common dependencies change, then one has to go to each one of these scripts to do the changes.
    
This pull request removes duplicate code from the ubuntu build scripts. All the ubuntu scripts now depend on another script that is responsible to install the common dependencies.

More details in the commit messages.

### How and what to test
Since the scripts now depend on another main script (`ubuntu_sim_common_deps.sh`), the first thing they do is to download this script from the location `https://raw.githubusercontent.com/PX4/Devguide/master/build_scripts/ubuntu_sim_common_deps.sh` (Note that the `ubuntu_sim_nuttx.sh` script depends on the `ubuntu_sim.sh` script, which then depends on `ubuntu_sim_common_deps.sh`. That's because the `ubuntu_sim_nuttx.sh` was a superset of the `ubuntu_sim.sh`).

So you need to test two things:

1. Test what happens in the case where the common dependencies cannot be found (network problems or broken link). This is the default behavior for the moment since this PR hasn't been accepted yet (thus, the depending script cannot be found in `https://raw.githubusercontent.com/PX4/Devguide/master/build_scripts/ubuntu_sim_common_deps.sh`). The scripts have been designed to check if the dependent script has been downloaded successfully, and if not, the user will be warned and the script will not run any further.

2. In order to test the successful execution of the scripts, edit the scripts and change the URL of the dependent script to point to my git branch at `https://raw.githubusercontent.com/cyberang3l/Devguide/ubuntu-scripts-commondeps/build_scripts/ubuntu_sim_common_deps.sh` (and for the nuttx script at `https://raw.githubusercontent.com/cyberang3l/Devguide/ubuntu-scripts-commondeps/build_scripts/ubuntu_sim.sh`

### DIFF
--- a/build_scripts/ubuntu_sim.sh
+++ b/build_scripts/ubuntu_sim.sh
@@ -7,11 +7,12 @@
 ## - Common dependencies libraries and tools as defined in `ubuntu_sim_common_deps.sh`
 ## - Gazebo8 simulator
 
+echo "Downloading dependent script 'ubuntu_sim_common_deps.sh'"
 # Source the ubuntu_sim_common_deps.sh script directly from github
 common_deps=$(wget https://raw.githubusercontent.com/PX4/Devguide/master/build_scripts/ubuntu_sim_common_deps.sh -O -)
 wget_return_code=$?
 # If there was an error downloading the dependent script, we must warn the user and exit at this point.
-if [[ $wget_return_code -ne 0 ]]; then echo "Error downloading 'ubuntu_sim_common_deps.sh'"; exit 1; fi
+if [[ $wget_return_code -ne 0 ]]; then echo "Error downloading 'ubuntu_sim_common_deps.sh'. Sorry but I cannot proceed further :("; exit 1; fi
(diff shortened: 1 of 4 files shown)


hamishwillee:
Fantastic - thank you (I tested just one file, but the others look good too!). 

@dagar Anything you want to add? If not, I hope to merge tomorrow.

I will need to do minor update to https://dev.px4.io/en/setup/dev_env_linux_ubuntu.html#convenience-bash-scripts

TSC21 (approved):

hamishwillee:
Merged! Thank you @cyberang3l

hamishwillee:
FYI, associated docs updates in https://github.com/PX4/Devguide/pull/360

PR #360, opened 2017-11-22
Update docs to reflect new build scripts

hamishwillee:


cyberang3l:
Just a comment in the following line:

This is not correct: `ubuntu_sim_nuttx.sh: ubuntu_sim_common_deps.sh + NuttX tools.`

It is actually: `ubuntu_sim_nuttx.sh: ubuntu_sim.sh + NuttX tools.`

The original NuttX script was downloading Gazebo8 as well, and the `ubuntu_sim.sh` is `ubuntu_sim_common_deps.sh + Gazebo8`. If we do not want Gazebo to be installed with the nuttx script, then we can change the script to depend on `ubuntu_sim_common_deps.sh` instead of depending on `ubuntu_sim.sh`.

hamishwillee:
@cyberang3l  Argggg! Thanks for the information. Fixed by #364

Issue #357, opened 2017-11-20
Simulation UDP port description is a bit confusing


cyberang3l:
In this page: https://dev.px4.io/en/simulation/, the section `Standard MAVLink UDP Ports` has the following content:

```
PX4 defines a number of standard UDP broadcast ports that ground control stations (e.g. QGroundControl, Offboard APIs and simulators) can listen to in order to connect to the autopilot:

Offboard APIs: 14540
Ground Control Stations: 14550 (QGroundControl automatically connects to PX4 broadcasting on this port).
Simulators: 14560
```

Technically, this is not precise.

As I understand after some further investigation, the simulators do not listen to port `14560`. They listen to a randomly chosen port. If one executes the `ss -upnl4` command when running a gazebo simulation, they will see something like this:

```bash
$ ss -upnl4
State      Recv-Q Send-Q          Local Address:Port                         Peer Address:Port              
UNCONN     0      0                           *:14556                                   *:*                   users:(("px4",pid=3992,fd=7))
UNCONN     0      0                           *:14557                                   *:*                   users:(("px4",pid=3992,fd=8))
UNCONN     0      0                           *:14560                                   *:*                   users:(("px4",pid=3992,fd=5))
UNCONN     0      0                           *:631                                     *:*                  
UNCONN     0      0                           *:42044                                   *:*                   users:(("gzserver",pid=3963,fd=34))
UNCONN     0      0                           *:35909                                   *:*                  
UNCONN     0      0                           *:5353                                    *:*
```

It is obvious from here that PX4 is listening to 14556, 14557 and 14560, and the simulator is listening to a random port (in this case 42044).

A tcpdump also shows that it is PX4 that is using the src ports 14556 and 14557 to broadcast data to the dst ports 14550 and 14540 respectively, but it is the simulator that is using the random src port (42044 in this example) to broadcast data to the port 14560.

```bash
$ sudo tcpdump -i any -n -c5 'udp and port 14556'
tcpdump: verbose output suppressed, use -v or -vv for full protocol decode
listening on any, link-type LINUX_SLL (Linux cooked), capture size 262144 bytes
18:49:30.656139 IP 127.0.0.1.14556 > 127.0.0.1.14550: UDP, length 70
18:49:30.660830 IP 127.0.0.1.14556 > 127.0.0.1.14550: UDP, length 70
18:49:30.667071 IP 127.0.0.1.14556 > 127.0.0.1.14550: UDP, length 70
[...]

PR #363, opened 2017-11-23
Simulator clarifications 

hamishwillee:
These address issues as per #357. Images not yet updated.


en/simulation/README.md
```diff
@@ -48,23 +48,23 @@ Message | Direction | Description
 [HIL_RC_INPUTS_RAW](http://mavlink.org/messages/common#HIL_RC_INPUTS_RAW) | Sim to PX4 | The RAW values of the RC channels received.
 
 
-## Standard MAVLink UDP Ports
+## Default PX4 MAVLink UDP ports
```
hamishwillee, line 51: FMI, capitalise "Ports"

en/simulation/README.md
```diff
@@ -48,23 +48,23 @@ Message | Direction | Description
...
 
-PX4 defines a number of standard UDP broadcast ports that ground control stations (e.g. *QGroundControl*, Offboard APIs and simulators) can listen to in order to connect to the autopilot:
+By default, PX4 uses well-defined MAVLink UDP ports for communication with ground control stations (e.g. *QGroundControl*), Offboard APIs (e.g. DroneCore, MAVROS) and simulator APIs (e.g. Gazebo). These ports are:
 
-- Offboard APIs: 14540
-- Ground Control Stations: 14550  (*QGroundControl* automatically connects to PX4 broadcasting on this port).
-- Simulators: 14560
+* Port **14540** is used for communication with offboard APIs. PX4 broadcasts data towards this port, and Offboard APIs are expected to listen.
```
hamishwillee: @cyberang3l  Might the first two be better as below (removing "broadcast" removes the "feeling" that this is single direction communication).

* Port **14540** is used for communication with offboard APIs. Offboard APIs are expected to listen for connections on this port. 
* Port **14550** is used for communication with ground control stations. GCS are expected to listen for connections on this port. *QGroundControl* listens to this port by default.

cyberang3l: I agree with this. Better remove `broadcast`.

cyberang3l:
As you explained to me in #357, the ports are not `well-defined` by means of being `standardized`. However these ports are the *defacto* standard across different autopilot software solutions so I would call them `commonly established`. Thus, I would only change this part:

`By default, PX4 uses well-defined MAVLink UDP ports for communication`

to this:

`By default, PX4 uses commonly established MAVLink UDP ports for communication`

The word `commonly` redirects me to think about something that is, ...well..., common to use.
The word `establish` makes me think about something has managed to rightfully claim a status: to become common in this case.

Figures are coming.

cyberang3l:
Black dashed lines: Internal component communication.
Purple solid lines: Serial communication.
Red dashed lines: MAVLink UDP communication. Can be wireless or not.

px4_simulator_messages

px4_sitl_overview

px4_hitl_overview

hamishwillee:
@cyberang3l  Thanks very much! I have integrated your suggestions an the images and merged. 

Minor point is that in SITL/HITL diagram we refer to comms as MAVLink API Communication. MAVLink is actually a protocol, so this should probably be  just "MAVLink" or "MAVLink Protocol" or "MAVLink Communication". Thoughts?

cyberang3l:
I agree. It should be a plain `MAVLink communication`.
I make a quick fix on that and send you the new figures.

cyberang3l:
px4_hitl_overview

px4_sitl_overview

hamishwillee:
Cheers, merged as above. Just as an aside, we could be *very* pedantic, that the HITL serial comms is actually MAVLink over serial. But we aren't going to be :-)

PR #366, opened 2017-11-23
Add I2C overview text

hamishwillee:
The I2C overview is currently empty. This provides minimal text about what I2C is. The aim though is to use this as a place to discuss what we actually want in the system.

This is a developer/integration guide. We want to make it easy for developers to:
1. Understand what the I2C bus is
2. Understand when you would use the I2C Bus, and when you would use UART or SMBus.
3. Understand what you need to do to integrate *any* hardware with this bus:
   - write a driver? If so, what are the requirements on that driver?
   - include driver in ROM? Start driver? 
   - What is best practise - ie how should the driver interact with a port? Should it be turned on by a parameter?
4. Provide concrete examples of best use.

Separate but related, what do you need to do that is specific to a particular hardware type. E.g. I want to integrate a rangefinder, what do I need to do in my driver for this. Where do I then talk to the rest of the system.

@bkueng @LorenzMeier I don't expect you to answer all that, but some pointers would be nice :-)


LorenzMeier (approved):

en/sensor_bus/i2c.md
```diff
@@ -1 +1,11 @@
-# I2C Bus Overview
\ No newline at end of file
+# I2C Bus Overview
+
+I2C is a packet-switched serial communication protocol that allows multiple master devices to connect to multiple slave devices using only 2 wires per connection. It is intended for attaching lower-speed peripheral ICs to processors and microcontrollers in short-distance, intra-board communication. 
+Pixhawk/PX4 support it for connecting off board components that require greater data rates than provided by a strict serial UART: e.g. rangefinders.
```
hamishwillee: Is this the only reason we choose I2C - because of the higher data rates than UART? Or are some hardwares likely only to be I2C compatible, so we'd choose in that case in order to be able to talk to this sort of hardware.

bkueng: Yes, compatibility is one point, and another is multiplexing: there can be multiple devices attached to an I2C bus, wheras on UART you only have one.

We should probably also mention that an IMU (accel/gyro) should not be attached via I2C.

bkueng (commented):
For an example we could point to some existing driver(s). Then it will be kept automatically up-to-date.
And adding it to the ROMFS depends usually on the specific case and can be discussed on a PR.

hamishwillee:
> We should probably also mention that an IMU (accel/gyro) should not be attached via I2C.

@bkueng Why is that?

en/sensor_bus/i2c.md
hamishwillee: Thanks, I'll update

bkueng:
> @bkueng Why is that?

It's faster

en/sensor_bus/i2c.md
```diff
@@ -1 +1,38 @@
...
+
+Pixhawk/PX4 support it for:
+* Connecting off board components that require greater data rates than provided by a strict serial UART: e.g. rangefinders.
+* Compatibility with peripheral devices that only support I2C.
+* Allowing multiple devices to attach to a single bus (useful for conserving ports). For example, LEDs, Compass, rangefinders etc.
+
+> **Tip** IMUs (accelerometers/gyroscopes) should not be attached via I2C (typically the [SPI](https://en.wikipedia.org/wiki/Serial_Peripheral_Interface_Bus) bus is used). The bus is not fast enough even with a single device attached to allow vibration filtering (for instance), and the performance degrades further with every additional device on the bus.
+
+
+## Integrating I2C Devices
```
hamishwillee, line 13: @bkueng @mcharleb Looking for advice on this section.
1. Below is supposed to be very high level advice on integrating I2C hardware to work with PX4. But since I have never done it may be incorrect. Is this the right way?
2. Looking at the drivers doc it seems that it might not be. They indicate that the drivers interface is deprecated and instead teh https://github.com/px4/DriverFramework should be used. However I cannot find any I2C examples using the framework.
   - Can you clarify what the rules are
   - Any reason not to pull this readme https://github.com/px4/DriverFramework into the documentation set?
   - What else should we say.
   - What other docs should we point to -https://github.com/px4/DriverFramework ? https://dev.px4.io/en/middleware/drivers.html ?

en/sensor_bus/i2c.md
```diff
@@ -1 +1,38 @@
-# I2C Bus Overview
\ No newline at end of file
+# I2C Bus Overview
```
hamishwillee, line 1: @bkueng This section updated. Is it OK now?

en/sensor_bus/i2c.md
```diff
@@ -1 +1,38 @@
...
+```
+drivers/sf1xx
+``` 
+
+For example, search for this driver in the [px4fmu-v4_default](https://github.com/PX4/Firmware/blob/master/cmake/configs/nuttx_px4fmu-v4_default.cmake#L49) configuration
+
+
+## I2C Driver Examples
+
+To find I2C driver examples, search for **i2c.h** in [\src\drivers\](https://github.com/PX4/Firmware/tree/master/src/drivers).
```
hamishwillee: @bkueng I added sf1xx driver as it uses I2C and it is automatically built into almost every ROM. Are there any more that you would point at as "great reference examples"?

en/sensor_bus/i2c.md
bkueng: 👍

en/sensor_bus/i2c.md
```diff
@@ -1 +1,38 @@
...
+* Connecting off board components that require greater data rates than provided by a strict serial UART: e.g. rangefinders.
+* Compatibility with peripheral devices that only support I2C.
+* Allowing multiple devices to attach to a single bus (useful for conserving ports). For example, LEDs, Compass, rangefinders etc.
+
+> **Tip** IMUs (accelerometers/gyroscopes) should not be attached via I2C (typically the [SPI](https://en.wikipedia.org/wiki/Serial_Peripheral_Interface_Bus) bus is used). The bus is not fast enough even with a single device attached to allow vibration filtering (for instance), and the performance degrades further with every additional device on the bus.
+
+
+## Integrating I2C Devices
+
+Drivers should `#include <drivers/device/i2c.h>` and the provide an implementation of the abstract base class `I2C` defined in **I2C.hpp** for the target hardware (i.e. for NuttX [here](https://github.com/PX4/Firmware/blob/master/src/drivers/device/nuttx/I2C.hpp#L53)).
```
bkueng: `and the provide` Do you mean: `and then provide`?

en/sensor_bus/i2c.md
```diff
@@ -1 +1,38 @@
...
+
+Drivers will also need to include headers for their type of device (**drv_*.h**) in [\src\drivers\](https://github.com/PX4/Firmware/tree/master/src/drivers) - e.g. [drv_baro.h](https://github.com/PX4/Firmware/blob/master/src/drivers/drv_baro.h).
```
bkueng: /src/drivers/. We're not on Windows :)

en/sensor_bus/i2c.md
```diff
@@ -1 +1,38 @@
...
+
+To include a driver in firmware you must add the driver to the [cmake config file](https://github.com/PX4/Firmware/tree/master/cmake/configs) that corresponds to the target you want to build for:
+```
+drivers/sf1xx
+``` 
+
+> **Tip** For example, you can see this driver in the [px4fmu-v4_default](https://github.com/PX4/Firmware/blob/master/cmake/configs/nuttx_px4fmu-v4_default.cmake#L49) configuration.
```
bkueng: Hmm this link will probably be out-of-date very quickly when adding/removing to that list, and then pointing to the wrong line.

en/sensor_bus/i2c.md
bkueng: Maybe the `ms5611`, which has a SPI and I2C interface.

en/sensor_bus/i2c.md
bkueng: 1. Yes it looks good.
2. Initially it was the plan to switch to DriverFramework. But that never really took off, so we recently decided to make sure we can run our existing NuttX drivers on other OSes too. Which means doing the opposite what the current docs say: moving away from DriverFramework (@dagar is working on that).

en/sensor_bus/i2c.md
hamishwillee: Thanks. Fixed.

en/sensor_bus/i2c.md
hamishwillee: I am :-) . Fixed

en/sensor_bus/i2c.md
hamishwillee: Good point. I have changed to see/search for, and removed the specific line number.

hamishwillee:
Thanks @bkueng - I have made suggested changes. 
- The driver framework doc will be looked at separately in https://github.com/PX4/Devguide/issues/373
- I deleted the sf1xx and uLanding docs as these are now in user guide. The UART doc is empty, but now doesn't even group anything. Will update it to match same pattern as this I2C doc - https://github.com/PX4/Devguide/issues/374

PR #375, opened 2017-12-01
Add a note about rebooting Sapog ESCs after enumeration

keenanjohnson:
Just a quick note clarifying that Sapog firmware CAN ESCs need to be rebooted after enumeration.


bkueng:
Thanks!

PR #377, opened 2017-12-04
fix zh setup dev_env archlinux install

a-wing:
fix zh setup dev_env archlinux install


hamishwillee:
Hi @a-wing
Thanks very much for this contribution! I have merged. 

Note, unfortunately the instructions haven't been properly verified in the English for some time, and are nowhere near as complete as what we have for Ubuntu. If you're working with archlinux and can create detailed instructions for the English version, they'd be very welcome.

Issue #359, opened 2017-11-22
Proposed improvements to build scripts


hamishwillee:
The build scripts are relatively convoluted/brittle/annoying. Issues for discussion/improvement.
- They put binaries in home directories and update the paths
- Rely on random PPAs that could break underneath us
- mix apt and pip for getting python dependencies

Some of the dependencies can be obtained from reliable/default packaging in 16.04 - e.g. ninja, cmake. Where possible we should do so. 

Open question is what ubuntu versions we support via scripts and the docs. IMO we should support the  current LTS for both, but provide links on github to older versions of the docs.

hamishwillee:
@dagar For common build script I removed the line below so we use default repos
```
#sudo add-apt-repository ppa:george-edison55/cmake-3.x -y
```
For some reason this then meant that `pip install pyulog` and `pyserial` failed, so I replaced them with 
```
sudo -H pip install pyserial pyulog
```
Last of all I used apt-get to fetch ninja-build. That is v1.5 while we were using 1.6 before.

Anyway, both nuttx and jmavsim build fine. Is it worth making this change "real"? It is certainly in-line with your point that we should avoid specials repos.

dagar:
Sounds good to me. With ubuntu 16.04 CI uses whichever ninja and cmake is packaged.

PR #381, opened 2017-12-06
Update build scripts and docs to use default repos/ninja etc

hamishwillee:
This addresses https://github.com/PX4/Devguide/issues/359

### DIFF
--- a/build_scripts/ubuntu_sim_common_deps.sh
+++ b/build_scripts/ubuntu_sim_common_deps.sh
@@ -16,38 +16,19 @@ trap "exit" INT TERM; trap "kill 0" EXIT; sudo -v || exit $?; sleep 1; while tru
 echo "We must first remove modemmanager"
 sudo apt-get remove modemmanager -y
 
-# Ninja build system
-ninja_dir=$HOME/ninja
-echo "Installing Ninja to: $ninja_dir."
-if [ -d "$ninja_dir" ]
-then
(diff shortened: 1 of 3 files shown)


build_scripts/ubuntu_sim_common_deps.sh
```diff
@@ -16,38 +16,19 @@ trap "exit" INT TERM; trap "kill 0" EXIT; sudo -v || exit $?; sleep 1; while tru
...
 sudo apt-get update -y
-sudo apt-get install python-argparse git git-core zip python-empy python-toml python-numpy qtcreator cmake build-essential genromfs -y
+sudo apt-get install python-argparse git git-core zip python-empy python-toml python-numpy qtcreator cmake build-essential genromfs ninja-build -y
 # Required python packages
 sudo apt-get install python-dev -y
 sudo apt-get install python-pip -y
 sudo -H pip install --upgrade pip
 sudo -H pip install pandas jinja2
-pip install pyserial
+sudo -H pip install pyserial
```
dagar: - combine all python installs here with pip?
 - git-core is a legacy debian thing that can be dropped

hamishwillee: Thanks, I've removed git-core. I've put pyserial along with jinja2. 
I've also grouped the apt-gets as it says in the text - all the python ones are no together. 
OK to merge?

PR #382, opened 2017-12-07
Split centos and archlinux docs and update archlinux to use scripts

hamishwillee:
The instructions for the boutique linux instructions were a bit "jumpy" because they shared some common instructions, which were displayed out of order. Given that the instructions are diverging a little, it makes sense to have separate docs for CentOS and ArchLinux, and to import any commonalities rather than displaying out of order and jumping around.

This changes splits the docs. It also adds the ArchLinux script.

Outstanding tasks:
* [x] Updates as result of feedback
* [x] Add introduction to each doc covering what the toolchain instructions enable. ie for Archlinux, just PX4 firmware, not SIM. Perhaps add request for people to extend them if they know how?


en/setup/dev_env_linux_arch.md
```diff
@@ -0,0 +1,48 @@
+# Development Environment on ArchLinux
+
+## Permissions
+
+The user needs to be added to the group "uucp":
```
hamishwillee: @jlecoeur @dagar This group needed to be set in the old version. Is it still required? Is this the right group?

I can't test this as I cannot get archlinux working. Also, I did get a message that sudo command is not supported on archlinux - thoughts?

en/setup/dev_env_linux_centos.md
```diff
@@ -27,72 +19,32 @@ easy_install toml
...
-
-After that, logging out and logging back in is needed.
-
-
-> **Note** Log out and log in for changes to take effect! Also remove the device and plug it back in!**
-
-
-### GCC Toolchain Installation
-<!-- import GCC toolchain common documentation -->
-{% include "_gcc_toolchain_installation.txt" %}
```
hamishwillee, line 98: @jlecoeur I removed this section (imported) on installing the gcce toolchain from zips rather than package manager. My assumption is that the versions installed in common dependencies above replace this. OK?

en/setup/dev_env_linux_centos.md
```diff
@@ -27,72 +19,32 @@ easy_install toml
...
+<!-- import docs ninja build system -->
```
hamishwillee, line 51: @jlecoeur @dagar According to Lorenz, ninja is recommended for all platforms. The import here downloads a zip and installs in home directory. Is there a version in package manager that can be used instead? Shouldn't we have this in the scripts?

en/setup/dev_env_linux_arch.md
jlecoeur: Yes it is still needed. uucp is the arch equivalent of dialout.

Weird for sudo. If you want to install an arch VM, I recommend Antergos. It has a GUI installer that takes care of everything

en/setup/dev_env_linux_centos.md
jlecoeur: Yes, latest arm-none-eabi is installed via pacman in the script.
Now question: I have been flying gcc7-compiled firmwares for a while, but this is not officially supported, should it be recommended to use gcc7 in the docs?

en/setup/dev_env_linux_centos.md
jlecoeur: Yes
sudo pacman -S ninja

en/setup/dev_env_linux_arch.md
```diff
@@ -0,0 +1,48 @@
...
+## Script-based Installation
+
+Once ArchLinux is installed you can use our standard docker script 
+[archlinux_install_script.sh](https://github.com/PX4/containers/blob/master/docker/px4-dev/scripts/archlinux_install_script.sh)
+to install all dependencies required for building PX4 firmware.
+
+To install using this script, enter the following in a terminal:
+```sh
+wget https://raw.githubusercontent.com/PX4/containers/master/docker/px4-dev/scripts/archlinux_install_script.sh
+source ./archlinux_install_script.sh
```
jlecoeur: This needs sudo

en/setup/dev_env_linux_arch.md
```diff
@@ -0,0 +1,48 @@
...
+    
+# Install Python dependencies
+pip install serial empy numpy toml jinja2
+
+# Install genromfs
+wget https://sourceforge.net/projects/romfs/files/genromfs/0.5.2/genromfs-0.5.2.tar.gz
+tar zxvf genromfs-0.5.2.tar.gz
+cd genromfs-0.5.2 && make && make install && cd ..
+rm genromfs-0.5.2.tar.gz genromfs-0.5.2 -r 
+```
```
jlecoeur, line 56: genromfs is also available in AUR. The previous doc recommended to install yaourt to install it. This is fine but yaourt is not usable in a Dockerfile. That is why *archlinux_install_script.sh* installs genromfs from source.

I would add a note like:
Note: Alternatively, genromfs is available in AUR (Archlinux User Repository) https://aur.archlinux.org/packages/genromfs/
You can install it using `yaourt -S genromfs`

en/setup/dev_env_linux_centos.md
hamishwillee: Can you add to the script?

en/setup/dev_env_linux_centos.md
hamishwillee: @dagar Is there a plan to support/use gcc7 as per @jlecoeur post above? Or should the script use the same downloaded version as ubuntu? IMO if we can use standard package manager versions everywhere that is better.

hamishwillee:
Thanks @jlecoeur - I work other jobs on Friday, so I will look at this properly on Monday.

en/setup/dev_env_linux_centos.md
jlecoeur: Done

en/setup/dev_env_linux_centos.md
hamishwillee: I think for docs, we can't recommend anything other than official version. I'll see if I can get that working.

en/setup/dev_env_linux_centos.md
hamishwillee: Thanks. I'll remove my "manual" installation instructions :-)

hamishwillee:
FYI, I tried to install the jmavsim dependencies. I can get things to build fine using:
```
sudo pacman -Sy --noconfirm apache-ant jre8-openjdk jdk8-openjdk
```
Unfortunately this doesn't actually work - something to do with x11.

en/setup/dev_env_linux_arch.md
hamishwillee: OK. I tried this again. It only works if I do
```
sudo -s
source ./archlinux_install_script.sh
```
Something to do with source being a shell command.

Anyway, will fix up docs appropriately.

hamishwillee:
@jlecoeur . Unfortunately I found another dependency - Fast RTPS. This does not build with the standard ubuntu instructions. Can you put that in the docker script too?

en/setup/dev_env_linux_arch.md
hamishwillee: Fixed. The problem is fixed if instructions are preceded by `sudo -s`

en/setup/dev_env_linux_arch.md
hamishwillee: Added note below, as you suggested.

hamishwillee:
I tried to get fastRTPS and the official GCCE to work, but neither build succeeds. The current docs based on @jlecoeur docker works, but I feel awkward recommending something using the unsupported version when I can't explain how to install the correct version. More generally, if we say something works, would be nice to have some easy mechanism for validating that. 

@dagar Is it possible to get a docker build for archlinux that installs both fastRTPS and the supported GCCE? I can then update docs appropriately. 

For bonus points, would be excellent if it could also include jmavsim (which builds, but can't run due to x11 issues.

hamishwillee:
@jlecoeur @dagar Merry Christmas. I want to get this in, because even though it is not ideal it is better than what we have. 

To make this "acceptable" I have added a disclaimer to the top of the docs explaining exactly what this delivers and how this diverges from the ideal (uses GCC from package manager and does not build RTPS, and that we hope to fix in future). 

Are you guys OK to merge with those changes?

dagar (approved):

dagar:
Yes let's get it in and continue iterating. 

Next year let's work towards unifying documentation, docker, CI, and developers. These detailed instructions tend to drift faster than you'd think. Forcing everything to be on the same page is the only way I can think of keeping it accurate.

LorenzMeier:
That makes sense. We're cleaning up anyways a bunch of things early 2018 and so forcing a consistent install should be part of it.

PR #386, opened 2017-12-13
add debugger to simulator startup info

hamishwillee:
Updates simulator startup information to explain:
1. Specifying debugger as well as simulator and model in the startup
2. How to list all vmd targets: `list_vmd_make_targets`
3. The existence of defaults for most values
4. How to use three underscores in a target.

### DIFF
--- a/en/simulation/README.md
+++ b/en/simulation/README.md
@@ -84,23 +84,40 @@ make posix_sitl_ekf2 gazebo_plane
 The syntax to call `make` with a particular configuration and initialisation file is:
 
 ```bash
-make [CONFIGURATION_TARGET] [SIMULATOR]_[INIT_FILE]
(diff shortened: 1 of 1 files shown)


hamishwillee:
@bkueng Only thing I think might be missing is the use of "none" in a vmd triplet (can it be used in the configuration target too?). Can you explain how this works/can be used?

bkueng (approved):
`none` is very handy for quick testing & restarts: if you start jmavsim yourself with `./Tools/jmavsim_run.sh` and keep it running, you can just (re)start px4 with the `none` target, and it will connect to the simulator.

hamishwillee:
Thanks! I added a note for this

- You can use a `none` value for `VIEWER_MODEL_DEBUGGER` to start PX4 and wait for a simulator. 
  For example start PX4 using `make posix_sitl_default none` and jMAVSim using `./Tools/jmavsim_run.sh`

Issue #387, opened 2017-12-13
Document other supported bus types


hamishwillee:
We have either docs or placeholders for UAVCAN, I2C, UART. What other types of bus do we have?
SPI? What else?

-----
Generally we should follow the pattern for these of [I2C](https://dev.px4.io/en/sensor_bus/i2c.html):
- What is bus. 
- How does it differ from other alternatives
- What is it good for - ie what peripherals, on main board or off board
- Whether it supports multiple devices on bus or just one
- Standard connectors
- How connected for standard boards.
- Any issues of powering
- Any limitations of our support on the bus - ie do we have only one port supported on a device.

Then developer integration questions
- How do we integrate with it
- What does the driver need to derive from
- How is driver included in hardware
- How is driver enabled.
- How are settings handled - ie typical params associated with this
- If there are multiple ports for this bus, how do you specify which one device is attached to.

hamishwillee:
@bkueng This is a bit of a placeholder for further work. At this point I was wondering if you could just let me know what busses we have that are not documented.

bkueng:
That looks pretty good. I don't have any other bus. (PWM can be used as input/output as well, but it's not a bus).

hamishwillee:
Do you think we should perhaps then change section to *Sensor and Actuator I/O*? Or are they better separate?

bkueng:
Yes, I think it makes sense to combine them.

hamishwillee:
> Yes, I think it makes sense to combine them.

Thanks - fixed  in #395

hamishwillee:
Notes for when documenting PWM Ports (From @bkueng here: https://github.com/PX4/Devguide/pull/395#discussion_r157917183 )

Re "also the main PWM outputs."
> What I/O is typically run on this port? Motor control obviously and other actuators (e.g. parachutes). But what about input?

Lidar Lite, or some ultrasonic sensors.

> What are the mechanics of integrating a sensor on these ports. Specifically, say I want to add X sensor to an AUX port, am I again writing a device driver that reads the inputs and outputs uORB topics?

Basically yes, but I think it's somewhat involved. I think it's the easiest to point to an existing driver

> What about actuators - say I want to open a brand new type of gripper?

actuators are simpler: you choose an actuator controls topic and then configure the mixer accordingly (I think it's documented on the mixer page).

PR #395, opened 2017-12-20
Rename bus section to I/O and add explanation of section

hamishwillee:
Make it clear sensor section covers both bus and main PWM port. Do this by renaming to I/O in topic name and also adding explanation of the purpose of the section.

### DIFF
--- a/en/sensor_bus/README.md
+++ b/en/sensor_bus/README.md
@@ -1,3 +1,3 @@
 # Sensor and Actuator I/O
 
-This section contains topics about integrating sensors and actuators into PX4. It covers both sensor busses ([I2C](../sensor_bus/i2c.md), [UAVCAN ](../uavcan/README.md), [UART](../sensor_bus/uart.md), SPI) and also the main PWM outputs.
+This section contains topics about integrating sensors and actuators into PX4. It covers both sensor busses ([I2C](../sensor_bus/i2c.md), [UAVCAN ](../uavcan/README.md), [UART](../sensor_bus/uart.md), SPI) and also the main PWM ports.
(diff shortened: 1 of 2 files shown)


en/sensor_bus/README.md
```diff
@@ -1 +1,3 @@
-# Sensor and Actuator Busses
\ No newline at end of file
+# Sensor and Actuator I/O
+
+This section contains topics about integrating sensors and actuators into PX4. It covers both sensor busses ([I2C](../sensor_bus/i2c.md), [UAVCAN ](../uavcan/README.md), [UART](../sensor_bus/uart.md), SPI) and also the main PWM outputs.
```
hamishwillee: @bkueng 
1. Re "also the main PWM outputs."
   - Is this the right/best name for the AUX/MAIN ports? (probably should change to I/O rather than outputs too)
   - What I/O is typically run on this port? Motor control obviously and other actuators (e.g. parachutes). But what about input?
   - What are the mechanics of integrating a sensor on these ports. Specifically, say I want to add X sensor to an AUX port, am I again writing a device driver that reads the inputs and outputs uORB topics?
   - What about actuators - say I want to open a brand new type of gripper?

bkueng: - You could name it 'PWM ports'
- input: Lidar Lite, or some ultrasonic sensors.
- Basically yes, but I think it's somewhat involved. I think it's the easiest to point to an existing driver
- actuators are simpler: you choose an actuator controls topic and then configure the mixer accordingly (I think it's documented on the mixer page).

hamishwillee: Thank you. I've updated to "PWM ports" and taken notes about what the associated topic will have to say in https://github.com/PX4/Devguide/issues/387#issuecomment-353212719

I will also merge on the assumption this is otherwise OK.

PR #396, opened 2017-12-20
Merge daemon docs into Architecture doc

hamishwillee:
This deletes the daemons doc and moves it into the high level architecture (this is the result of a discussion with @bkueng earlier in the month). 
As "Middleware & Architecture" no longer contains architecture topics this now has been changed to just "Middleware". Will consider renaming to "Communications" in future.


en/concept/architecture.md
```diff
@@ -119,34 +119,53 @@ PX4 runs on various operating systems that provide a POSIX-API
...
+  Work queues are used for periodic tasks, 
+  such as sensor drivers or the land detector.
 
 > **Note** Tasks running on a work queue do not show up in `top` 
 > (only the work queues themselves can be seen - e.g. as `lpwork`).
 
-The following sections provide additional OS-specific information.
 
-### NuttX
+### Background Programs/Daemons
```
hamishwillee: @bkueng Note that the terminology is not very "comfortable here". A daemon is commonly referred to as a background process, but we're also talking here about tasks, threads, and processes, where "process" has a different an distinct meaning, 
Anyway, I'm calling them background programs, which isn't perhaps all that great either.

bkueng: Yes, I can imagine that this section could be rather confusing. Maybe we should reduce it to just something like:
'`px4_task_spawn_cmd` is used to launch new tasks that run independently from the calling (parent) task'?
Plus the API example.

hamishwillee: Good idea. I've used your words and removed the words process and daemon (I'm OK with that because I don't think that is how people will search for the concept). See  https://github.com/PX4/Devguide/pull/396/commits/49ca6e2e78cdc2c7f9f3b2a7c380e921bffff2e2

You otherwise ok to merge this?

bkueng (approved):

PR #400, opened 2017-12-21
Controller diagram

hamishwillee:
Add position controller diagram from https://github.com/PX4/Firmware/issues/8316. You can see this rendered [here](https://hamishwillee.gitbooks.io/havdevguide/content/v/33be49513ca4fa6aca9d96053bc6725e9e054cf7/en/flight_stack/controller_diagrams.html).

@Stifael @bresch This is primarily for discussion. We might for example decide that:
- We want individual documents for each diagram
- We want a document for each controller, which happens to include the diagram. 

Depends on what else is worth saying about controllers - e.g. What would you put in slides if you had to present on this in a webinar say :-) Links to source code etc?

@bresch there is also a placeholder doc for the notation. If you provide me the source text, I can add it (up to you).


hamishwillee:
@bresch You may be on holiday, but if not, can you send me your notation doc? I'd rather get this  merged and then iterate than wait on perfection.

@Stifael Any thoughts on discussion in https://github.com/PX4/Devguide/pull/400#issue-283772927 . Specifically, what else would you like said here or is it OK? Is there a lot more that we might say in a position controller document such that this would just be part of that document?

bresch:
> can you send me your notation doc?

The glossary? Here is its [repo](https://github.com/bresch/Glossary)

PR #401, opened 2017-12-21
Add dronecode high level diagram

hamishwillee:
High level diagram showing future looking dronecore platform. 

Image, content, and location are all up for discussion!


en/concept/dronecode_architecture.md
```diff
@@ -0,0 +1,9 @@
+# Dronecode Platform Hardware/Software Architecture
+
+The diagram below provides a forward-looking high level overview of the Dronecode Platform. The left hand side shows ones possible hardware configuration with a *flight controller* (light blue) connected to a *perception computer* (dark blue) via RTPS. The perception computer provides vision control and object avoidance using a camera sensor array, and has a separate payload camera.
```
hamishwillee, line 3: "one possible", not one's possible. @bkueng What sort of information do you think would be useful in this section? Perhaps talk about the possible architectures for the perception computer. Also maybe expand on the different levels - providing links at various parts? i.e. to DroneCore, to PX4 architecture diagram etc?

Feel free to suggest content!

LorenzMeier (approved):

en/concept/dronecode_architecture.md
bkueng: Adding links sounds good. Especially for ROS/DroneCore/RTPS.
For the perception computer there are probably still too many unknowns to already document it in more detail. Once we have the avoidance in place, that will be something we can add.

PR #402, opened 2017-12-25
Fix for mRo X2.1

tops4u:


dagar (approved):

mhkabir:
Why was this merged? This seems to be incorrect.

mhkabir:
The correct target is `px4fmu-v3_default`.

mhkabir:
mRo X2.1 is this : https://store.mrobotics.io/mRo-X2-1-Rev-2-p/mro-x2.1rv2-mr.htm. 

The one changed in this PR is a standard Pixhawk board.

dagar:
My bad, I was reviewing from my phone. Good catch Kabir.

dagar:
Do we even need to mention the mRo pixhawk separately from the pixhawk 1?

mhkabir:
Not really, I guess.

tops4u:
FMUv3 build will not load on the X-2.1 since it has a different Board identifier. AFAIK FMUv3 is similar to v2 except the Silicon Version? X-2.1 has different IMU HW and slso different Orientation of sensors than FMUv2.

tops4u:
Oh my bad... when I was looking through the Docu I was just looking out for mRo, but indeed this line is for the Standard Pixhawk from mRo. So we should add a line for the X-2.1 Variant. Shall I issue a new PR?

LorenzMeier:
Please do a new PR, yes.

PR #404, opened 2017-12-26
Added Entry for mRo X-2.1

tops4u:
mRo X-2.1 (former AUAV X-2.1) is a separate Platform that has its own Build Target.


tops4u:
@mhkabir  & @dagar sorry for the confusion with the earlier PR.

mhkabir:
Thanks, this is good now!

PR #407, opened 2017-12-28
Add Gazebo section for HITL

arwagoner:


en/simulation/hitl.md
```diff
@@ -26,7 +28,35 @@ These parameter are optional and allow to run HITL without a radio remote contro
...
+
+### Using Gazebo
+* Update the environment variables: 
+``` 
+cd <Firmware_clone>
+make posix_sitl_default gazebo
+source Tools/setup_gazebo.bash $(pwd) $(pwd)/build/posix_sitl_default
+```
+
+* Open the vehicle model's sdf file (e.g. Tools/sitl_gazebo/models/iris/iris.sdf).
```
hamishwillee, line 49: Iris is a [Quadrotor Wide](https://docs.px4.io/en/airframes/airframe_reference.html#quadrotor-wide) - does it matter that there is a mismatch between this model and the options you can select in QGC (Quadrotor + and x)?

en/simulation/hitl.md
```diff
@@ -26,7 +28,35 @@ These parameter are optional and allow to run HITL without a radio remote contro
 
 ## Connect to the Simulator
 
-Currently either jMAVSim or X-Plane and QGroundControl are supported. Follow either setup step below.
+Currently Gazebo, jMAVSim, or X-Plane and QGroundControl are supported. Follow either setup step below.
```
hamishwillee, line 39: The docs don't say when/how you should connect the autopilot/vehicle to the computer. Does it matter?

hamishwillee:
Thanks for this. There a are a couple of questions inline. 

Once those are answered I will merge this and do a subedit for the rendering/structure - easier for me to do that explain.

As part of that I'll do some other general work:
- Cross link to https://dev.px4.io/en/simulation/#hitl-simulation-environment - perhaps even reproduce the image.
- Update QGC images to latest
- Move the information about what we support to the top of the page ("PX4 supports HITL for multicopters ....")

## Questions

You may be able to help with improving this documentation ...

As I understand it, the way this works is that you enable HITL mode, and then you can in theory select any airframe and it will be used in HITL mode. However we have a number of specific HITL targets in QGC, and those are the ones that we know work. These correspond to the ones we list in the doc (X-frame multicopters (autostart IDs 4001, 4011), Standard AERT planes, HITL X-frame multicopter (autostart ID 1001), HITL AERT plane (autostart ID 1000))
- Is that correct?
- What is likely to happen if you use an unsupported frame with HIL enabled?
- What makes a frame "HIL Enabled" - ie say I want to work on a VTOL, what do I have to do to make it work properly?
- What does HIL Enabled actually do? Fake sensor data?

More generally, it would be good to add an overview of what our HITL solution "does" and how it compares to SITL. 
So for example the assumption I make is that the simulator provides all sensor data for PX4, and that this is fed in from the MAVLink Simulation API, and then into uORB topics. I assume that the HITL is therefore exactly like really flying but with data from simulator. 

However @dagar seemed to indicate some data has been "faked" somehow and that there are limitations. What are they?

en/simulation/hitl.md
arwagoner: It does not appear to make a difference.

en/simulation/hitl.md
arwagoner: I think it does matter when you are first setting it up. I also forgot to add an important step for QGC configuration. I will add these steps in.

arwagoner:
> What is likely to happen if you use an unsupported frame with HIL 

I tested HIL with a standard vtol airframe (SYS_AUTOSTART=13008). It did not behave as expected and did not appear to respond to any landing commands. Maybe @dagar or @TSC21 can give us more insight into why that happened and what the differences are between the HIL enabled frames and the other frames?

dagar:
HITL and SITL simulation data enter the system at different points. Core modules like commander and sensors have HIL modes at startup that bypasses some of the normal functionality. In SITL we have fake simulation drivers, and the rest of the system behaves normally.

hamishwillee:
@dagar Thanks. Can we delve a little?

> HITL and SITL simulation data enter the system at different points. 

The [diagrams](https://dev.px4.io/en/simulation/) indicate that they come in via different paths - ie SITL and Gazebo communicate direct via  UDP while HITL routes via QGC but they both use the simulation API and then generate appropriate uORB topics. 
So does this make any difference, and if so, what?

> Core modules like commander and sensors have HIL modes at startup that bypasses some of the normal functionality. In SITL we have fake simulation drivers, and the rest of the system behaves normally.

Do you know approximately what functionality is "missing" in HITL and what difference that makes? 

Basically I want to be able to make broad statements about where users are likely to unexpectedly find that HITL is not the better platform for simulation.

hamishwillee:
@arwagoner Thanks for updates. Would I be correct in thinking that the Gazebo support is from 1.8? Ie I should either have a mechanism to merge this in after 1.8 appears OR add a note to say that this appears in 1.8.?

arwagoner:
@hamishwillee The intent is for it to be merged upstream as soon as possible.

hamishwillee:
@arwagoner Is that the same as saying "probably PX4 7.3?" IMO this can go in as soon as the associated code is in a released firmware (and I will do my tidy up at that point).

@dagar Would be great to get detail on https://github.com/PX4/Devguide/pull/407#issuecomment-354689325 - at the moment I have an unsatisfying feeling that HITL might suck, but that isn't very helpful or clear to say.

hamishwillee:
I have updated HITL docs in #417. Would appreciate review/answers to open questions there.

PR #411, opened 2018-01-02
Add notation from@bresch

hamishwillee:
Add notation, kindly donated by @bresch - https://github.com/bresch/Glossary. This is rendered using the mathjax plugin.

This originates from https://github.com/PX4/Devguide/pull/400

[Rendered view here](https://hamishwillee.gitbooks.io/havdevguide/content/v/ba86638ccf23a93527fab26ae21562dd3190c641/en/contribute/notation.html). I had to refresh page to get the maths to display - anyone else see the same thing?

@Stifael @bresch anything to add/remove? Otherwise I'll merge in the next day or so.

### DIFF
--- a/en/contribute/notation.md
+++ b/en/contribute/notation.md
@@ -1,6 +1,11 @@
-# Terminology/Notation
+# Terminology
 
 The following terms, symbols, and decorators are used in text and diagrams throughout this guide.
+
+## Notation
+- Bold face variables indicate vectors or matrices and non-bold face variables represent scalars. 
+- The default frame for each variable is the local frame $$\ell$$. Right superscripts represent the coordinate frame. If no right superscript is present, then the default frame $$\ell$$ is assumed. An exception is given by Rotation Matrices, where the lower right subscripts indicates the current frame and the right superscripts the target frame.
+- Variables and subscripts can share the same letter, but they always have different meaning.
 
 ## Acronyms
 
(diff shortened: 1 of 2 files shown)


LorenzMeier (approved):

en/contribute/notation.md
```diff
@@ -2,11 +2,65 @@
...
+$$\beta$$ | Angle of sideslip (AOS).
+$$\bar{c}$$ | Mean aerodynamic chord (mac).
+$$c$$ | Wing chord length.
+$$\delta_{a,e,r}$$ | Aerodynamic control surface angular deflection. Subscripts $$a$$, $$e$$ and $$r$$ stand for *aileron*, *elevator* and *rudder*, respectively. A positive deflection generates a negative moment.
+$$\Psi$$ | Attitude vector. $$\Psi = [\phi \quad \theta \quad \psi]^T$$.
+$$\phi$$ | Roll euler angle. Also named *Bank angle* in aviation.
+$$\psi$$ | Yaw euler angle. Also named *Heading*.
+$$\theta$$ | Pitch euler angle.
+$$\boldsymbol{\mathrm{F}}$$ | Force vector. $$\boldsymbol{\mathrm{F}} = [X \quad Y \quad Z]^T$$.
+$$\boldsymbol{\mathrm{F}}_{Aero}^w$$ | Aerodynamic forces in wind frame. *Lift* $$L$$, *drag* $$D$$ and *cross-wind force* $$C$$. $$\boldsymbol{\mathrm{F}}_{Aero}^w = [-D \quad -C \quad -L]^T$$.
```
Stifael: I would keep F as symbol, Aero etc. under subscripts.

Stifael: I would just use F as symbol for force, and aero etc as subscripts

Stifael (changes requested):

Stifael:
I would keep decorators, symbols and other definition completely separated. 
 For instance: 
Under symbol I only would keep a non-bold face F for Force instead of writing down each specific force itself. Under subscripts I would then add a T (= Thrust) and Aero (aerodynamic force). I only would add boldface symbols where it has no other meaning (for instance attitude). 
I also would define on a default frame. For instance, the default frame is local, which means that any vector that does not have a superscript, is automatically described in local frame. Otherwise, a little "b" has to added if it is specific. 
Let me do a revision of if.

bresch:
@Stifael How would you do to define the components of the aerodynamic forces for example? Add _D_, _C_ and _L_ in the symbols list and make a new table with the definition of special vectors (relative airspeed, body rates, ..)?

en/contribute/notation.md
```diff
@@ -2,11 +2,65 @@
...
+$$\boldsymbol{\mathrm{M}}_T^b$$ | Body thrust moments. $$\boldsymbol{\mathrm{M}}_T^b = [\ell_T \quad m_T \quad n_T]^T$$.
+$$M$$ | Mach number. Can be neglected for scale aircrafts.
+$$\boldsymbol{\mathrm{\tilde{q}}}$$ | Hamiltonian attitude quaternion. $$\boldsymbol{\mathrm{\tilde{q}}} = (q_0, q_1, q_2, q_3) = (q_0, \boldsymbol{\mathrm{q}})$$.<br>A vector in the local NED frame $$\ell$$ can be represented in the body frame $$b$$ using $$\boldsymbol{\mathrm{\tilde{v}}}^b = \boldsymbol{\mathrm{\tilde{q}}} \, \boldsymbol{\mathrm{\tilde{v}}}^\ell \, \boldsymbol{\mathrm{\tilde{q}}}^*$$ (or $$\boldsymbol{\mathrm{\tilde{q}}}^{-1}$$ instead of $$\boldsymbol{\mathrm{\tilde{q}}}^*$$ if $$\boldsymbol{\mathrm{\tilde{q}}}$$ is not unitary).  $$\boldsymbol{\mathrm{\tilde{v}}}$$ represents a *quaternionized* vector: $$\boldsymbol{\mathrm{\tilde{v}}} = (0,\boldsymbol{\mathrm{v}})$$.
+$$\boldsymbol{\mathrm{r}}$$ | Position vector $$\boldsymbol{\mathrm{r}} = [x \quad y \quad z]^{T}$$.
+$$\boldsymbol{\mathrm{R}}_a^b$$ | Rotation matrix. Rotates a vector from frame $a$ to frame $b$. $$\boldsymbol{\mathrm{v}}^b = \boldsymbol{\mathrm{R}}_a^b \boldsymbol{\mathrm{v}}^a$$.
```
bresch: _a_ and _b_ frames ares displayed `$a$` and `$b$`. Double `$` has been forgotten.

en/contribute/notation.md
```diff
@@ -2,11 +2,65 @@
...
+$$\boldsymbol{\mathrm{F}}_T^b$$ | Thrust force in body frame. $$\boldsymbol{\mathrm{F}}_T^b = [X_T^b \quad Y_T^b \quad Z_T^b]^T$$.
+$$\boldsymbol{\mathrm{g}}$$ | Gravity vector.
+$$\boldsymbol{\mathrm{M}}_{Aero}^b$$ | Body aerodynamic moments. $$\boldsymbol{\mathrm{M}}_{Aero}^b = [\ell \quad m \quad n]^T$$.
+$$\boldsymbol{\mathrm{M}}_T^b$$ | Body thrust moments. $$\boldsymbol{\mathrm{M}}_T^b = [\ell_T \quad m_T \quad n_T]^T$$.
+$$M$$ | Mach number. Can be neglected for scale aircrafts.
+$$\boldsymbol{\mathrm{\tilde{q}}}$$ | Hamiltonian attitude quaternion. $$\boldsymbol{\mathrm{\tilde{q}}} = (q_0, q_1, q_2, q_3) = (q_0, \boldsymbol{\mathrm{q}})$$.<br>A vector in the local NED frame $$\ell$$ can be represented in the body frame $$b$$ using $$\boldsymbol{\mathrm{\tilde{v}}}^b = \boldsymbol{\mathrm{\tilde{q}}} \, \boldsymbol{\mathrm{\tilde{v}}}^\ell \, \boldsymbol{\mathrm{\tilde{q}}}^*$$ (or $$\boldsymbol{\mathrm{\tilde{q}}}^{-1}$$ instead of $$\boldsymbol{\mathrm{\tilde{q}}}^*$$ if $$\boldsymbol{\mathrm{\tilde{q}}}$$ is not unitary).  $$\boldsymbol{\mathrm{\tilde{v}}}$$ represents a *quaternionized* vector: $$\boldsymbol{\mathrm{\tilde{v}}} = (0,\boldsymbol{\mathrm{v}})$$.
```
bresch: I think I made a mistake here. The attitude quaternion represents the attitude of the body frame relative to the local frame. This means that the operation `q v q*` rotates from body to local frame (https://github.com/PX4/Matrix/blob/e595ebb9a704c24aeae990dac768d26949fcaee0/matrix/Quaternion.hpp).
@Stifael Could you confirm?

bresch (changes requested):

en/contribute/notation.md
Stifael: @bresch. 
Yes. the quaternion q represents the orientation relative to the local frame.
By the way, I also think it is not necessary to to add the tilde. In fact, what is even the meaning of the tiled?

bresch: Thanks.
The tilde means that this is a quaternion or a "quaternionized" vector. In my definition, this is needed since the bold **q** only represents the vector part of the quaternion. This should also be added in the decorators.

Stifael:
@bresch 
> @Stifael How would you do to define the components of the aerodynamic forces for example? Add D, C and L in the symbols list and make a new table with the definition of special vectors (relative airspeed, body rates, ..)?

D, C and L are just symbols where each has its own entry. Not sure why a new table is required for special vectors. All it needs is a notation definition that says that if a symbol is bold-face, then it is either a vector or a matrix. In addition, a definition for superscripts and subscripts is beneficial, where superscripts are frames, subscripts are descriptions. Consequently, if someone writes (boldface)F_airspeed, then it should be clear that the variable in question is a Force vector in local frame (local is default because no b is provided) where the description is airspeed.

bresch:
@Stifael I agree, but I'm still missing something: 
Where do you specify that the aerodynamic forces resolved in the wind frame is [-D -C -L]^T? 
It's not obvious.

Stifael:
let me do a PR to clarify it.

Stifael:
@hamishwillee If I opened your link that shows the rendering the first time everything was rendered correctly. However, If I now make changes and let it render on my own copy of px4devguide, I have the same issue that I need to refresh it to get the math rendered correctly.

hamishwillee:
Thanks all. @Stifael your fixes merged and make sense to me. I added subheadings before the subscript/superscript tables to make them more visible. 

Regarding rendering, I added a script to the page to force the typesetting to render on page load. This change would be better if added to the plugin: https://github.com/GitbookIO/plugin-mathjax/issues/34#issuecomment-355155398 but I can't reliably reproduce the error so hoping someone else will fix that ...

Anyone got any objections to this being merged?

Stifael:
no objection

@bresch 
> Where do you specify that the aerodynamic forces resolved in the wind frame is [-D -C -L]^T

You are right, that is definitely not obvious. The same is true that it is not obvious how to transform a vector from body frame to local frame or vice versa. However, I think it is also important that notation stays independent of mathematical construct. For instance, if you document a mathematical implementation, then you need to explain the mathematical construct anyway. Consequently, you will mention the aerodynamic force vector F_Aero and fill it with the symbols required. By the way, I am also not entirely sure if it is the best idea to have an example of mathematical transformation within the glossary such as quaternion rotation.

Stifael:
https://github.com/hamishwillee/Devguide/pull/2

hamishwillee:
OK. I am going to merge this now. Feel free to create your own additional PRs to fix.

PR #417, opened 2018-01-04
Hitl updates

hamishwillee:
This does an overall rebuild of the HITL story, tidying up new additions from #407 but also improving the whole picture. 

Some changes:
1. Simulation Environment docs moved out of parent into this doc (and cross linked).
2. Updated images that use latest QGC and focus better on actual change required.
3. General text improvements. 
4. Clear mapping of simulator to vehicle supported. 

The section on X-Plane is tidier, but still not revisited/re-tested.


en/simulation/README.md
```diff
@@ -225,14 +225,7 @@ For more information about using the MAVLink module see [Modules Reference: Comm
 
 With Hardware-in-the-Loop (HITL) simulation the normal PX firmware is run on real hardware. *QGroundControl* is connected to the physical hardware over USB and acts as a gateway to forward data between the simulator, PX4 and any offboard API.
 
-The diagram below shows a typical HITL simulation environment:
```
hamishwillee, line 228: FYI makes sense to move this content to HITL doc and just link from here because it is relevant to understanding how HITL is set up.

en/simulation/hitl.md
```diff
@@ -1,101 +1,133 @@
...
-## Step 1: Enable HITL mode and configure airframe
+PX4 supports HITL for multicopters (using jMAVSim or Gazebo) and fixed wing (using X-Plane demo/full version or Gazebo). 
 
-Connect the autopilot directly to QGroundControl.
 
-First enable HITL mode in the safety setup as shown in the screenshot below. Select the "Enabled" in the drop down on the HITL enabled section \(first on the page\):![QGroundControl HITL configuration](../../assets/qgc_hitl_config.png)
+## HITL-Compatible Airframes {#compatible_airframe}
 
-PX4 supports HITL for multicopters \(using jMAVSim or Gazebo\) and fixed wing \(using X-Plane demo or full version or Gazebo\). Generally any compatible airframe can be put into HITL mode. Compatible airframes are right now:
+The current list of compatible airframes vs Simulators is:
```
hamishwillee, line 10: @dagar 
1. What makes a simulator _compatible_? I.e. what change would we have to make to another frame to ensure it is supported?
2. Or conversely, if I look at configuration, how can I tell that something is properly HITL enabled?
3. Or even further - why not all of them for Gazebo?

en/simulation/hitl.md
```diff
@@ -1,101 +1,133 @@
...
-![](../../assets/sim/QGC_AutoConnect.png)
+![HITL setup](../../assets/simulation/px4_hitl_overview.png)
 
-Once configuration in QGroundControl is complete, close QGroundControl and disconnect the autopilot from the computer. 
+> **Tip** For more information see: [Simulation](../simulation/README.md).
 
-## Configure Joystick and Failsafe \(Optional\)
 
-These parameter are optional and allow to run HITL without a radio remote controller connected. Use the search bar on the top of the parameter editor to bring them up.
+## HITL vs SITL
```
hamishwillee, line 36: @dagar Would be good if you could check this section. I would have preferred something that says a bit more precisely the relative "accuracy" but I think this at least removes the impression that HITL is a "perfect" autopilot simulation.

en/simulation/hitl.md
```diff
@@ -1,101 +1,133 @@
...
-```
-gazebo /Tools/sitl_gazebo/worlds/iris.world
-```
-* Then run QGroundControl - it will auto-connect.
+1. (Optional) Configure Joystick and Failsafe.
 
-### Using jMAVSim \(Quadrotor only\)
+   Set the following [parameters](https://docs.px4.io/en/advanced_config/parameters.html#finding-a-parameter) in order to use a joystick instead of an RC remote control transmitter:
+   * [COM_RC_IN_MODE](../advanced/parameter_reference.md#COM_RC_IN_MODE) to "Joystick/No RC Checks". This allows joystick input and disables RC input checks.
+   * [NAV_DLL_ACT](../advanced/parameter_reference.md#NAV_DLL_ACT) to "Disabled". This ensures that no RC failsafe actions interfere when not running HITL with a radio control.
```
hamishwillee, line 71: @arwagoner @dagar  is NAV_DLL_ACT always needed if you use COM_RC_IN_MODE - shouldn't this automatically be set if COM_RC_IN_MODE is set to ""Joystick/No RC Checks"? I ask mostly because the QGC docs for Joystick do not mention this parameter.

en/simulation/hitl.md
```diff
@@ -1,101 +1,133 @@
...
+> **Note** Make sure *QGroundControl* is not running!
+
+1. Update the environment variables: 
+  ```sh
+  cd <Firmware_clone>
+  make posix_sitl_default gazebo
+  source Tools/setup_gazebo.bash $(pwd) $(pwd)/build/posix_sitl_default
   ```
-  ./Tools/jmavsim_run.sh -q -d /dev/ttyACM0 -b 921600 -r 250
+1. Open the vehicle model's sdf file (e.g. **Tools/sitl_gazebo/models/iris/iris.sdf**).
```
hamishwillee: @arwagoner What is this sdf file? If I go to **\Firmware\Tools\sitl_gazebo\models\iris** all I see is **model.config**

en/simulation/hitl.md
```diff
@@ -1,101 +1,133 @@
...
+1. Under the `mavlink_interface plugin` section, change the `serialEnabled` and `hil_mode` parameters to `true`. 
+
+  ![HIL Parameters](../../assets/simulation/gazebo_sdf_model_hil_params.png)
+1. Replace the `serialDevice` parameter (`/dev/ttyACM0`) if necessary.
```
hamishwillee: @arwagoner How will I know if changing serialDevice is necessary? Is this also set in the SDF file? Is it obvious - ie can I just search for it, or is the section name useful?

en/simulation/hitl.md
```diff
@@ -1,101 +1,133 @@
...
-#### Enable HITL in QGroundControl
+1. In **Settings > Net Connections**, in the *Data* tab, set localhost and port 49005 as IP address, as shown below:
 
-Widgets -&gt; HIL Config, then select X-Plane 10 in the drop-down and hit connect. Once the system is connected, battery status, GPS status and aircraft position should all become valid:
+   ![X-Plane network configuration](../../assets/gcs/xplane_net_config.png)
 
-![](../../assets/gcs/qgc_sim_run.png)
+1. Enable HITL in QGroundControl
+   1. Open *QGroundControl*
+   1. Open **Widgets > HIL Config**. Select X-Plane 10 in the drop-down and hit connect. Once the system is connected, battery status, GPS status and aircraft position should all become valid:
```
hamishwillee: @dagar A bit annoying that X-Plane requires specific setup. Should I change this to "Enable X-Plane in QGC"?

en/simulation/hitl.md
```diff
@@ -1,101 +1,133 @@
...
+      ![](../../assets/gcs/qgc_sim_run.png)
 
-Switch to the flight planning view and put a single waypoint in front of the plane. Click on the sync icon to send the waypoint.
 
-Then select MISSION from the flight mode menu in the toolbar and click on DISARMED to arm the plane. It will take off and loiter around the takeoff waypoint.
-
-![](../../assets/gcs/qgc_sim_mission.png)
+## Fly an Autonomous Mission in HITL
 
+You should not be able to use *QGroundControl* to [run missions](../qgc/README.md#planning-missions) and otherwise control the vehicle.
```
hamishwillee: FYI only, decided that we don't need crappy out of date instructions. So here we link to QGC. In future though I'd like a generic "using the simulator instructions" for stuff like flying, disabling sensors, etc. When/if that exists will link to it instead.

en/simulation/hitl.md
```diff
@@ -1,101 +1,133 @@
...
 
-![](../../assets/sim/hil_params.png)
+1. Connect the autopilot directly to *QGroundControl* via USB.
+1. Enable HITL Mode
+   1. Open **Setup > Safety** section.
+   1. Enable HITL mode by selecting **Enabled** from the *HITL Enabled* list:
 
-* Replace the `serialDevice` parameter (`/dev/ttyACM0`) if necessary.
+      ![QGroundControl HITL configuration](../../assets/gcs/qgc_hitl_config.png)
+1. Select Airframe
```
hamishwillee, line 55: @arwagoner  Is there a plan for SITL to also support plane?

en/simulation/hitl.md
arwagoner: iris.sdf is automatically generated when the software is built.

en/simulation/hitl.md
arwagoner: If the user is connecting the autopilot to the computer using a serial link, the serial device will likely be /dev/ttyACM0. If the autopilot is connected via TELEM2, then the device might change to /dev/ttyUSB0. If there are multiple devices connected, the device might show up as /dev/ttyUSB1 or /dev/ttyACM1. An easy way to check on Ubuntu is to plug in the autopilot, open up a terminal, and type `dmesg | grep "tty"`. The correct device will be the last one shown.

en/simulation/hitl.md
dagar: Most of the barrier is the mixer and how it interacts with the simulation. After that it's a matter of getting matching models on each side.

en/simulation/hitl.md
dagar: With HIL and QGC (xplane or flightgear) data link loss ends up being irrelevant because the mavlink connection has to be there for the simulation data.

Otherwise it can be set to whatever's appropriate for the usage.

en/simulation/hitl.md
hamishwillee: Thanks!

en/simulation/hitl.md
dagar: I'm wondering if we should not publicize xplane HIL at the moment. There are several issues, and even when it works the update rate (framerate) is way too slow to run the system realistically. Things in the vehicle running at 40-50 Hz (your game framerate) that typically run at 250 Hz.

en/simulation/hitl.md
hamishwillee: Thanks. I will add a note for that.

en/simulation/hitl.md
hamishwillee: I'll just do it.

hamishwillee: @dagar  If we supported HTIL Plane on Gazebo that would be a no brainer, but at the moment this is the only simulator for HITL and plane - @arwagoner about plans to support HITL/FW?

Is anyone looking at fixing this? I don't want to make an arbitrary decision because there is a bit of work to untangle X-Plane out of this.

en/simulation/hitl.md
hamishwillee: OK, I think I'll just leave this in as it will "cause no harm".

hamishwillee:
@dagar @arwagoner I have merged this as "better than what was there before".

However it would be good to understand when Gazebo HITL is going to be supported for FW. 
@dagar do you think we should add a disclaimer/note about the low quality of the X-Plane simulation?

dagar:
Yes, I suppose that could work. Steer people towards Gazebo SITL primarily.

en/simulation/hitl.md
hamishwillee: As I understand it the mixer translates movement commands from controllers into airframe-specific motor commands. So what I think you are saying is that:
- the mixer for a particular airframe might not match actual behaviour in the simulated vehicle. 
- what you have to do therefore is update the model in Gazebo so that it matches the output of the mixer.

Is that right? 

I don't understand "After that it's a matter of getting matching models on each side." so I intepreted that as update gazebo. However you could be suggesting that PX4 also needs to be modified - perhaps with a HIL mode for the mixer?

PR #423, opened 2018-01-08
Add disclaimer for Xplane use

hamishwillee:


dagar:
Looks good, but can we drop the specifics? Framerate is on estimate based on random machines I've used. PX4 inner loop rate could easily change.

hamishwillee:
Removed specifics. OK now?

dagar (approved):

PR #424, opened 2018-01-08
Update sensor_thermal_calibration.md

priseborough:
The previous description has been obsoleted by changes to the logger.

### DIFF
--- a/en/tutorials/sensor_thermal_calibration.md
+++ b/en/tutorials/sensor_thermal_calibration.md
@@ -91,7 +91,7 @@ This method provides a way to visually check the quality of data and curve fit b
 2. Power up the board and set the TC_A_ENABLE, TC_B_ENABLE and TC_G_ENABLE parameters to 1
 3. Set all CAL_GYR and CAL_ACC parameters to defaults
 4. Set the SYS_LOGGER parameter to 1 to use the new system logger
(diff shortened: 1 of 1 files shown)


hamishwillee:
Thanks @priseborough 

BTW, Should this be in the user guide under [advanced configuration](https://docs.px4.io/en/advanced_config/)? The theory is that content goes in the user guide if it is useful for people that are using *existing* supported airframes and hardware. Content goes into the developer guide if it is useful only for someone who wants to support a new airframe or new hardware (flight controller/sensor/whatever). Which is this?

If it is DevGuide, I'm trying to move things out of "tutorials" section into sections that kind of capture what they are for. So in this case I'd possibly move both this doc and https://dev.px4.io/en/tutorials/tuning_the_ecl_ekf.html into a new folder "Flight Stack". Would that work for you?

priseborough:
The feature has moved from being experimental to part of the released code, so having it in https://docs.px4.io/en/advanced_config/ makes sense.

hamishwillee:
Thanks @priseborough - this sensor calibration doc was moved to user guide in https://github.com/PX4/Devguide/pull/425 and https://github.com/PX4/px4_user_guide/pull/129/files . There are updates planned for us to discuss in https://github.com/PX4/px4_user_guide/pull/130

Shouldn't the https://dev.px4.io/en/tutorials/tuning_the_ecl_ekf.html also move to user guide?

Issue #415, opened 2018-01-03
Gazebo Simulation - needs a section on sensor/GPS noise/delay


hamishwillee:
Bug Page: [Gazebo Simulation · PX4 Developer Guide](https://dev.px4.io/en/simulation/gazebo.html)

Should have a section explaining how to customise the GPS noise/delay as per https://github.com/PX4/Firmware/pull/8555#issuecomment-354979106. More generally, lets document all the important vehicle-specific settings in SDF and any other common environmental settings that people might want to change (wind?).

Probably group under a parent topic along with this heading: https://dev.px4.io/en/simulation/gazebo.html#set-custom-takeoff-location

ndepal:
This would be great.

The only SDF parameter that I'm aware of is the GPS noise:

https://github.com/PX4/sitl_gazebo/blob/1de7366d4e0cb7f784367d8669138d45781c3994/src/gazebo_gps_plugin.cpp#L54-L59

hamishwillee:
@ndepal Can you create some text? The reason I don't want to is that I really have no experience of gazebo. 

Let me outline my questions/concerns I'd want such a section to answer.
1. I understand that the SDF file is created as part of the PX4 build process? Doesn't that mean that it might get over-written every time you build PX4?
2. I did a build against current Firmware. I can find SDF files, but none seem to have gpsNoise in them. 
   - Is it not present by default? If it is, in which sdf files?
   - If it is not present by default that implies that gpsNoise is OFF by default?
   - Assuming we have to add this section in, what does it look like. i.e similar to this from file:
```
    <plugin name='rotors_gazebo_imu_plugin' filename='librotors_gazebo_imu_plugin.so'>
      <robotNamespace></robotNamespace>
      <linkName>plane/imu_link</linkName>
      <imuTopic>/imu</imuTopic>
      <gyroscopeNoiseDensity>0.0003394</gyroscopeNoiseDensity>
      <gyroscopeRandomWalk>3.8785e-05</gyroscopeRandomWalk>
      <gyroscopeBiasCorrelationTime>1000.0</gyroscopeBiasCorrelationTime>
      <gyroscopeTurnOnBiasSigma>0.0087</gyroscopeTurnOnBiasSigma>
      <accelerometerNoiseDensity>0.004</accelerometerNoiseDensity>
      <accelerometerRandomWalk>0.006</accelerometerRandomWalk>
[...]

PR #426, opened 2018-01-09
Add GPS noise information to Gazebo simulation doc

hamishwillee:
Adds a "Simulating GPS Noise" section and groups this and other "general setup" options under a heading "Usage/Configuration Options".

The noise section is constructed from info from @ndepal here: https://github.com/PX4/Devguide/issues/415#issuecomment-356033924

@TSC21 Can you please review. In particular I want to confirm that SDF files aren't overwritten after first being built. Is there anything else you think we should add? 

Note, I am assuming it is not necessary/useful to add information about where the GPS noise values are set and exactly how these map to  HDOP/VDOP/EPH/EPV values. If you think that would be useful I would need to get information from you on this mapping.

### DIFF
--- a/en/simulation/gazebo.md
+++ b/en/simulation/gazebo.md
@@ -113,7 +113,7 @@ make posix_sitl_default gazebo_hippocampus
 
 ## Change World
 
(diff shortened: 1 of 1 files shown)


ndepal:
This is great!

> I am assuming it is not necessary/useful to add information about where the GPS noise values are set and exactly how these map to HDOP/VDOP/EPH/EPV values

From a documentation perspective I agree. However, I believe the estimators use EPH/EPV as measurement noise parameters, which is why I thought it would be valuable to have them correspond to the noise added. @TSC21, what do you think?

hamishwillee:
> I thought it would be valuable to have them correspond to the noise added. @TSC21, what do you think?

FWIW I tend to think that whether these should be documented depends on whether or not we expect/allow developers to want to modify them (and I can certainly imagine a scenario where we want to do so). BUT if we desire to support that case, I'd hope we'd expose the parameters through the SDF too.

Issue #433, opened 2018-01-17
Doc Bug: HITL Simulation · PX4 Developer Guide


jingego:
DESCRIBE PROBLEM WITH DOCS HERE

Bug Page: [HITL Simulation · PX4 Developer Guide](https://dev.px4.io/en/simulation/hitl.html)

The documentation is hard to understand.  
as described in HITL Simulation Environment, the application  "simulator" on hardware will handle the mavlink messages and publish in relevant topics.  In PX4 configuration section, the autopilot board flashed normal Firmware and then selected HITL airframe. For either px4fmu-v2_default or px4fmu-v3_default, the 'modules/simulator' was not built. how could this HITL structure work?
The main problem is that I cannot understand this HITL structure. Could anyone kindly help to descript it in detail?
I have not successfully run this HITL simulation since I cannot see the same thing in step 2 of PX4 configuration. So I simply set the SYS_HIL parameter to enable. I may make some mistakes. I will try it again.

jingego:
![](https://dev.px4.io/assets/simulation/px4_hitl_overview.png)

I further read the code, it should be [mavlink_receiver.cpp](https://github.com/PX4/Firmware/blob/85ba1607575575340a4d4e02f7a2176475255d44/src/modules/mavlink/mavlink_receiver.cpp#L1828-L1965) will handle the HIL message. Note (Line 33) in [ Simulation](https://github.com/PX4/Devguide/blob/645f103e0b6989da44676be848d8966cd5f5840d/en/simulation/README.md#L33) may better to revise for better understanding.

hamishwillee:
@jingego Thanks very much for posting this. I *think* you are right. I also agree the solution is to change the diagram above to say mavlink_reciever.cpp rather than simulator_mavlink.cpp.

@dagar As per this post, my *new* understanding is that the simulator API is implemented in  **simulator_mavlink.cpp** for SITL. For HIL it is implemented in [mavlink_receiver.cpp](https://github.com/PX4/Firmware/blob/85ba1607575575340a4d4e02f7a2176475255d44/src/modules/mavlink/mavlink_receiver.cpp#L1828-L1965). The mavlink receiver code is only executed in HIL mode and the simulator_mavlink isn't present in firmware, so there is never the case where there are two handlers present for the simulation API messages.

Upshot, the HIL diagram needs to be changed to say mavlink_reciever.cpp. Does that make sense to you?

hamishwillee:
[...]

PR #443, opened 2018-01-31
Update HIL docs to show correct sim API handler -  mavlink_receiver.cpp

hamishwillee:
This addresses https://github.com/PX4/Devguide/issues/433#issuecomment-361791273


assets/simulation/px4_hitl_overview.svg
```diff
@@ -885,7 +885,7 @@
          id="tspan2761"
          x="-83.335098"
          y="147.11555"
-         style="font-size:3.88055563px;writing-mode:lr-tb;stroke-width:0.26458335px">simulator_mavlink.cpp</tspan></text>
+         style="font-size:3.88055563px;writing-mode:lr-tb;stroke-width:0.26458335px">mavlink_receiver.cpp</tspan></text>
```
bkueng: I find this a bit confusing: `mavlink_main.cpp` and `mavlink_receiver.cpp` are in different blocks, but it's actually the same module (and there are more source files involved for the communication). Should this not be a single block?

Also I noticed from the diagram and the text `QGroundControl is connected to the physical hardware over USB and acts as a gateway to forward data between the simulator running on a development computer, PX4, and any offboard API.`: this does not apply when jMAVSim is used for HITL: in that case, jMAVSim acts as gateway, connects to UART and forwards MAVLink so that QGC can connect (QGC is optional in that case).

hamishwillee:
@bkueng Thank you for your review. Updated, can you please check again?

> I find this a bit confusing: mavlink_main.cpp and mavlink_receiver.cpp are in different blocks, but it's actually the same module (and there are more source files involved for the communication). Should this not be a single block?

Yes. That is confusing. This was done originally because those are the main files where the messages are handled. Fixed.

> Also I noticed from the diagram and the text QGroundControl is connected to the physical hardware over USB and acts as a gateway to forward data between the simulator running on a development computer, PX4, and any offboard API.: this does not apply when jMAVSim is used for HITL: in that case, jMAVSim acts as gateway, connects to UART and forwards MAVLink so that QGC can connect (QGC is optional in that case).

I did not know that! Thanks - to fix this I've split this into two diagrams and two descriptions. Please check the jMAVSim text/diagram.

en/simulation/hitl.md
```diff
@@ -20,17 +20,34 @@ Airframe | `SYS_AUTOSTART` | X-Plane | Gazebo | jMAVSim
...
-> **Tip** For more information see: [Simulation](../simulation/README.md).
+
+### jMAVSim HITL Environment
+
+*jMAVSim* is connected to the flight controller hardware via USB/UART and acts as gateway to forward MAVLink data to *QGroundControl* or any offboard API (if required).
+
+The diagram below shows the jMAVSim HITL simulation environment:
+* A HITL configuration is selected (via *QGroundControl*) that doesn't start any real sensors.
+* *jMAVSim* is connected to the flight controller via USB.
+* (Optional) *QGroundControl* and any Offboard API connects to jMAVSim via UDP. jMAVSim forwards commands to PX4.
```
bkueng: - Offboard is not supported (no port 14540 is opened from jMAVSim), so the Offboard block should be removed as well from the diagram.
- jMAVSim uses the port 14550 to communicate (or forward MAVLink) with QGC (same as for SITL)

bkueng (commented):
Thanks for the update. I did some more digging in the code and found other things that are incorrect.

Also looking at https://github.com/PX4/sitl_gazebo/pull/146: the setup for gazebo seems to be the same as for jMAVSim: gazebo directly connects to the UART and forwards MAVLink to QGC. So it looks like the exception is x-plane.

hamishwillee:
@bkueng Thanks for digging/clarifications - I got review but didn't actually run/check the HITL stuff. I'll fix this up next week.

en/simulation/hitl.md
hamishwillee: So is offboard supported for Gazebo? It must be for ROS to work, so just making sure!

hamishwillee: @bkueng I was thinking perhaps if the only difference is support for offboard, I'd update the current jmavsim doc like this:
image

... and fix up the current gazebo doc to refer to X-Plane. Does that make sense

bkueng: Yes that looks good (plus the port 14550 instead of 14560). I'm not entirely sure if offboard is supported in gazebo HIL, but most likely it is.

hamishwillee:
Thanks @bkueng. I've updated the image as discussed, and the text too - now jMAVSim/Gazebo share a section and XPlane is the one on its own. Ready to merge?

en/simulation/README.md
```diff
@@ -223,9 +223,7 @@ For more information about using the MAVLink module see [Modules Reference: Comm
 
 ## HITL Simulation Environment
 
-With Hardware-in-the-Loop (HITL) simulation the normal PX firmware is run on real hardware. *QGroundControl* is connected to the physical hardware over USB and acts as a gateway to forward data between the simulator, PX4 and any offboard API.
-
-The HITL Simulation Environment in documented in: [HITL Simulation](../simulation/hitl.md).
+With Hardware-in-the-Loop (HITL) simulation the normal PX firmware is run on real hardware. The HITL Simulation Environment in documented in: [HITL Simulation](../simulation/hitl.md).
```
bkueng: Typo: PX**4**

en/simulation/hitl.md
```diff
@@ -20,17 +20,38 @@ Airframe | `SYS_AUTOSTART` | X-Plane | Gazebo | jMAVSim
 
 ## HITL Simulation Environment {#simulation_environment}
 
-With Hardware-in-the-Loop (HITL) simulation the normal PX firmware is run on real hardware. *QGroundControl* is connected to the physical hardware over USB and acts as a gateway to forward data between the simulator running on a development computer, PX4, and any offboard API.
+With Hardware-in-the-Loop (HITL) simulation the normal PX firmware is run on real hardware. The HITL configuration is slightly different for Gazebo, jMAVSim and X-Plane.
```
bkueng: Same: PX**4**

en/simulation/hitl.md
```diff
@@ -20,17 +20,38 @@ Airframe | `SYS_AUTOSTART` | X-Plane | Gazebo | jMAVSim
...
+
+### JMAVSim/Gazebo HITL Environment
+
+JMAVSim or Gazebo (running on a development computer) are connected to the flight controller hardware via USB/UART. The simulator acts as gateway to share MAVLink data between PX4 and *QGroundControl*. 
+
+> **Tip** Gazebo can additionally share MAVLink data with an offboard API!
+
+The diagram below shows the simulation environment:
+* A HITL configuration is selected (via *QGroundControl*) that doesn't start any real sensors.
+* *jMAVSim* or *Gazebo* are connected to the flight controller via USB.
```
bkueng, line 35: I just remembered that there is another possibility: you can do HIL via Ethernet (or even WiFi, but you need a good one with low latency). I've done that on the Raspberry Pi. This ist the startup configuration, including how to run jMAVSim:
https://github.com/PX4/Firmware/blob/master/posix-configs/rpi/px4_hil.config
fyi @julianoes

bkueng (commented):
I found a typo, and we could mention that HIL works on the Rapsberry Pi too and point to the configuration. Other than that it's good to go.

en/simulation/hitl.md
hamishwillee: Thanks, I have fixed typos and merged as is. Created separate issue for Ethernet option: https://github.com/PX4/Devguide/issues/466

PR #444, opened 2018-01-31
Add Nuttx menuconfig shortcut info

hamishwillee:
Adds info about the shortcuts. It links to more info about menuconfig but does not explain why you might use this - as discussed with @dagar this is quite a specialised use case, where readers that need the information will probably already understand how it should be used (so we're just providing a pointer).

@dagar Note that the location of this section in the porting guide isn't ideal (it is under "Porting Guide > *Architecture* > NuttX > "). This is the only NuttX topic in current structure, so the best place "for now". When we have a more complete porting guide it probably makes sense to have a structure that divides first by board - we will see.


dagar:
Thanks!

PR #450, opened 2018-02-05
Update building_px4.md

IeiuniumLux:
The Native Build section needs to be updated with the new location of the px4 file. In newer release of the flight stack, it is located just under ./build/posix_rpi_native/px4


hamishwillee:
Thank you very much for this. Merging.

PR #457, opened 2018-02-09
Added further description to flight modes

barzanisar:
Added further description to flight modes for ease of understanding to new users.


en/concept/flight_modes.md
```diff
@@ -14,23 +14,23 @@ The sections below provide an overview of the modes, followed by a [flight mode
 "Manual" modes are those where the user has direct control over the vehicle via the RC control (or joystick). Vehicle movement always follows stick movement, but the level/type of response changes depending on the mode. For example, experienced fliers can use modes that provide direct passthrough of stick positions to actuators, while beginners will often choose modes that are less responsive to sudden stick-position changes.
 
 * **Fixed wing aircraft/ rovers / boats:** 
-  * **MANUAL:** The pilot's control inputs are passed directly to the output mixer.
-  * **STABILIZED:** The pilot's inputs are passed as roll and pitch *angle* commands and a manual yaw command.
+  * **MANUAL:** The pilot's control inputs or RC commands are passed directly to the output mixer.
```
hamishwillee: Not sure about this - the control inputs are RC commands. What is it you're trying to capture here?

en/concept/flight_modes.md
```diff
@@ -14,23 +14,23 @@ The sections below provide an overview of the modes, followed by a [flight mode
...
+  * **STABILIZED:** The pilot's inputs are passed as roll and pitch *angle* commands and a manual yaw command. If the RC roll and pitch sticks are centered, the autopilot regulates the roll and pitch angles to zero, hence stabilizing (leveling-out) the attitude against any wind disturbances. However, in this mode the position of the aircraft is not controlled by the autopilot, hence the position can drift due to wind.
```
hamishwillee: @bkueng Is this correct?

@barzanisar Assuming this is correct, it is an excellent update.  Essentially you're saying that this keeps the vehicle upright if roll/pitch sticks are centred - that wasn't clear before. 

What does the first sentence actually imply/mean? ("The pilot's inputs are passed as roll and pitch *angle* commands and a manual yaw command.") 
- ie how does this improve stabilisation?

en/concept/flight_modes.md
```diff
@@ -14,23 +14,23 @@ The sections below provide an overview of the modes, followed by a [flight mode
...
 * **Multirotors:**
-  * **ACRO:** The pilot's inputs are passed as roll, pitch, and yaw *rate* commands to the autopilot.  This allows the multirotor to become completely inverted.  Throttle is passed directly to the output mixer
-  * **RATTITUDE** The pilot's inputs are passed as roll, pitch, and yaw *rate* commands to the autopilot if they are greater than the mode's threshold.  If not the inputs are passed as roll and pitch  *angle* commands and a yaw *rate* command.  Throttle is passed directly to the output mixer.
-  * **STABILIZED** The pilot's inputs are passed as roll and pitch *angle* commands and a yaw *rate* command.  Throttle is passed directly to the output mixer.
+  * **ACRO:** The pilot's inputs are passed as roll, pitch, and yaw *rate* commands to the autopilot. The autopilot controls the angular rates, but not the attitude. Hence, if the RC sticks are centered the vehicle will not level-out. This allows the multirotor to become completely inverted.  Throttle is passed directly to the output mixer.
```
hamishwillee: So as I understand it, the autopilot passes *rate* commands for roll, pitch, yaw. Does this mean that the vehicle responds to speed of movement of the RC controls, not their position. If the RC is centred there will be no movement of sticks, hence no rate change,, so the vehicle would just keep its previous roll/pitch/yaw?

Personally I didn't find this new sentence helpful "The autopilot controls the angular rates, but not the attitude. Hence, if the RC sticks are centered the vehicle will not level-out. ". I think what you are trying to do is compare this to the "STABILIZED" mode? Essentially saying in stabilised mode the system will level you out, but in ACRO you're on your own. Is that correct?

en/concept/flight_modes.md
```diff
@@ -14,23 +14,23 @@ The sections below provide an overview of the modes, followed by a [flight mode
...
+  * **RATTITUDE** The pilot's inputs are passed as roll, pitch, and yaw *rate* commands to the autopilot if they are greater than the mode's threshold, i.e. if the RC sticks are a certain distance away from the center position.  If not the inputs are passed as roll and pitch  *angle* commands and a yaw *rate* command.  Throttle is passed directly to the output mixer. In short, the autopilot acts as an angular rate controller when the RC sticks are away from center (like in the ACRO mode), whereas when the RC sticks are centered, the autopilot acts as an attitude controller (like in the Stabilized mode).
```
hamishwillee: This is good :-) As in previous comments I just need to better understand how passing angle vs passing rate commands actually affects flight.

en/concept/flight_modes.md
```diff
@@ -14,23 +14,23 @@ The sections below provide an overview of the modes, followed by a [flight mode
...
+  * **STABILIZED** The pilot's inputs are passed as roll and pitch *angle* commands and a yaw *rate* command. Throttle is passed directly to the output mixer. The autopilot controls the attitude, meaning it regulates the roll and pitch angles to zero when the RC sticks are centered, consequently leveling-out the attitude. However, in this mode the position of the vehicle is not controlled by the autopilot, hence the position can drift due to wind.
```
hamishwillee: Again it would be good (for me) to understand the actual impact/meaning of each part of this (ie angle vs rate commands):
- roll input is passed as angle* command
- pitch input is passed as *angle* command
- ywa input is passed as *rate* command.
- Throttle is passed directly to the output mixer.

en/concept/flight_modes.md
```diff
@@ -14,23 +14,23 @@ The sections below provide an overview of the modes, followed by a [flight mode
...
         
 ### Assisted flight modes
 
 "Assisted" modes are also user controlled but offer some level of "automatic" assistance - for example, automatically holding position/direction, against wind. Assisted modes often make it much easier to gain or restore controlled flight.
 
 * **ALTCTL** (Altitude Control)
-  * **Fixed wing aircraft:** When the roll, pitch and yaw inputs (RPY) are all centered (less than some specified deadband range) the aircraft will return to straight and level flight and keep its current altitude. It will drift with the wind.
-  * **Multirotors:** Roll, pitch and yaw inputs are as in MANUAL mode. Throttle inputs indicate climb or sink at a predetermined maximum rate. Throttle has large deadzone.
+  * **Fixed wing aircraft:** When the roll, pitch and yaw inputs (RPY) or sticks in RC are all centered (less than some specified deadband range) the aircraft will return to straight and level flight and keep its current altitude. It's x,y position will drift with the wind.
```
hamishwillee: We should see if https://docs.px4.io/en/flight_modes/altitude.html also requires update as a result of this change. 

" It's x,y position" > remove apostrophe in "It's"

en/concept/flight_modes.md
```diff
@@ -14,23 +14,23 @@ The sections below provide an overview of the modes, followed by a [flight mode
...
 * **POSCTL** (Position Control)
-  * **Fixed wing aircraft:** Neutral inputs give level, flight and it will crab against the wind if needed to maintain a straight line.
-  * **Multirotors** Roll controls left-right speed, pitch controls front-back speed over ground. When roll and pitch are all centered (inside deadzone) the multirotor will hold position. Yaw controls yaw rate as in MANUAL mode. Throttle controls climb/descent rate as in ALTCTL mode.
+  * **Fixed wing aircraft:** Neutral inputs (centered RC sticks) give level flight and it will crab against the wind if needed to maintain a straight line.
+  * **Multirotors** Roll controls left-right speed, pitch controls front-back speed over ground. When roll and pitch sticks are all centered (inside deadzone) the multirotor will hold x, y position. Yaw controls yaw rate as in MANUAL mode. Throttle controls climb/descent rate as in ALTCTL mode. This means that the x, y, z position of the vehicle is held steady by the autopilot against any wind disturbances, when the roll, pitch and throttle sticks are centered.
```
hamishwillee: "the multirotor will hold x, y position" - I don't think you need "x,y" here as the position is held steady in all dimensions.

en/concept/flight_modes.md
```diff
@@ -41,7 +41,7 @@ The sections below provide an overview of the modes, followed by a [flight mode
   * **Multirotors:**  The multirotor hovers / loiters at the current position and altitude.
 * **AUTO_RTL** (Return to Land)
   * **Fixed wing aircraft:** The aircraft returns to the home position and loiters in a circle above the home position. 
-  * **Multirotors:** The multirotor returns in a straight line on the current altitude (if higher than the home position + loiter altitude) or on the loiter altitude (if higher than the current altitude), then lands automatically.
+  * **Multirotors:** The multirotor returns in a straight line on the current altitude (if the current altitude is higher than the home position + loiter altitude) or on the loiter altitude (if the loiter altitude is higher than the current altitude), then lands automatically.
```
hamishwillee: Actual behaviour is a bit more complicated than this, and is affected by parameters (e.g. a plane might land). The "loiter altitude" is not the correct parameter - should be RTL return height. We should tidy this a little and link to https://docs.px4.io/en/flight_modes/rtl.html.  Actually we should link to all the dedicated pages in user guide for each of these.

hamishwillee:
Excellent. Comments/request for clarifications added.

en/concept/flight_modes.md
barzanisar: When I was first reading the dev guide without reading the user guide, I got confused thinking "pilot" meant the "autopilot" and the "control inputs" were the commands computed by the controllers in autopilot. Maybe this confusion arose because I have always learned in my control systems lectures that "control inputs" is the term commonly used for commands computed by the controller for correction of errors, whereas in the manual mode (correct me if I am wrong) the user just sends the RC commands as "references"  and an open loop control is involved. 
Maybe this confusion is personal and due to not reading the user guide before. 
But if you think others like me can also get confused then it will be nice to add "(RC commands)" in brackets next to the "pilot's control inputs" the first time it is used for further clarification of where the "control inputs" are actually coming from. :)

en/concept/flight_modes.md
barzanisar: In the first line, I don't understand what "**manual** yaw command" means and how it is different from the "roll and pitch commands". Can anyone please explain this? Does this mean that the "roll and pitch commands" from RC are sent as references to the autopilot's attitude controller which tries to track these references, whereas the the "yaw command" from RC is sent directly as output to the rudder to control yaw manually? 

If this is true, the first line can be further elaborated as:
"The pilot's inputs are passed as roll and pitch *angle* commands to the autopilot and a yaw command is passed directly to the output mixer to control rudder manually."

en/concept/flight_modes.md
hamishwillee: If you can make this assumption someone else might too. How about "The user's input from the RC transmitter ..."?

en/concept/flight_modes.md
hamishwillee: Lets see what @bkueng says :-)

en/concept/flight_modes.md
barzanisar: RC always sends the commands corresponding to the **position** of its sticks. I think of RC sticks as potentiometers, whose voltage output range (e.g. 0 - 5V) is first mapped to "-100 to 100" range within the RC and then this range can be mapped to any numerical range (e.g. -180 degrees to +180 degrees for stabilised mode, and -15 rad/sec to +15 rad/sec for acro mode) we want within the autopilot. So when the autopilot receives a certain value in the range -100 to 100 coming from RC, it maps this value to a certain number, which is then directed to the attitude controller or rate controller (depending on the flight mode) as their reference values to track. This is what I think roughly happens. I will confirm after I complete reading the dev guide. Also the numbers above I made up for explaining.

So in Stabilised mode, the "position" of RC sticks are translated into roll, pitch, yaw angle references to be sent to the autopilot's "attitude controller". Whereas in the Acro mode, the "position" of RC sticks determines roll, pitch and yaw rate references to be sent to the autopilot's "rate controller". This is what the PX4 team also confirmed. So if the RC sticks are centered, it is passed as "zero" roll, pitch, yaw _rate_ reference to the autopilot. Which means we are telling the autopilot to not change the roll, pitch and yaw angles anymore, consequently (depending on the speed of the system response) trying to keep the current roll, pitch, yaw angles as they are. Whereas in Stabilised mode, centering the sticks position means we are sending zero roll,pitch, yaw _angle_ references so the controller tries to keep the angles zero.

Also, in the Acro mode, I personally think it is important to make this distinction between the stabilised and acro mode because one uses "attitude controller, meaning roll/pitch angles controller" vs the other which triggers "attitude _rate_ controller, meaning roll/pitch rate controller".

en/concept/flight_modes.md
barzanisar: This is what I understand: 
RC roll and pitch sticks position is translated into roll and pitch angle references that are sent to the attitude controller of the autopilot. 
RC yaw stick position is translated into yaw rate reference sent to the yaw rate controller of the autopilot.
Throttle is translated into motor speeds directly without any closed loop control.

en/concept/flight_modes.md
bkueng: Ok, so I had to ask for that as well :)

> Does this mean that the "roll and pitch commands" from RC are sent as references to the autopilot's attitude controller which tries to track these references

Yes this is correct.

> the "yaw command" from RC is sent directly as output to the rudder to control yaw manually?

That is correct too. But there is one thing missing: the vehicle does a [coordinated turn](https://en.wikipedia.org/wiki/Coordinated_flight), to have zero sideslip (the acceleration in y-direction (sidewards) is zero). This means during a turn, the rudder is used to control the sideslip. If the pilot gives non-zero yaw input, the input will just be added to what the vehicle commands for the coordinated turn.

en/concept/flight_modes.md
barzanisar: Agreed. Do I have to install Gitbook editor to insert the link?

en/concept/flight_modes.md
hamishwillee: @barzanisar Even better, how about "The raw inputs from the RC receiver"

hamishwillee:
Hi @barzanisar I might not have time to respond to stuff today. HOwever, can you please keep each logical change in separate PR. So all changes for flight mode can go together, but changes related to system console do not belong here. 

Why? 
- Too much to review
- Different reviewers have different expertise
- Waiting on any part of the PR to be approved/complete stops all of it being merged.

en/concept/flight_modes.md
hamishwillee: No, you can add links manually. An external link in markdown has the format 
`[Text to display](Link URL)`.

en/concept/flight_modes.md
hamishwillee: Thank you for that description. So what I think you're saying is that with the yaw rate commands (say) we're controlling the rate of angular rotation of yaw while with the angle commands we're setting the target yaw angle. Angle commands are therefore better for stabilisation because the vehicle tracks the position of the sticks, while rate commands affect rate of change, and are hence more responsive. Correct?

en/concept/architecture.md
```diff
@@ -29,7 +29,7 @@ again. -->
 The source code is split into self-contained modules/programs (shown in `monospace` in the
 diagram). Usually a building block corresponds to exactly one module. 
 
-> **Tip** At runtime, you can inspect which modules are executed with the `top` command, 
+> **Tip** At runtime, you can inspect which modules are executed with the `top` command in nuttx shell,
```
hamishwillee: Should it be NuttX shell or nsh shell? Either way, not nuttx

bkueng: I'd say just `shell`. While `top` is NuttX-specific, the rest can be used in the SITL shell (`pxh>`) as well.

en/concept/flight_modes.md
barzanisar: Yes, that is correct.
One correction in my comment above:
In Stabilised mode,  the "position" of RC sticks are translated into roll, pitch angles and yaw **rate** references to be sent to the autopilot's "attitude controller" and "yaw rate controller". Centering the sticks position means we are sending zero roll and pitch angles and zero yaw rate references so the controller tries to keep the roll, pitch angles and yaw rate zero.

en/concept/flight_modes.md
barzanisar: Since everywhere in the document it is written "pilot's inputs" so rather than changing everywhere, it will be easier to mention in brackets the first time pilot's inputs are used with (raw commands from the RC). What do you think?

en/concept/flight_modes.md
hamishwillee: @barzanisar Doc on coordinated turn should be added.

en/concept/flight_modes.md
hamishwillee: Awesome. OK, so please make sure that the stabilised doc reflects how it works :-)

Re this ACRO .. do you think this is an improvement:
* **ACRO:** The pilot's inputs control the roll, pitch, and yaw *angular rotation rate*. Throttle is passed directly to the output mixer. 
  The multirotor can become completely inverted. If the RC sticks are centered the vehicle will try to keep the angular rates at zero (maintain the vehicle's current attitude).

en/concept/flight_modes.md
```diff
@@ -41,7 +41,7 @@ The sections below provide an overview of the modes, followed by a [flight mode
   * **Multirotors:**  The multirotor hovers / loiters at the current position and altitude.
 * **AUTO_RTL** (Return to Land)
   * **Fixed wing aircraft:** The aircraft returns to the home position and loiters in a circle above the home position. 
-  * **Multirotors:** The multirotor returns in a straight line on the current altitude (if the current altitude is higher than the home position + loiter altitude) or on the loiter altitude (if the loiter altitude is higher than the current altitude), then lands automatically.
+  * **Multirotors:** The multirotor returns in a straight line on the current altitude (if the current altitude is higher than the home position + <span id="RTL_RETURN_ALT"></span>[RTL_RETURN_ALT](../advanced_config/parameter_reference.md#RTL_RETURN_ALT)) or on the <span id="RTL_RETURN_ALT"></span>[RTL_RETURN_ALT](../advanced_config/parameter_reference.md#RTL_RETURN_ALT) (if the <span id="RTL_RETURN_ALT"></span>[RTL_RETURN_ALT](../advanced_config/parameter_reference.md#RTL_RETURN_ALT) is higher than the current altitude), then lands automatically.
```
hamishwillee: 1. You don't need the "span"s here - you only need them if you expect to link to this point. 
2. FW and MC have the same behaviour with respect to RTL_RETURN_ALT. The only real difference is that the default behaviour on return for FW is to circle rather than land. 
3. Good to link to the user guide.

How about for this section:
 * **[AUTO_RTL](https://docs.px4.io/en/flight_modes/rtl.html)** (Return): The vehicle ascends to altitude `RTL_RETURN_ALT` (or maintains altitude if already at a greater height) and returns to the home position.  By default a multicopter will land at the home position, while a fixed-wing vehicle will loiter in a circle above it.

hamishwillee:
@barzanisar I've added a few more comments. See what you think. After that I might merge this and do my own pass across it for consistency (and get your feedback!)

en/concept/flight_modes.md
barzanisar: @bkueng in https://docs.px4.io/en/flight_modes/ I found that for fixed wing aircrafts, the state of ACRO mode decides whether to switch to Stabilised mode or Manual mode. And from the Dev guide's flight mode evaluation diagram I understood the following, so correct me if I am wrong:

For fixed wing aircrafts:
If stabilised mode is switched on--> vehicle goes to stabilised mode.
If manual mode is switched on--> vehicle goes to manual mode.
If Acro **or** Rattitude mode is switched on--> vehicle goes to stabilised mode.
If Acro mode **and** Rattitude modes are switched off--> vehicle goes to manual mode.

en/concept/flight_modes.md
barzanisar: When I said: "Which means we are telling the autopilot to not change the roll, pitch and yaw angles anymore, **consequently (depending on the speed of the system response) trying to keep the current roll, pitch, yaw angles as they are.**" 
I meant that we want to zero the change the roll, pitch, yaw rate. So when we center the sticks instantaneously, the autopilot will take some some to decrease the angular rates to zero. But when the autopilot does make the rates zero, the roll, pitch, yaw angles at this point will be different from the angles at the moment we centered the sticks. So autopilot just cares about zeroing the rates not exactly "maintaining the current roll,pitch, yaw".  

Ignore my comment "**consequently (depending on the speed of the system response) trying to keep the current roll, pitch, yaw angles as they are**", it is causing confusion I guess.

en/concept/flight_modes.md
barzanisar: @bkueng do you mean that in stabilised mode the aircraft always does a coordinated turn for nonzero roll/pitch/yaw inputs?

bkueng: > do you mean that in stabilised mode the aircraft always does a coordinated turn for nonzero roll/pitch/yaw inputs?

Yes, that's how I understand it.

> @bkueng in https://docs.px4.io/en/flight_modes/ I found that for fixed wing aircrafts, the state of ACRO mode decides whether to switch to Stabilised mode or Manual mode. And from the Dev guide's flight mode evaluation diagram I understood the following, so correct me if I am wrong:

This is incorrect (outdated) and needs to be updated. There is now a separate acro mode for fixed-wing: the user controls the vehicle's body rates and the controller stabilizes the rates (all 3 axis: roll, pitch & yaw).
(whereas stabilize will level out and do the coordinated turns, so stabilize the attitude angles)

hamishwillee:
It looks like all comments have been addressed, so I have merged this. Thanks @barzanisar 

Just one remaining point, for another PR if needed.

>> @bkueng in https://docs.px4.io/en/flight_modes/ I found that for fixed wing aircrafts, the state of ACRO mode decides whether to switch to Stabilised mode or Manual mode. And from the Dev guide's flight mode evaluation diagram I understood the following, so correct me if I am wrong:

> This is incorrect (outdated) and needs to be updated. There is now a separate acro mode for fixed-wing: the user controls the vehicle's body rates and the controller stabilizes the rates (all 3 axis: roll, pitch & yaw).
(whereas stabilize will level out and do the coordinated turns, so stabilize the attitude angles)

@bkueng @barzanisar  Can you confirm what needs to be done here? 

1. I think most of the  https://dev.px4.io/en/concept/flight_modes.html section on FW is correct but still needs to addition of a FW ACRO section - e.g.  
```
* **Fixed wing aircraft/ rovers / boats:** 
  * **MANUAL:** T...
  * **STABILIZED:** ...
> **Acro**:  the pilot controls the vehicle's body rates and the controller stabilizes the rates (all 3 axis: roll, pitch & yaw).
```
- Is that correct?
- @barzanisar If so, could you add a PR and make the docs consistent.

2. Same text should migrate to [PX4 User Guide > Flight Modes](https://docs.px4.io/en/flight_modes/) in section on Manual Modes fixed wing. My understanding of the above is that this still incorrectly documents a relationship between stabilised and ACRO modes. Correct?
3. Does the diagram need to change: https://dev.px4.io/en/concept/flight_modes.html#flight-mode-evaluation-diagram - if so how?

bkueng:
> Is that correct?

Yes.

> Same text should migrate to PX4 User Guide > Flight Modes in section on Manual Modes fixed wing. My understanding of the above is that this still incorrectly documents a relationship between stabilised and ACRO modes. Correct?

Correct. In addition the stabilized should mention the coordinated turns.

> Does the diagram need to change: https://dev.px4.io/en/concept/flight_modes.html#flight-mode-evaluation-diagram - if so how?

This is for the (old) hierarchical mode switching. It's still relevant, but I think the acro needs to be updated. It should also mention that by default (on the QGC side, which might be documented already), a single mode-switch (usually with 3 states) is used. @RomanBapst can help with that as well.

hamishwillee:
Thanks @bkueng 

So, @barzanisar If so, could you add a PR and make the docs consistent? Specifically:

1. https://dev.px4.io/en/concept/flight_modes.html section needs to addition of a FW ACRO section - e.g.
   > **Acro**:  the pilot controls the vehicle's body rates and the controller stabilizes the rates (all 3 axis: roll, pitch & yaw).
   - Feel free to improve on that text!
1. In the Stablised section add info about coordinated turns that @bkueng added in one of the threads above 
1. And another PR to update PX4 User Guide > Flight Modes in section on Manual Modes fixed wing. No need to fix everything else - just this bit that is actually wrong (to match the changes made here). 

If not, let me know and I'll try find some time next week.

PR #461, opened 2018-02-13
Update multi-vehicle-simulation.md

lamping7:
Changes were introduced https://github.com/PX4/Firmware/pull/8831 and https://github.com/PX4/Firmware/pull/8839. I added additional details for adding more vehicles as well. Let me know if this works for you.

### DIFF
--- a/en/simulation/multi-vehicle-simulation.md
+++ b/en/simulation/multi-vehicle-simulation.md
@@ -1,8 +1,10 @@
 # Multi-Vehicle Simulation
 
-This tutorial explains how to simulate multi UAV in Gazebo. This is a software in the loop \(SITL\) simulation.
+This tutorial explains how to simulate multiple UAV vehicles using Gazebo and SITL.
 
-**Note**: This is tested only in Linux.
+It demonstrates an example setup that opens the Gazebo client GUI showing two Iris vehicles in an empty world. You can then control the vehicles with *QGroundControl* and MAVROS in a similar way to how you would manage a single vehicle. 
+
+> **Note** This is tested only in Linux.
 
 ## Required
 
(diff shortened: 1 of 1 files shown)


en/simulation/multi-vehicle-simulation.md
```diff
@@ -39,88 +38,72 @@ To test an example setup, follow the below steps,
...
+* **Gazebo model**: This is defined as `xacro` file in `Firmware/Tools/sitl_gazebo/models/rotors_description/urdf/<model>_base.xacro` see [here](https://github.com/PX4/sitl_gazebo/tree/02060a86652b736ca7dd945a524a8bf84eaf5a05/models/rotors_description/urdf). Currently, the model `xacro` file is assumed to end with base.xacro. This model should have an argument called  `mavlink_udp_port` which defines the UDP port on which gazebo will communicate with PX4 node. The model's `xacro` file will be used to generate an `urdf` model that contains UDP port that you select. To define the UDP port, set the `mavlink_udp_port` in the launch file for each vehicle, see [here](https://github.com/PX4/Firmware/blob/4d0964385b84dc91189f377aafb039d10850e5d6/launch/multi_uav_mavros_sitl.launch#L37) as an example.  
   **NOTE: If you are using the same vehicle model, you don't need  a separate **`xacro`** file for each vehicle. The same **`xacro`** file is adequate.
```
hamishwillee: Can you change this from:
```
**NOTE: If you are using the same vehicle model, you don't need  a separate **`xacro`** file for each vehicle. The same **`xacro`** file is adequate.    
```
TO
```
> **Note** If you are using the same vehicle model, you can use the same **xacro** file for each vehicle.
``` 
where the `>` is at the beginning of the line.

en/simulation/multi-vehicle-simulation.md
```diff
@@ -39,88 +38,72 @@ To test an example setup, follow the below steps,
...
+  ```
+      <!-- MAVROS -->
+      <include file="$(find mavros)/launch/px4.launch">
+          <arg name="fcu_url" value="$(arg fcu_url)"/>
+          <arg name="gcs_url" value=""/>
+          <arg name="tgt_system" value="$(arg ID)"/>
+          <arg name="tgt_component" value="1"/>
+      </include>
+  ```
+  * Note: The complete block for each vehicle is enclosed in a set of `<group>` tags to seperate the ROS namespaces of the vehicles.
```
hamishwillee: I'll fix up this note as a post-process. Hard to get right if you're not rendering this using gitbook toolchain.

en/simulation/multi-vehicle-simulation.md
```diff
@@ -39,88 +38,72 @@ To test an example setup, follow the below steps,
...
+  * selects ports for MAVROS communication by modifying both port numbers in the `fcu_url` arg
+* create a startup file, and change the file as follows:
+  * make a copy of an existing iris rcS startup file (`iris_1` or `iris_2`) and rename it `iris_3`
+  * `MAV_SYS_ID` value to `3`
+  * `SITL_UDP_PRT` value to match that of the `mavlink_udp_port` launch file arg
+  * the first `mavlink start` port and the `mavlink stream` port values to the same values, which is to be used for QGC communication
+  * the second `mavlink start` ports need to match those used in the launch file `fcu_url` arg (Note: Be aware of which port is src and dst for the different endpoints.)
+
+Additional resources:
+* See [Simulation](../simulation/README.md) for a discription of the UDP port configuraion.
```
hamishwillee: configuraion -> configuration

en/simulation/multi-vehicle-simulation.md
```diff
@@ -39,88 +38,72 @@ To test an example setup, follow the below steps,
...
+* **PX4 node**: This is the SITL PX4 app. It communicates with the simulator, Gazebo, through the same UDP port defined in the Gazebo vehicle model, i.e. `mavlink_udp_port`. To set the UDP port on the PX4 SITL app side, you need to set the `SITL_UDP_PRT` parameter in the startup file to match the `mavlink_udp_port` discussed previously, see [here](https://github.com/PX4/Firmware/blob/4d0964385b84dc91189f377aafb039d10850e5d6/posix-configs/SITL/init/ekf2/iris_2#L46). The path of the startup file in the launch file is generated based on the `vehicle` and `ID` arguments, see [here](https://github.com/PX4/Firmware/blob/4d0964385b84dc91189f377aafb039d10850e5d6/launch/multi_uav_mavros_sitl.launch#L36). The `MAV_SYS_ID` for each vehicle in the startup file, see [here](https://github.com/PX4/Firmware/blob/4d0964385b84dc91189f377aafb039d10850e5d6/posix-configs/SITL/init/ekf2/iris_2#L4), should match the `ID` for that vehicle in the launch file [here](https://github.com/PX4/Firmware/blob/4d0964385b84dc91189f377aafb039d10850e5d6/launch/multi_uav_mavros_sitl.launch#L25). This will help make sure you keep the configurations consistant between the launch file and the startup file.
```
hamishwillee: Links to specific lines are discouraged as things can move around - I see that you've linked to a specific blobs in all the cases so they shouldn't, but is it likely that linking to what will become an older version will cause problems in future?

In this case I would add code fragments showing the lines of interest - you can have the link as well, but perhaps to master (knowing full well that the line might move, but at least users can see the code fragment).

en/simulation/multi-vehicle-simulation.md
```diff
@@ -19,14 +19,13 @@ To test an example setup, follow the below steps,
   ```
   cd Firmware_clone
   git submodule update --init --recursive
-  make posix_sitl_default gazebo
+  make posix_sitl_default sitl_gazebo
```
hamishwillee: Does this (using `sitl_gazebo`) just build the code without running it? (Just interested, normally I only see `make posix_sitl_default gazebo_"target"`

en/simulation/multi-vehicle-simulation.md
```diff
@@ -19,14 +19,13 @@ To test an example setup, follow the below steps,
...
   ```
 * source your environment,
```
hamishwillee: These bullets should be numbered items (all items can be `1. ` in markdown) if the steps must be followed in order. I would also sentence capitalise them.

en/simulation/multi-vehicle-simulation.md
```diff
@@ -39,88 +38,72 @@ To test an example setup, follow the below steps,
 
 For each simulated vehicle, the following is required,
 
-* **Gazebo model**: This is defined as `xacro` file in `Firmware/Tools/sitl_gazebo/models/rotors_description/urdf/<model>_base.xacro` see [here](https://github.com/PX4/sitl_gazebo/tree/02060a86652b736ca7dd945a524a8bf84eaf5a05/models/rotors_description/urdf). Currently, the model `xacro` file is assumed to end with base.xacro. This model should have an argument called  `mavlink_udp_port` which defines the UDP port on which gazebo will communicate with PX4 node. The model's `xacro` file will be used to generate an `sdf` model that contains UDP port that you select. To define the UDP port, set the `mavlink_udp_port` in the launch file, see [here](https://github.com/PX4/Firmware/blob/master/launch/multi_uav_mavros_sitl.launch#L48).
```
hamishwillee: mark up base.xacro in "assumed to end with base.xacro."

en/simulation/multi-vehicle-simulation.md
```diff
@@ -39,88 +38,72 @@ To test an example setup, follow the below steps,
...
+* **Gazebo model**: This is defined as `xacro` file in `Firmware/Tools/sitl_gazebo/models/rotors_description/urdf/<model>_base.xacro` see [here](https://github.com/PX4/sitl_gazebo/tree/02060a86652b736ca7dd945a524a8bf84eaf5a05/models/rotors_description/urdf). Currently, the model `xacro` file is assumed to end with base.xacro. This model should have an argument called  `mavlink_udp_port` which defines the UDP port on which gazebo will communicate with PX4 node. The model's `xacro` file will be used to generate an `urdf` model that contains UDP port that you select. To define the UDP port, set the `mavlink_udp_port` in the launch file for each vehicle, see [here](https://github.com/PX4/Firmware/blob/4d0964385b84dc91189f377aafb039d10850e5d6/launch/multi_uav_mavros_sitl.launch#L37) as an example.
```
hamishwillee: - I would actually copy/paste a small fragment of the file as an example at the end. For the link, I'd perhaps link to master.
- "Currently, the model `xacro` file is assumed to end with base.xacro. " What is it that makes this assumption? Is it the launch file script? HOw does the launch script know to use this particular model? What do you do if you want to add a different model or different models for each vehicle?
- Assuming we use same file for all models, does that mean gazebo uses same port for all models to communicate with SITL? Presumably you could use different models if you wanted? In this case, are there any special rules?

en/simulation/multi-vehicle-simulation.md
```diff
@@ -39,88 +38,72 @@ To test an example setup, follow the below steps,
...
+  * make a copy of an existing iris rcS startup file (`iris_1` or `iris_2`) and rename it `iris_3`
+  * `MAV_SYS_ID` value to `3`
+  * `SITL_UDP_PRT` value to match that of the `mavlink_udp_port` launch file arg
+  * the first `mavlink start` port and the `mavlink stream` port values to the same values, which is to be used for QGC communication
+  * the second `mavlink start` ports need to match those used in the launch file `fcu_url` arg (Note: Be aware of which port is src and dst for the different endpoints.)
+
+Additional resources:
+* See [Simulation](../simulation/README.md) for a discription of the UDP port configuraion.
+* See [URDF in Gazebo](http://gazebosim.org/tutorials/?tut=ros_urdf) for more information about spawning the model with xacro.
+* See [RotorS](https://github.com/ethz-asl/rotors_simulator/tree/master/rotors_description/urdf) for more xacro models.
```
hamishwillee: Any reason why the S in RotorS is capitalised

en/simulation/multi-vehicle-simulation.md
hamishwillee: What if we want to launch a different model - e.g. tailsitter?

hamishwillee:
This looks pretty clear - though I haven't tried it yet. Generally speaking I think it would be good to copy the relevant things you expect people to change as code fragments and link to the version in master rather than some "blob" somewhere. 

Most of my concerns with this predate your changes. Specifically,

1. This could do with a better introduction that explains what the example actually does - ie it launches two iris UAVs in Gazebo using ROS that you can then (do what with?). It then explains how the multi-simulation works and how you can further extend to support a third UAV.
  - we also state that this is only tested on Linux. That is fine. Should we say with SITL on Linux?
1. Similarly in the build and run section when you've run the launch file, what happens? Would be good to state what the user sees after this point and what they can do (as simple as the line "You should see Gazebo start and ... You can then do X, Y, Z"
1. What if I want to launch the in Gazebo without having to use ROS? Perhaps out of scope?

In other words, this covers a specific case, but would be good to think about how could be extended to more general case.

I can probably fix up rendering and linking as a post process, but would appreciate your thoughts ...

@TSC21 Can you please give this a technical review?

lamping7:
A bunch of those suggestions refer to what was existing. However, I'll give this time to respond to each of them a little later. I was only trying to update this regarding the newly introduced changes (the old lines didn't sync up to the intended lines anymore) and add a little more info based on questions I've seen regarding this topic, but I now see that you wanted a more thorough review.

en/simulation/multi-vehicle-simulation.md
lamping7: If the files change down the road where the parameters shown in those lines don't exist I'd assume that the user could see that this is out of date and open an issue. If the file changes down the line, and the parameter still exists, there is no harm is using the specific version.

As it was (using master), it was linking to lines that didn't deal with the comment, which is even less helpful in a file that is full of a list of configuration parameters.

I'm not sure if using code fragments would make this more clear, or further confuse all of the parameters and how they interact. The benefit of the way it is now is that you can read it like a paragraph, and then look at the references or have the files open on a second monitor to follow along.

en/simulation/multi-vehicle-simulation.md
lamping7: `make posix_sitl_default gazebo` makes the `sitl_gazebo` submodule, but it also launches the gazebo viewer. We don't want to open the gui yet. That is what the launch file is for.

lamping7: If you want to use a different vehicle, you need a xacro that represents the model and you need to specify the `vehicle` arg in the launch to reflect this.

I guess this is confusing because in this case, when using ROS, we don't want make to end up starting sitl and opening things. We just want it to build the firmware and sitl app.

en/simulation/multi-vehicle-simulation.md
lamping7: The `_base` is assumed [here](https://github.com/PX4/Firmware/blob/4d0964385b84dc91189f377aafb039d10850e5d6/launch/single_vehicle_spawn.launch#L19)

You need to have xacros for different vehicles, set the `vehicle` arg in the launch accordingly and then have rcS files for them. I imagine this (the iris model) came from the ETHz RotorS package -- they have more examples [here](https://github.com/ethz-asl/rotors_simulator/tree/master/rotors_description/urdf) -- within the PX4 repo this will only work with the iris. That is why I put the links on the bottom if users what to do something more specialized for their purpose.

The launch file defines the SITL port, which is passed to xacro and used to generate urdf model with the specified port. So, the ROS param server xacro generated urdf gazebo models communicate on different ports for gazebo <--> SITL, BUT the rcS needs to be made manually with a `SITL_UDP_PRT` param that matches this port number to work.

Rules -- I don't think so. Just make sure the ports are available on your system and won't overlap other applications or vehicles. We do this a bit differently in the lab. In order to make this less confusing we script all this port stuff and use a base like 14400, such that we have 144xy, where we increment x for each vehicle and y matches the digits used in the normal single vehicle launch.

en/simulation/multi-vehicle-simulation.md
lamping7: Not sure why, but that's how I've seen it spelled. https://github.com/ethz-asl/rotors_simulator/blob/master/README.md

lamping7:
I made some non-formatting related comments above. I agree that much of this needs work and assumes you know what you're doing for the most part already. It is very confusing if you haven't been playing with it for a while. This is only for ROS SITL. Maybe see https://github.com/PX4/Firmware/pull/7235 where this was introduced for more info.

en/simulation/multi-vehicle-simulation.md
hamishwillee: I like code fragments because people can search for them and see them in the context of the current file. That said, you've thought this through, so good enough for now.  Thanks.

hamishwillee:
@lamping7 Thanks for this. I suggest we merge this to get it in now, and then work on improving it going forward (?). 
If you agree with this, do you think you can take ownership of making this a bit less "ROS expert-ish" :-)

lamping7:
@hamishwillee I'd merge it in now (after you correct the typo and formatting issues), as it is more correct in content over what is existing. 

However, after reading it through again, I do see what you mean about code snippets, as it could be easy to get lost in the first few "What's happening" bullet points. I'm not sure what kind of audience you want this written for. I assume it is for a research developer who already knows how to use ROS, Gazebo, and possibly MAVROS with PX4 and wants to expand from single to multiple vehicle simulations for their specific application. At the other end of the spectrum, I'm also not sure if we want an intro to ROS or Gazebo guide here. I'm my opinion, this file exists to show the capability of using PX4 for multi-vehicle simulations and give a basic description of the pieces that are needed to do a multi-vehicle simulation through this one scenario or application. I see this as a basis for something users are looking to develop, not something that will be used out-of-the-box. It think it is accurate that the original writer called this an "example setup." Thoughts here?

I can definitely help to clear up ambiguities and fill in gaps, but I'll need help identifying where we have content that is lacking.

Lastly, I think you're correct that it wouldn't hurt to say that this will: Open the Gazebo client GUI (unless `gui:=false` is specified in the `roslaunch` command) showing two iris vehicles in an empty world and allow you to control the vehicles with QGC and MAVROS similar to a single vehicle. The difference being that QGC will have a drop-down to select the vehicle that is "in focus" and for MAVROS control you will need to include the proper namespace before the topic/service path (e.g. for `<group ns="uav1">` you'll use **/uav1**/mavros/mission/push).

en/simulation/multi-vehicle-simulation.md
hamishwillee: OK, thanks

hamishwillee:
Hi @lamping7 

OK, so I have fixed the minor layout issues and added your introduction explaining what the example does (I repeated some of this info in "build and test" because it is relevant to testing).

> However, after reading it through again, I do see what you mean about code snippets, as it could be easy to get lost in the first few "What's happening" bullet points. 

Yes. Something to think about. All I see in the [rendered version](https://dev.px4.io/en/simulation/multi-vehicle-simulation.html) is a sea of "heres". I think showing the lines might provide a little more context.

> It think it is accurate that the original writer called this an "example setup." Thoughts here?

I agree, this does match what the original writer intended - it is a basic setup with explanation. 

As to the audience, I think you're right about who they are and their level of knowledge. We do not want to reproduce all the gazebo, ROS an intro docs.

I think most of my concerns in https://github.com/PX4/Devguide/pull/461#issuecomment-365448240 have been addressed - essentially I think it is more clear what the thing does without having to run it, and how to test it once you've run the example. 

My main remaining concerns are that it might be difficult for people to extend in a couple of fairly obvious way - ie I want to use a different vehicle, I want to use with Gazebo but not ROS. Yes this was not the intent of the author, but I think it would be useful information for that research audience.  

Have a think about it , and thank you for your work on this. Really appreciated!

weiminshen99:
Hi, @lamping7,
Could you please give a short instruction how to run and test this new change? Thanks!
Wingman

lamping7:
I'm not sure what you're asking. Just follow the list of commands up top. You should use this guide: https://github.com/PX4/Devguide/pull/473

weiminshen99:
Hi, Anthony,

Thanks for your reply.

What I was hoping is that your new change would also fix the simulation of multiple-vehicles WITHOUT ROS. (with ROS, it works fine). 

In other words, do you know if the script Firmware/Tools/sitl_multiple_run.sh would work properly now? I have been trying to use that script for a long time, but never succeeded. I would really appreciate your help if you can point me to the right direction. Thanks!

Best wishes,

Wingman

lamping7:
This only address the ROS version.

For non-ROS, take a look at https://github.com/PX4/Firmware/issues/8512 and the other discussions/PRs that it links to.

PR #462, opened 2018-02-18
Clarifications added and typos fixed

barzanisar:


en/airframes/adding_a_new_frame.md
```diff
@@ -115,12 +115,14 @@ S: 0 1   6500   6500      0 -10000  10000
...
-* O: Indicates the output scaling (*1 in negative, *1 in positive), offset (zero here), and output range (-1..+1 here).  If you want to invert your PWM signal, the signs of the output scalings has to be changed. (```O:      -10000  -10000      0 -10000  10000```)
-* S: Indicates the first input scaler: It takes input from control group #0 (attitude controls) and the first input (roll). It scales the input * 0.6 and reverts the sign (-0.6 becomes -6000 in scaled units). It applies no offset (0) and outputs to the full range (-1..+1)
-* S: Indicates the second input scaler: It takes input from control group #0 (attitude controls) and the second input (pitch). It scales the input * 0.65 and reverts the sign (-0.65 becomes -6500 in scaled units). It applies no offset (0) and outputs to the full range (-1..+1)
+* M: Indicates two scalers for two control inputs. It indicates the number of control inputs mixer will receive.
+* O: Indicates the output scaling (*1 in negative, *1 in positive), offset (zero here), and output range (-1..+1 here).  If you want to invert your PWM signal, the signs of the output scalings have to be changed. (```O:      -10000  -10000      0 -10000  10000```)
```
hamishwillee: - what does "(*1 in negative, *1 in positive)" mean? - multiplying by one doesn't do anything, and these appear to be same. 
- output range is marked as "output range (-1..+1 here)" but the output value is listed as "-10000  10000" - ie the numbers aren't the same

So the descriptions don't make much sense. It would be good to understand what the output scaling, range and offset actually do - ie what ranges make sense.

en/airframes/adding_a_new_frame.md
```diff
@@ -115,12 +115,14 @@ S: 0 1   6500   6500      0 -10000  10000
...
+* S: Indicates the first input scaler: It takes input from control group #0 (Flight Control) and the first input (roll). It scales the roll control input * 0.6 and reverts the sign (-0.6 becomes -6000 in scaled units). It applies no offset (0) and outputs to the full range (-1..+1)
```
hamishwillee: - What are the possible control groups? WOuld be good to have a mapping.
- What are the inputs - you say the first one is "roll" (0) and the second one is pitch (1). Is that input mapping fixed? ie is 0 always roll?
- The description here about scaled units and outputting to the full range is interesting, but should be pulled out before this point.

en/airframes/adding_a_new_frame.md
```diff
@@ -115,12 +115,14 @@ S: 0 1   6500   6500      0 -10000  10000
...
+* S: Indicates the second input scaler: It takes input from control group #0 (Flight Control) and the second input (pitch). It scales the pitch control input * 0.65. It applies no offset (0) and outputs to the full range (-1..+1)
 
-Behind the scenes, both scalers are added, which for a flying wing means the control surface takes maximum 60% deflection from roll and 65% deflection from pitch, i.e., SERVO = (0.60 * roll) + (0.65 * pitch). As it is over-committed with 125% total deflection for maximum pitch and roll, it means the first channel (roll here) has priority over the second channel / scaler (pitch). 
+In short, the output of this mixer would be SERVO = ( (roll input * -0.6 + 0) + (pitch input * 0.65 + 0) ) * 1 + 0
```
hamishwillee: Pulling this out is useful. Might we worth adding as a `> **Note**`

en/airframes/adding_a_new_frame.md
```diff
@@ -115,12 +115,14 @@ S: 0 1   6500   6500      0 -10000  10000
...
+
+Behind the scenes, both scalers are added, which for a flying wing means the control surface takes maximum 60% deflection from roll and 65% deflection from pitch.
```
hamishwillee: You have dropped this comment " As it is over-committed with 125% total deflection for maximum pitch and roll, it means the first channel (roll here) has priority over the second channel / scaler (pitch). "

Now I don't know what that means, but it seems significant!

en/airframes/adding_a_new_frame.md
```diff
@@ -115,12 +115,14 @@ S: 0 1   6500   6500      0 -10000  10000
 
 Where each number from left to right means:
```
hamishwillee, line 118: @barzanisar At the top of this section is a line: "A typical mixer file is shown below (for general information about mixing see: [Concepts > Mixing](../concept/mixing.md)). "

Suggest you change this to (something like):

```
> **Note** First read [Concepts > Mixing](../concept/mixing.md). This provides background information required to interpret this mixer file.

A typical mixer file is shown below.
```

You will note that I just deleted a bunch of my own comments asking for more clarification ... because I didn't read the concepts article first :-)

en/concept/mixing.md
```diff
@@ -214,8 +214,10 @@ are performed as floating-point operations, the values stored in the definition
 file are scaled by a factor of 10000; i.e. an factor of 0.5 is encoded as 5000.
 
 Roll, pitch and yaw inputs are expected to range from -1.0 to 1.0, whilst the
-thrust input ranges from 0.0 to 1.0.  Output for each actuator is in the 
+thrust input ranges from 0.0 to 1.0. Output for each actuator is in the 
 range -1.0 to 1.0.
 
+Idlespeed can range from 0.0 to 1.0. Idlespeed is relative to the maximum speed of motors and it is the speed with which the motors are commanded to rotate when all control inputs are zero.
```
hamishwillee: "it is the speed at which" (at, rather than "with")

en/middleware/modules_communication.md
```diff
@@ -35,8 +35,8 @@ The implementation uses 2 threads, a sending and a receiving thread. The sender
 reduces the rates of the streams if the combined bandwidth is higher than the configured rate (`-r`) or the
 physical link becomes saturated. This can be checked with `mavlink status`, see if `rate mult` is less than 1.
 
-**Careful**: some of the data is accessed and modified from both threads, so when changing code or extend the
-functionality, this needs to be take into account, in order to avoid race conditions and corrupt data.
+**Careful**: some of the data is accessed and modified from both threads, so when changing code or extending the
```
hamishwillee: Should be a note like this 

> **Caution** Some of the data is accessed and modified from both threads. When changing code/extending functionality care must be taken in order to avoid race conditions and corrupt data.

en/ros/external_position_estimation.md
```diff
@@ -83,7 +83,7 @@ Be sure to perform the following checks:
 
 If those steps are consistent, you can try your first flight.
 
-Put the robot on the ground and start streaming mocap feedback. Lower your left stick and arm the motors.
+Put the robot on the ground and start streaming mocap feedback. Lower your left (throttle) stick and arm the motors.
```
hamishwillee, line 86: Good. I'd be tempted to remove "left" and just have "throttle" (and change left/right elsewhere too).

en/simulation/hitl.md
```diff
@@ -59,12 +59,12 @@ In summary, HITL runs PX4 on the actual hardware using standard firmware, but SI
 
       ![Select Airframe](../../assets/gcs/qgc_hil_config.png)
 1. Setup UDP
-   1. Under the *General* tab of the settings menu, uncheck all *AutoConnect* boxes except for **UDP**. 
+   1. Under the *General* tab of the settings menu, uncheck all *AutoConnect* boxes except for **UDP**. This will disconnect the vehicle from QGC.
```
hamishwillee: Really? Even if correct, is it useful to note here?

hamishwillee:
Thanks @barzanisar Looks good - I put a few suggestions inline.

FYI Normally it is useful to put related changes in their own PRs. 99% of this could have been accepted as-is, but that all has to wait on the one or two clarifications in other docs.

en/airframes/adding_a_new_frame.md
barzanisar: According to the mixer example given above, the output is not over-committed since it is -60% roll + 65% pitch. Even if the output does get over-committed (e.g. the roll is -ve), the statement " the first channel (roll here) has priority over the second channel / scaler (pitch)" is wrong according to the px4 team. 

When the mixer output exceeds the defined limit of [-1,1], it just gets scaled to the max/min output limit i.e. 1 or -1.

en/simulation/hitl.md
barzanisar: It is true because QGC is connected to PX4 via serial port. Serial port's equivalent in QGC is "Pixhawk", which when we uncheck, px4 gets disconnected from QGC.

When I was following these steps, my QGC disconnected and I got confused that something has gone wrong, until a px4 team member clarified that it is normal. So I think, it will be helpful to let the user know that QGC disconnecting at this point is normal and part of the process.

barzanisar: Also, the HITL Simulation Environment Diagram above is wrong according to @bkueng. In HITL, there is no serial connection between QGC and PX4 since we uncheck "Pixhawk" in this step. QGC is connected to the Simulator via UDP port and then the Simulator is connected to px4 via serial port. It is something like this. 
@bkueng can you please confirm?

bkueng: No, this is not correct. QGC does not disconnect when unchecking the box. @barzanisar I'm not sure what you refer to, but we experienced another oddity: QGC showed some mavlink error message when we were connected to the Pixracer WiFi and USB at the same time. I don't remember exactly, but I think it was a problem between mavlink 1 and 2.

@barzanisar yes, the updated HITL PR is here: https://github.com/PX4/Devguide/pull/443.

en/airframes/adding_a_new_frame.md
```diff
@@ -115,12 +117,14 @@ S: 0 1   6500   6500      0 -10000  10000
...
-* S: Indicates the second input scaler: It takes input from control group #0 (attitude controls) and the second input (pitch). It scales the input * 0.65 and reverts the sign (-0.65 becomes -6500 in scaled units). It applies no offset (0) and outputs to the full range (-1..+1)
+* M: Indicates two scalers for two control inputs. It indicates the number of control inputs the mixer will receive.
+* O: Indicates the output scaling (*1 in negative, *1 in positive), offset (zero here), and output range (-1..+1 here).  If you want to invert your PWM signal, the signs of the output scalings have to be changed. (```O:      -10000  -10000      0 -10000  10000```)
+* S: Indicates the first input scaler: It takes input from control group #0 (Flight Control) and the first input (roll). It scales the roll control input * 0.6 and reverts the sign (-0.6 becomes -6000 in scaled units). It applies no offset (0) and outputs to the full range (-1..+1)
+* S: Indicates the second input scaler: It takes input from control group #0 (Flight Control) and the second input (pitch). It scales the pitch control input * 0.65. It applies no offset (0) and outputs to the full range (-1..+1)
```
RicardoM17, line 123: This line (123) is not consistent with the example above in which the sign is not reverted, so either change this line or the example above.
The error is most likely due to choosing the wrong block of code from the complete mixer. The first block of code (for Servo 0 was chosen) but the explanation is only correct for the second block of code (for Servo 1).

en/airframes/adding_a_new_frame.md
hamishwillee: I see that is done. Thanks.

en/airframes/adding_a_new_frame.md
hamishwillee: @barzanisar Can you check this

en/simulation/hitl.md
hamishwillee: @barzanisar Can you remove that line. Other than this, and @RicardoM17 comment (https://github.com/PX4/Devguide/pull/462#discussion_r169814109) this is good to go.

en/airframes/adding_a_new_frame.md
barzanisar: @RicardoM17  line 123 doesn't say the sign is reverted. Previously (line 121) it said it reverts the sign and this is why I changed it. I don't understand why line 123 is still not consistent with the example given. Can you please explain?

en/airframes/adding_a_new_frame.md
barzanisar: @hamishwillee Do you know who wrote this line 123?

en/airframes/adding_a_new_frame.md
RicardoM17: @barzanisar maybe it was already fixed but in the documentation ( see here: https://dev.px4.io/en/airframes/adding_a_new_frame.html) it still says revert the sign. But you are correct that line 123 here doesn't say it reverts the sign. However line 123 here is not consistent with the one in the link I sent you. I was discussing this with @hamishwillee and when I commented here I didn't re-read everything (I assumed it was the same) so maybe it's already been fixed.

hamishwillee: Sounds like there is no problem them. Thanks guys for working on this.

hamishwillee:
Thanks @barzanisar . I merged this. I had to undo some of your changes to multi uav simulation. This just got separately updated and I know the master version works. If you want to look at this again and do further changes, please handle as a separate PR.

Issue #427, opened 2018-01-09
Add additional controller diagrams


hamishwillee:
As discussed in https://github.com/PX4/Firmware/issues/8316#issuecomment-354956186

@bresch and @Stifael volunteered!

bresch:
Here is the pitch controller (roll is the same but with _phi_ and _p_ instead of _theta_ and _q_)
![fw_pitch_controller](https://user-images.githubusercontent.com/14822839/35610610-3f05aa8c-0662-11e8-8ed5-1e4700b356bd.png)
@tubeme FYI

hamishwillee:
Thanks very much @bresch ! I hope to get these into draw.io next week.

hamishwillee:
@bresch Thank you for this. I've created a version on draw.io here you can edit further: https://drive.google.com/file/d/1ibxekmtc6Ljq60DvNMplgnnU-JOvKYLQ/view?usp=sharing 

![image](https://user-images.githubusercontent.com/5368500/35897881-8993371c-0c17-11e8-838b-2fdb2b1cdfc7.png)

A few statements/questions:
1. The subscript r is "required" I presume? I have changed this everywhere to d (desired) as that is consistent with other image. 
1. Everything here is a scalar right (ie I have removed all bold)
1. Please check I got it right. 
1. What "words" would you like with this - see https://dev.px4.io/en/flight_stack/controller_diagrams.html#multicopter-position-controller

Once you're sure this is OK I'll modify to create the roll controller.

@Stifael @bkueng Anything you would like to suggest/add for this?

bresch:
@hamishwillee Thank you for translating it into _draw.io_.
1. You did it right, I didn't respect the notation there!
2. Correct, however I'm thinking about including roll and yaw in the same diagram. It would also be nicer to have a vector going to the Jacobian block. I don't want to have mathematician trying to kill me xD
3. I will do the modifications and tell you when I'm done

hamishwillee:
Thanks very much. I hope you can also help with "4" :-)

bresch:
@hamishwillee I updated the schematic and it includes roll and yaw now.
![fw_attitude_controller](https://user-images.githubusercontent.com/14822839/36304964-ed770d66-1311-11e8-985a-879f18ca6781.png)

hamishwillee:
Thanks @bresch . PR created https://github.com/PX4/Devguide/pull/464
[...]

PR #464, opened 2018-02-19
Add FW attitude controller diagram

hamishwillee:
This adds the FW attitude controller diagram fromhttps://github.com/PX4/Devguide/issues/427#issuecomment-366207384

@bresch @Stifael  This just includes the diagram. Can you provide some general comments to go with the image - in the same way as for the [Multicopter position controller](https://dev.px4.io/en/flight_stack/controller_diagrams.html#multicopter-position-controller)?

I have made this a branch on the master repo so you should be able to edit directly ... but if not, just send me the text you'd like.


bresch:
In a nutshell, the attitude controller works using a cascaded loop method. The outer loop computes the error between the attitude setpoint and the estimated attitude that, multiplied by a gain (P controller), generates a rate setpoint. The inner loop then computes the error in rates and uses a PI (proportional + integral) controller to generate the desired angular acceleration.

The angular position of the control effectors (ailerons, elevators, rudders, ...) is then computed using this desired angular acceleration and _a priori_ knowledge of the system through _control allocation_ (also known as _mixing_). Furthermore, since the control surfaces are more effective at high speed and less effective at low speed, the controller - tuned for cruise speed - is scaled using the airspeed measurements (if such a sensor is used). ~(or estimated airspeed if such a sensor is not available on the aircraft)~ **NO**.

The feedforward gain is used to compensate for aerodynamic damping. Basically, the two main components of body-axis moments on an aircraft are produces by the control surfaces (ailerons, elevators, rudders, ... - producing the motion) and the aerodynamic damping (proportional to the body rates - counteracting the motion). In order to keep a constant rate, this damping can be compensated using feedforward in the rate loop.

The roll and pitch controllers have the same structure and the longitudinal and lateral dynamics are assumed to be uncoupled enough to work independently. The yaw controller, however, generates its yaw rate setpoint using the _turn coordination_ constraint in order to minimize lateral acceleration, generated when the aircraft is slipping. The yaw rate controller also helps to counteract adverse yaw effects (https://youtu.be/sNV_SDDxuWk) and to damp the _Dutch roll mode_ (https://en.wikipedia.org/wiki/Dutch_roll) by providing extra directional damping.

en/flight_stack/controller_diagrams.md
```diff
@@ -13,4 +13,21 @@ The diagrams use the standard [PX4 notation](../contribute/notation.md) (and eac
...
+
+<!-- The drawing is on draw.io: https://drive.google.com/file/d/1ibxekmtc6Ljq60DvNMplgnnU-JOvKYLQ/view?usp=sharing . Request access from dev team. -->
+
+The attitude controller works using a cascaded loop method. The outer loop computes the error between the attitude setpoint and the estimated attitude that, multiplied by a gain (P controller), generates a rate setpoint. The inner loop then computes the error in rates and uses a PI (proportional + integral) controller to generate the desired angular acceleration.
+
+The angular position of the control effectors (ailerons, elevators, rudders, ...) is then computed using this desired angular acceleration and a priori knowledge of the system through control allocation (also known as mixing). Furthermore, since the control surfaces are more effective at high speed and less effective at low speed, the controller - tuned for cruise speed - is scaled using the airspeed measurements (if such a sensor is used).
+
+> **Note** If no airspeed sensor is used then gain scheduling for the FW attitude controller is  disabled (it's open loop); no correction is/can be made in TECS using airspeed feedback.
```
hamishwillee, line 29: @bresch FYI This note is the only change to your original text.

hamishwillee:
@bresch Thanks for your comments - now added. 

Attitude vector is shown in the diagram index, but not in the diagram. Conversely, the roll, pitch, yaw are shown it the diagram but not the index. Is there a reason?

Are you otherwise OK for this to be published.

bresch:
@hamishwillee Thank you!
The attitude vector is used in the diagram (just before the jacobian) (_Edit: **Correct!** I used phi instead of psi in the index, thanks!_), but you're right, I'll add roll/pitch/yaw in the index.

bresch:
@hamishwillee Is it fine like this or would you prefer to have roll/pitch/yaw on a new line?
fw_att_control

hamishwillee:
Hi @bresch 

> @hamishwillee Is it fine like this or would you prefer to have roll/pitch/yaw on a new line?

Thanks for the update. I like it as you have done it- but as per discussion in Slack ( https://px4.slack.com/archives/C3B9NSHRQ/p1519340952000122 ) we should collaborate to get the best possible diagram with others. 

I'm going to merge this as is for now - no point making people wait. 





 as it is and I will merge as is.

bresch:
@hamishwillee Absolutely, we can now iterate on it and add more useful details instead of having many people re-doing the same effort. Asking and sketching control diagrams is a recurrent topic in discuss.px4.io and slack, I hope this will answer many questions (it was also a thing that I was looking for when I started to work with PX4).

Side question for you: Is it possible to justify paragraphs in the devguide?

hamishwillee:
> @hamishwillee Absolutely, we can now iterate on it and add more useful details instead of having many people re-doing the same effort. Asking and sketching control diagrams is a recurrent topic in discuss.px4.io and slack, I hope this will answer many questions (it was also a thing that I was looking for when I started to work with PX4).

Great. I'm going to assume you are working on this and will call me in "if needed".

> Side question for you: Is it possible to justify paragraphs in the devguide?

It isn't an out-of-the-box option, but if you wanted to you could update the theme to do this. Would not be that hard, but I'm not sure it would be a real/significant improvement.

bresch:
> It isn't an out-of-the-box option, but if you wanted to you could update the theme to do this. Would not be that hard, but I'm not sure it would be a real/significant improvement.

I just asked because it would be nicer to my eyes, but if it's more that a few clicks/tags don't spend time on it, it won't change my life xD

hamishwillee:
THanks, I won't then. It is about 2 hours work including testing and pushing out to all our sites.

PR #472, opened 2018-02-23
Notation - Fix pitch and yaw symbols

bresch:
Theta and psi were swapped.


hamishwillee:
Thanks @bresch . Merged.

PR #473, opened 2018-03-01
Update multi-vehicle-simulation.md

lamping7:
make the firmware, then the sitl plugins @hamishwillee

### DIFF
--- a/en/simulation/multi-vehicle-simulation.md
+++ b/en/simulation/multi-vehicle-simulation.md
@@ -21,6 +21,7 @@ To build an example setup, follow the step below:
    ```
    cd Firmware_clone
    git submodule update --init --recursive
+   make posix_sitl_default
    make posix_sitl_default sitl_gazebo
    ```
 1. Source your environment:


hamishwillee:
I don't think so. 99.9% sure that everything that needs to be done is done with `make posix_sitl_default sitl_gazebo`.

Edited. 0% sure, after testing :-(

hamishwillee:
@bkueng This adds line shown to build PX4 SITL:
```
+   make posix_sitl_default
    make posix_sitl_default sitl_gazebo
```
Just FMI, why does `make posix_sitl_default sitl_gazebo` NOT build PX4, and exactly what does it do? THe option `sitl_gazebo` isn't even listed as a VMD target when using `make posix list_vmd_make_targets` .

The addition of this line is done in CI too, which is indicative that it is needed.

lamping7:
I'm not sure if this answers your question, but the magic happens here when you use `gazebo` rather than `sitl_gazebo`: https://github.com/PX4/Firmware/blob/2549b702f0fe71ce271e64d388eb7805abbd96d9/platforms/posix/cmake/sitl_target.cmake#L85

The target for `sitl_gazebo` just builds the external project, also defined in that file linked above. It doesn't depend on `px4`.

You can see targets with: `make posix_sitl_default help` or a massive list with `make help` if you're inside the firmware directory.

bkueng:
@lamping7 I think this is correct, so it's indeed required. I usually do `make posix_sitl_default gazebo`, and I never had to build PX4 separately.

hamishwillee:
@lamping7 @bkueng I've merged. Thanks for the explanation.

Issue #463, opened 2018-02-18
Testing and CI needs review and restructure


hamishwillee:
[Platform Testing and CI](https://dev.px4.io/en/test_and_ci/)

- The CI (https://dev.px4.io/en/test_and_ci/continous_integration.html) topic doesn't mention Jenkins (https://dev.px4.io/en/test_and_ci/jenkins_ci.html). It isn't clear then if the doc there is still up to date.
- Needs to present the things people most want to do first:
   - what tests do we have: integration, unit
   - how do I run them locally.
   - how do I add to them.
   - how to interpret test fails.
- The infrastructure stuff on docker, and jenkins etc is important, but is not first thing people need to see.
- Need to point to test information in source tree on what each test does and how to interpret fails.
- In future integration/Jenkins tests will be moving from ROS to DroneCore and will be simplified through an extensible "basic test". That test will need to be documented.

@dagar is expert and can help with confirming status of above.

julianoes:
Adding to this:
- The run_container.bash script is no longer there, so the instructions at https://dev.px4.io/en/test_and_ci/jenkins_ci.html#test-execution seem outdated.

PR #475, opened 2018-03-03
Added link to the example ported to python

nkhedekar:
This request is to add a link to the python version of the offboard example (webpage : https://dev.px4.io/en/ros/mavros_offboard.html) 

The main purpose is to provide python developers an example to help them start working with PX4, Gazebo and ROS

### DIFF
--- a/en/ros/mavros_offboard.md
+++ b/en/ros/mavros_offboard.md
@@ -195,3 +195,5 @@ while(ros::ok()){
 The rest of the code is pretty self explanatory. We attempt to switch to *Offboard* mode, after which we arm the quad to allow it to fly. We space out the service calls by 5 seconds so to not flood the autopilot with the requests. In the same loop, we continue sending the requested pose at the appropriate rate.
 
 > **Tip** This code has been simplified to the bare minimum for illustration purposes. In larger systems, it is often useful to create a new thread which will be in charge of periodically publishing the setpoints.
+
+The same example in python can be found at https://github.com/nkhedekar/offb_py


hamishwillee:
Thanks @nkhedekar - excellent.

@TSC21 Generally speaking I'm not a fan of links to "unofficial" github repos. IMO it would be better to either have both examples existing in a repo somewhere where they can easily be tested and kept in sync. Does the [original version](https://dev.px4.io/en/ros/mavros_offboard.html) exist in a repo anywhere? 

If not, should we do so. Either way, might be good to minimally include the Python script inline in this page as well.

Thoughts?

lamping7:
@hamishwillee Offboard examples can be seen in the CI test python source files. https://github.com/PX4/Firmware/tree/master/integrationtests/python_src/px4_it/mavros

There is some bloat due to the rostest framework, but it should serve as a good working example.

hamishwillee:
Thanks @lamping7. @nkhedekar Do you think perhaps instead linking to https://github.com/PX4/Firmware/tree/master/integrationtests/python_src/px4_it/mavros would achieve what you want for this section?

nkhedekar:
Thank you @hamishwillee  and @lamping7.

@hamishwillee yes, I think adding this link (https://github.com/PX4/Firmware/tree/master/integrationtests/python_src/px4_it/mavros) would serve as a good starting point for python developers as it has multiple examples and is properly commented.

lamping7:
These should probably be documented in the CI section of the guide and linked to. But, what is preventing this work, at least for me, are the unknowns about the eventual migration away from the ROS test framework, hinted at here: https://github.com/PX4/Devguide/issues/463 (i.e. I'm not going to spend time writing this if it's going to be tossed out.).

hamishwillee:
@nkhedekar Thanks. I have created https://github.com/PX4/Devguide/pull/490 for the links to integration test code. Will close this now.

hamishwillee:
> These should probably be documented in the CI section of the guide and linked to. But, what is preventing this work, at least for me, are the unknowns about the eventual migration away from the ROS test framework, hinted at here: #463 (i.e. I'm not going to spend time writing this if it's going to be tossed out.).

@lamping7 The plan in #463 is to use DroneCore for most testing, because it is far more lightweight than ROS and puts us much closer to PX4 when testing PX4 functionality.

I think it is worth documenting all the tests to some minimal level even if they might disappear in future. Though I guess it is worth waiting on https://github.com/PX4/Firmware/pull/8915 so that a "pattern is established.

That said, best

PR #482, opened 2018-03-14
Reorganise/improve getting started docs

hamishwillee:
The getting started docs were pretty old and ugly. This change:
* moves the contributing section to the top level (end).
* Updates the initial setup to suggest more recent hardware (from "in progress" dev list). Also links to User guide docs on basic and advanced setup (removing duplication). 
* Removes the "Advanced Config". The relevant info is linked in the initial setup doc.

I think much cleaner structure.


bkueng (approved):
yep, this looks good. Thanks @hamishwillee

Issue #467, opened 2018-02-23
Document template for new modules


hamishwillee:
https://github.com/PX4/Firmware/pull/8711 defines a template for new modules. This new document markup strings (PRINT_MODULE_DESCRIPTION) and code to generate module doc from them (ie., [Modules & Commands Reference](https://dev.px4.io/en/middleware/modules_main.html)

The existence of the template should be documented and linked - perhaps from https://dev.px4.io/en/tutorials/tutorial_hello_sky.html - depends on what defines a "module". 

The doc strings should be documented as part of that.

@bkueng is person to talk to about this.

PR #483, opened 2018-03-14
Adds description for using out-of-tree message definition

ksschwabe:
Related to PR PX4/Firmware#9073.

I wonder whether we should possibly rename this page from out_of_tree_modules.md to something that now represents that it now also includes the out-of-tree uORB message definitions?

@hamishwillee : Do you think it is necessary for me to explain that the new uORB headers and sources will be generated in the same location as the normal uORB messages?

### DIFF
--- a/en/advanced/out_of_tree_modules.md
+++ b/en/advanced/out_of_tree_modules.md
@@ -38,9 +38,43 @@ px4_add_module(
 	)
 
 ```
+
+
+# Out-of-tree uORB Message Definitions
+
+uORB messages can also be defined out-of-tree. For this, the
+`$EXTERNAL_MODULES_LOCATION/msg` needs to exist.
+
+- Place all new message definitions within the `$EXTERNAL_MODULES_LOCATION/msg`
+  directory. The format of these new out-of-tree message definitions are the same
+  as for any other [uORB message definition](../middleware/uorb.md#adding-a-new-topic).
+- Add a file `$EXTERNAL_MODULES_LOCATION/msg/CMakeLists.txt` with content:
+
+```
+set(config_msg_list_external
+    <message1>.msg
+    <message2>.msg
+    <message3>.msg
+    PARENT_SCOPE
+    )
+
+```
+where `<message#>.msg` is the name of the uORB message definition file to be processed 
+and used for uORB message generation.
+
+The out-of-tree uORB messages will be generated in the same locations as the normal uORB messages.
+The uORB topic headers are generated in `<build_dir>/uORB/topics/`, and the message source files are
+generated in `<build_dir>/msg/topics_sources/`.
+
+The new uORB messages can be used like any other uORB message as described [here](../middleware/uorb.md#adding-a-new-topic).
+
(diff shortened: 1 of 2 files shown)


hamishwillee:
Hi @ksschwabe 

Thanks for this!

> I wonder whether we should possibly rename this page from out_of_tree_modules.md to something that now represents that it now also includes the out-of-tree uORB message definitions?

No, it is fine.

> @hamishwillee : Do you think it is necessary for me to explain that the new uORB headers and sources will be generated in the same location as the normal uORB messages?

I think that is a good idea.

Other things:
1. I think that it is worth updating https://dev.px4.io/en/middleware/uorb.html with a note or section on "out of tree messages". This can simply contain link to the new doc. I can add later if you like.
2. There will be some minor layout issues from this, but I will fix in a post-process.

hamishwillee:
Outside of scope for this fix - @bkueng (and @ksschwabe )

This doc could use a little more context on the when/why/limitations of external modules. So can you tell me:
- what are external modules typically used for? When would you use one?
- Do external modules get included in firmware builds when you build PX4? 
-  Are there things you can't do with them?


Where do you think it should live in the devguide tree (currently  [this doc](https://dev.px4.io/en/advanced/out_of_tree_modules.html) is an "Advanced Topic - but I'm trying to reduce that folder to a minimum, grouping topics by "what they are for")?
- As I see it we have 3 topics on apps/modules - [Writing an application](https://dev.px4.io/en/tutorials/tutorial_hello_sky.html), [Out of Tree Modules](https://dev.px4.io/en/advanced/out_of_tree_modules.html), and your new module template (see #467).
- We could leave this where it is - after all it is an advanced topic
- We could move it into "Getting Started" alongside the others
- We could remove [Writing an application](https://dev.px4.io/en/tutorials/tutorial_hello_sky.html) from getting started, and group all the related topics on writing modules in a "Writing Modules" section.
- Other?

Thoughts?

bkueng:
Looks good.

>  what are external modules typically used for? When would you use one?

It's for downstream adopters: people/companies who want to either replace or add one or more (proprietary) module - for example a vehicle-specific position controller. It makes it easier to see which modules are company-internal, and which ones are upstream (of course we encourage people to bring their changes upstream, but this is a method that can be used for cases where they can't/don't want to).

>  Do external modules get included in firmware builds when you build PX4?

I'm not sure I understand. They get included when following the instructions here.

>  Are there things you can't do with them?

Hmm, not sure. You can access all the usual PX4 API's, so you have pretty much all the flexibility.

> We could leave this where it is - after all it is an advanced topic

I think that makes sense. It's not something that you need to go though when getting started.

ksschwabe:
@hamishwillee: I have updated my PR with the description related to where the topics are generated.

I also added a link in https://dev.px4.io/en/middleware/uorb.html to point to the new out-of-tree uORB section.

hamishwillee (approved):

hamishwillee:
@ksschwabe Thanks very much! I've merged this now.  @bkueng Thanks for the review. I'll do a further minor tidy up with your additional info.

PR #494, opened 2018-03-20
Tidy up out of tree modules doc

hamishwillee:
@bkueng This improves readability - fixes up some of the rendering and explains why you would use external modules. Can you please just sanity check it.

It follows on from #483

### DIFF
--- a/en/advanced/out_of_tree_modules.md
+++ b/en/advanced/out_of_tree_modules.md
@@ -10,7 +10,10 @@ This topic explains how to add an external ("out of tree") module to the PX4 bui
 
 To create an external module:
 
-- Create a directory for external modules at the same level and with the same structure as Firmware (i.e. it must contain a directory called **src**). This is the *external directory* that we later refer to using `EXTERNAL_MODULES_LOCATION`.
+- Create an *external directory* directory for grouping the external modules:
+  - This can be located anywhere outside of the **Firmware** tree.
+  - It must have the same structure as **Firmware** (i.e. it must contain a directory called **src**).
+  - Later we refer to this directory using `EXTERNAL_MODULES_LOCATION`.
 - Copy an existing module (e.g. **examples/px4_simple_app**) to the external directory, or directly create a new module.
 - Rename the module (including `MODULE` in **CMakeLists.txt**) or remove it from the existing Firmware *cmake* build config. This is to avoid conflicts with internal modules.
 - Add a file **CMakeLists.txt** in the external directory with content:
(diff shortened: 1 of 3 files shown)


en/advanced/out_of_tree_modules.md
```diff
@@ -1,80 +1,74 @@
...
+# External Modules (Out-of-Tree)
 
-This describes the possibility to add an external module to the PX4 build.
+External modules provide a convenient mechanism for developers to manage/group proprietary modules that they want add to (or update in) PX4 firmware. External modules can use the same includes as internal modules and can interact with internal modules via uORB.
 
-External modules can use the same includes as internal modules and can interact
-with internal modules via uORB.
+This topic explains how to add an external ("out of tree") module to the PX4 build.
+
+> **Tip** We encourage you to contribute your changes into PX4, where possible!
```
bkueng, line 7: :+1:

en/advanced/out_of_tree_modules.md
```diff
@@ -1,80 +1,74 @@
...
-
-uORB messages can also be defined out-of-tree. For this, the
-`$EXTERNAL_MODULES_LOCATION/msg` needs to exist.
-
-- Place all new message definitions within the `$EXTERNAL_MODULES_LOCATION/msg`
-  directory. The format of these new out-of-tree message definitions are the same
-  as for any other [uORB message definition](../middleware/uorb.md#adding-a-new-topic).
+To create an external module:
+
+- Create a directory for external modules at the same level and with the same structure as Firmware (i.e. it must contain a directory called **src**). This is the *external directory* that we later refer to using `EXTERNAL_MODULES_LOCATION`.
```
bkueng: I don't quite understand what you mean with 'at the same level'

bkueng (commented):
Thanks @hamishwillee, this is indeed better.I have one remark, it looks good otherwise.

en/advanced/out_of_tree_modules.md
hamishwillee: I'm making an assumption here that the folder structure is:
```
Firmware
   /src
external_dir
   /src
```
It could of course be that external_dir can even be inside `Firmware/src`. Any rules for this?

bkueng: It can be anywhere, except inside Firmware, which does not make sense, since the reason to use external modules is to separate the trees (in terms of git and directory structure).

hamishwillee: Thanks. Fixed. I'll merge now :-)

Issue #466, opened 2018-02-23
 HIL Via Ethernet + RaPi


hamishwillee:
2 undocumented options in [Hardware in the Loop Simulation (HITL)](https://dev.px4.io/en/simulation/hitl.html)
1. You can do HIL via Ethernet (or even WiFi, but you need a good one with low latency). 
2. Raspberry Pi is supported. Startup config, including how to run jMAVSim:
[posix-configs/rpi/px4_hil.config](https://github.com/PX4/Firmware/blob/master/posix-configs/rpi/px4_hil.config).

PR #495, opened 2018-03-20
Add HITL via Ethernet for Raspberry Pi

bkueng:
Fixes https://github.com/PX4/Devguide/issues/466

### DIFF
--- a/en/simulation/hitl.md
+++ b/en/simulation/hitl.md
@@ -28,6 +28,10 @@ With Hardware-in-the-Loop (HITL) simulation the normal PX4 firmware is run on re
 
 JMAVSim or Gazebo (running on a development computer) are connected to the flight controller hardware via USB/UART. The simulator acts as gateway to share MAVLink data between PX4 and *QGroundControl*. 
 
+> **Note** The simulator can also be connected via UDP if the flight controller has networking support and uses a stable, low-latency connection (e.g. a wired Ethernet connection - WiFi is usually not sufficiently reliable).
+For example, this configuration has been tested with PX4 running on a Raspberry Pi connected via Ethernet to the computer (a startup configuration that includes the command for running jMAVSim can be found [here](https://github.com/PX4/Firmware/blob/master/posix-configs/rpi/px4_hil.config)).
+
+<span></span>
 > **Tip** Gazebo can additionally share MAVLink data with an offboard API!
 
 The diagram below shows the simulation environment:


en/simulation/hitl.md
```diff
@@ -26,7 +26,14 @@ With Hardware-in-the-Loop (HITL) simulation the normal PX4 firmware is run on re
 
 ### JMAVSim/Gazebo HITL Environment
 
-JMAVSim or Gazebo (running on a development computer) are connected to the flight controller hardware via USB/UART. The simulator acts as gateway to share MAVLink data between PX4 and *QGroundControl*. 
+JMAVSim or Gazebo (running on a development computer) are connected to the
```
hamishwillee: This is a bit complex to parse, and no longer "quite" reflects the diagram. How about if we pull out the RPi stuff into a note?

How about:

----

JMAVSim or Gazebo (running on a development computer) are connected to the flight controller hardware via USB/UART. The simulator acts as gateway to share MAVLink data between PX4 and *QGroundControl*. 

> **Note** The simulator/flight controller can also be connected via UDP if the flight controller has networking support and uses a stable, low-latency connection (e.g. a wired Ethernet connection - WiFi is not usually sufficiently reliable).  For example, this configuration has been tested using Raspberry Pi via Ethernet (a startup configuration that includes the command for running jMAVSim can be found [here](https://github.com/PX4/Firmware/blob/master/posix-configs/rpi/px4_hil.config)).

hamishwillee (changes requested):
I'm not sure this fully addresses #466. This makes it clear that some setup with RaPi works and can use Ethernet rather than serial. However it isn't clear what the environment looks like and the instructions aren't detailed enough to understand how to set this up.

- I might assume that JMAVSim/Gazebo are running on RaPi, or I might assume that RaPi is just a bridge between the computer and flight controller.
- I don't know if the instructions in https://github.com/PX4/Devguide/blob/6726c8357da5e21417ee6c0b097fa79dbe4640a4/en/simulation/hitl.md#setting-up-hitl will work for RaPi - ie is it the same setup for QGC, simulator etc. I suspect not because I have a different config. 

Upshot, don't we need enough so that someone can easily set up this config?

en/simulation/hitl.md
bkueng: I almost did it that way but then I thought we start to have too many Notes and Tips there.

bkueng:
> I might assume that JMAVSim/Gazebo are running on RaPi, or I might assume that RaPi is just a bridge between the computer and flight controller.

jMAVSim is running on the PC and PX4 directly on the RPi.

> ie is it the same setup for QGC, simulator etc. I suspect not because I have a different config.

None of these steps are required since the startup script takes care of that, and the port does not interfere with QGC autoconnect.

Ok, so it needs some further explanations. Should I take a second pass or do you want to do it?

hamishwillee:
Better if you do it - I hesitate to doc something I have never tried.

en/simulation/hitl.md
hamishwillee: Up to you, but either way that sentence needs to be made more readable. You could use my text in the note, removing the "Note" bit.

bkueng:
Ok, I updated accordingly.

hamishwillee:
Hi @bkueng 

Thanks for this. Looks good. Merged.

For RaPi I'm assuming that, modulo building PX4 for RaPi](https://dev.px4.io/en/setup/building_px4.html#raspberry-pi-23-boards) the QGC and other setup is "exactly the same" for HITL as the other platforms. If that is not the case it would be good to have an RPI section at the end of the doc with links to other setup docs. In fact it would be nice to have this section anyway with that statement and links to the RaPi setup - so there is no question. Something to think about and add if you agree.

PR #496, opened 2018-03-20
Add a page to document the module template

bkueng:
@hamishwillee it's a bit short, which doesn't mean I want you to extend it, I just didn't know what else to write :)
Also, the Hello Sky page already describes various aspects in more detail, like uORB subscriptions, and I didn't want to duplicate that.

Fixes https://github.com/PX4/Devguide/issues/467

### DIFF
--- a/en/tutorials/module_template.md
+++ b/en/tutorials/module_template.md
@@ -1,22 +1,27 @@
-# Module Template
+# Module Template to write a Full Application
 
 The PX4 Firmware contains a template for writing a new application, which can be
 found under
 [src/templates/module](https://github.com/PX4/Firmware/tree/master/src/templates/module).
 It can be used for writing an application that runs on its own
-stack (as opposed to a work-queue task. This will be explained in more detail on
+stack (as opposed to a work-queue task. This is explained in more detail on
 the [Architecture Overview page](../concept/architecture.md#runtime-environment)).
 
-The template includes various aspects that might be useful for a full
+> **Note** All the things learned in the [First Application
+> Tutorial](tutorial_hello_sky.md) are relevant for writing a full application,
+> but the template contains a few more things.
+
+The template includes various aspects that are required or are useful for a full
 application:
 
 - Accessing parameters and reacting to parameter updates.
 - uORB subscriptions and waiting for topic updates.
 - Controlling the task that runs in the background via `start`/`stop`/`status`.
   The `module start [<arguments>]` command can then be directly added to the
-  startup script.
+  [startup script](../advanced/system_startup.md).
 - Command-line argument parsing.
-- Documentation: the `PRINT_MODULE_*` methods serve two purposes:
+- Documentation: the `PRINT_MODULE_*` methods serve two purposes (and the API is
+  documented [in the source code](https://github.com/PX4/Firmware/blob/master/src/platforms/px4_module.h#L387)):
   - They are used to print the command-line usage when entering `module help` on the console.
   - They are automatically extracted via script to generate the
 [Modules & Commands Reference](../middleware/modules_main.md) page.
--- a/en/tutorials/module_template.md
+++ b/en/tutorials/module_template.md
@@ -1,0 +1,24 @@
+# Module Template
+
+The PX4 Firmware contains a template for writing a new application, which can be
+found under
+[src/templates/module](https://github.com/PX4/Firmware/tree/master/src/templates/module).
+It can be used for writing an application that runs on its own
+stack (as opposed to a work-queue task. This will be explained in more detail on
+the [Architecture Overview page](../concept/architecture.md#runtime-environment)).
+
+The template includes various aspects that might be useful for a full
+application:
+
+- Accessing parameters and reacting to parameter updates.
+- uORB subscriptions and waiting for topic updates.
+- Controlling the task that runs in the background via `start`/`stop`/`status`.
+  The `module start [<arguments>]` command can then be directly added to the
+  startup script.
+- Command-line argument parsing.
+- Documentation: the `PRINT_MODULE_*` methods serve two purposes:
+  - They are used to print the command-line usage when entering `module help` on the console.
+  - They are automatically extracted via script to generate the
+[Modules & Commands Reference](../middleware/modules_main.md) page.
+
+
--- a/en/SUMMARY.md
+++ b/en/SUMMARY.md
@@ -14,6 +14,7 @@
     * [Fast RTPS installation](setup/fast-rtps-installation.md)
   * [Building the Code](setup/building_px4.md)
   * [Writing an Application](tutorials/tutorial_hello_sky.md)
+  * [Module Template](tutorials/module_template.md)
 * [Concepts](concept/README.md)
   * [PX4 Architectural Overview](concept/architecture.md)
     * [Controller Diagrams](flight_stack/controller_diagrams.md) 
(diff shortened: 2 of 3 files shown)


en/tutorials/module_template.md
```diff
@@ -0,0 +1,24 @@
+# Module Template
+
+The PX4 Firmware contains a template for writing a new application, which can be
+found under
+[src/templates/module](https://github.com/PX4/Firmware/tree/master/src/templates/module).
+It can be used for writing an application that runs on its own
+stack (as opposed to a work-queue task. This will be explained in more detail on
```
hamishwillee: will be -> "is"

en/tutorials/module_template.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Accessing parameters and reacting to parameter updates.
+- uORB subscriptions and waiting for topic updates.
+- Controlling the task that runs in the background via `start`/`stop`/`status`.
+  The `module start [<arguments>]` command can then be directly added to the
+  startup script.
+- Command-line argument parsing.
+- Documentation: the `PRINT_MODULE_*` methods serve two purposes:
+  - They are used to print the command-line usage when entering `module help` on the console.
+  - They are automatically extracted via script to generate the
```
hamishwillee: Is there nothing else that needs to be done - ANY module with these strings will have the commands reference generated?

hamishwillee: ... probably even if it is true, we should say where it is generated, and note that it still needs to be copied into the library and TOC needs to be updated.

en/tutorials/module_template.md
```diff
@@ -0,0 +1,24 @@
...
+stack (as opposed to a work-queue task. This will be explained in more detail on
+the [Architecture Overview page](../concept/architecture.md#runtime-environment)).
+
+The template includes various aspects that might be useful for a full
+application:
+
+- Accessing parameters and reacting to parameter updates.
+- uORB subscriptions and waiting for topic updates.
+- Controlling the task that runs in the background via `start`/`stop`/`status`.
+  The `module start [<arguments>]` command can then be directly added to the
```
hamishwillee, line 20: link to docs on startup script?

hamishwillee:
Hi @bkueng 

First, thanks for taking the time to write this. It needed to be done, and realistically I wasn't going to get around to it any time soon.

The thoughts that occurred to me as I read this:
1. How does the module created using the template differ from the type of module in Hello Sky tutorial? 
  - If they are essentially the same and we want people to start using the module template then I think the hello sky should be re-written based on this template and the topics merged.
  - if they are different, then I think we need to make a clear statement about the types of modules you can write and where the "basic app" vs the module style app.  We probably need to have a comment about that in the "Hello sky doc anyway".
2. I'd like to see expanded documentation of the PRINT_MODULE_*  methods. Listed out with comments about what they do. 
3. Is this in the right place - "Getting Started"? Essentially same question as I asked for out of tree modules - if there are enough types of modules and ways of building them, then grouping them might make sense. 

So the upshot is that the content is good and useful, but it does feel a bit like it is dangling. If this template is a useful tool, would love to get hello-sky to use it.

en/tutorials/module_template.md
bkueng: Yes it's fully automatic and the Reference page already contains the command how to generate it (`make module_documentation`). The TOC only needs to be updated for a new/unknown category & the script will complain it that case.
The copying is done via CI.

bkueng:
> How does the module created using the template differ from the type of module in Hello Sky tutorial?

- hello sky: it's a minimal example to start tampering with PX4 code and get familiar with the most important things (uORB for example), while ignoring irrelevant things like background tasks or module documentation. It serves more as a learning experience, the template should then be used to write the 'full' application.
- template: full-blown module with everything needed to get it upstream (pretty much a superset of the hello sky, but assuming you already know about the basics).

I think separating it makes it less overwhelming, blending out less relevant things in the beginning. But if you think we can do all in one, I'm ok with that. But I'd like to avoid adding comments explaining every detail to the template.

> I'd like to see expanded documentation of the PRINT_MODULE_* methods. Listed out with comments about what they do.

The API is documented in the source (https://github.com/PX4/Firmware/blob/master/src/platforms/px4_module.h#L387). Do you think linking to that is enough? It will make maintenance easier.

> Is this in the right place - "Getting Started"?

I would say so since it's a follow-up to the hello sky example for building a full module.

hamishwillee:
>> I'd like to see expanded documentation of the PRINT_MODULE_* methods. Listed out with comments about what they do.

> The API is documented in the source (https://github.com/PX4/Firmware/blob/master/src/platforms/px4_module.h#L387). Do you think linking to that is enough? It will make maintenance easier.

Yes, linking would be fine. Thanks.

> I think separating it makes it less overwhelming, blending out less relevant things in the beginning. But if you think we can do all in one, I'm ok with that. But I'd like to avoid adding comments explaining every detail to the template.

You've convinced me that it belongs there and that it is "different"  and should be separate to the hello sky. However it still looks "floating" - ie it is not clear that a topic named *Writing an Application* (hello sky) is a hello world and  *Module Template* is a "guide" to your real application.

Perhaps we should change Hello Sky into "Writing your First Application" and slightly refocus the new document as "Writing a Full Application". Essentially it would start by saying that everything you learned in the Hello Sky is relevant, but in a more fully featured app you need to do X, Y, Z, which this module template does for you out of the box. ....

Does that make any sense? Past my bed time so I could be gibbering :-(

hamishwillee:
Hi @bkueng, Did the above make sense? https://github.com/PX4/Devguide/pull/496#issuecomment-374896489

bkueng:
Yes, sorry, I did not get to update it yet. But I agree with you that it needs a bit more guidance. Your proposal sounds good.

hamishwillee:
Thanks! Just wanted to make sure you did see this - appreciate you're probably drowning in github notifications.

bkueng:
Ready for the next round - it might still need a bit more.

hamishwillee:
Thanks @bkueng Looks good. I merged and made some further updates in https://github.com/PX4/Devguide/pull/499 (sorry, didn't realise immediately that these could have been done on the shared branch).

Generally I think the final version is OK. The style of the Module Template doc is a bit odd because it is supposed to be about advanced features, but actually all we do is state "these are the features", the implication being someone should go look. I have no problem with that.

Arguably the only other thing would be to provide some instructions on using the template 
- ie how and where to copy it, 
- what needs to be changed for uniqueness
- How to add it to firmware.

Thoughts?

bkueng:
> Arguably the only other thing would be to provide some instructions on using the template

The required modifications are the same as written in https://dev.px4.io/en/apps/hello_sky.html#minimal-application and the next section. Do you think this needs to be made more clear?

hamishwillee:
>> Arguably the only other thing would be to provide some instructions on using the template

> The required modifications are the same as written in https://dev.px4.io/en/apps/hello_sky.html#minimal-application and the next section. Do you think this needs to be made more clear?

It is clear from the tutorial perspective, but it is not very obvious what is actually _relevant_ in all that copy pasting. That is the downside of the tutorial format.

IFF this can be summarised to apply more concisely to the module template then it is worth doing.  Some of the steps below .... the fact that this is not complete tells you that I'm not 100% sure what actually needed to change - though I am sure I could work it out by actually trying it out.

----

# Using the Template

1. Copy /src/templates/module/ to anywhere in the Firmware /src tree
1. Rename the folder, cpp and h files to something appropriate - e.g. `myapp`. 
1. Rename class `Module` appropriately and update the main function to `<module_name>_main` and exported from the module as shown: .... some example
1. Update the headers and documentation.
1. Add the module to the Firmware cmake file - as discussed in https://dev.px4.io/en/apps/hello_sky.html#build-the-applicationfirmware

PR #500, opened 2018-04-03
Tidy up EKF warnings

hamishwillee:
This tidies up the doc so that:
1. Descriptions of parameters in text are linked to the headings below
2. Descriptions of parameters in sections are also linked to "official" documentation.
3. Bullets are indented 2 spaces. This merely tidies the rendering in github - things always worked in final gitbook version
4. Headings have space between text and heading marker (just improves rendering).

@priseborough A few questions:
1. Should we delete the descriptions of each of the parameters and instead just link to the official parameter docs (as I have done in the lower sections). The advantage of leaving text there is that it can be read "in context". Disadvantage is duplication and possibility of things getting out of sync.
1. We have a report of someone getting error "Preflight fail: EKF internal Checks", but this is not documented. Is there any way for someone to interpret this/what should they do? [Note, users says manual mode, RC input is there. battery is calibrated. Sensor *say* they are calibrated.]
1. These look like public messages (ie displayed/played in QGC). Shouldn't this information be in the user guide? The only problem is that I don't think those users need to know about the commander. I'd assume this would go along side https://docs.px4.io/en/flying/led_meanings.html

Thoughts?

### DIFF
--- a/en/tutorials/pre_flight_checks.md
+++ b/en/tutorials/pre_flight_checks.md
@@ -1,59 +1,70 @@
 # Preflight Sensor and EKF Checks
-The commander module performs a number of preflight sensor quality and EKF checks which are controlled by the COM_ARM<> parameters. If these checks fail, the motors are prevented from arming and the following error messages are produced:
+
+The commander module performs a number of preflight sensor quality and EKF checks which are controlled by the [COM_ARM_](../advanced/parameter_reference.md#commander) parameters. If these checks fail, the motors are prevented from arming and the following error messages are produced:
 
 * PREFLIGHT FAIL: EKF HGT ERROR
- * This error is produced when the IMU and height measurement data are inconsistent.
- * Perform an accel and gyro calibration and restart the vehicle. If the error persists, check the height sensor data for problems.
- * The check is controlled by the COM_ARM_EKF_HGT parameter.
+  * This error is produced when the IMU and height measurement data are inconsistent.
+  * Perform an accel and gyro calibration and restart the vehicle. If the error persists, check the height sensor data for problems.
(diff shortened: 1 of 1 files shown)


hamishwillee:
@priseborough Can you please comment?

priseborough:
@hamishwillee In answer to your questions:

1) I would prefer to link to the official parameter documentation.

2) The "Preflight fail: EKF internal Checks" message is generated if the innovation magnitudes of either the horizontal GPS velocity, magnetic yaw, vertical GPS velocity or vertical position sensor (Baro by default but could be range finder or GPS if non-standard params are being used) are excessive. Innovations are the difference between the value predicted by the inertial navigation calculation and measured by the sensor.

Users should check the innovation levels in the log file to determine the cause. These can be found under the ekf2_innovations message. Common problems include:

a) IMU drift on warmup. May be resolved by restarting the autopilot. May require an IMU accel and gyro cal.
b) Adjacent magnetic interference combined with vehicle movement. Resolve my moving vehicle and waiting or re-powering.
c) Bad magnetomer calibration combined with vehicle movement. Resolve by recalibrating.
d) Initial shock or rapid movement on startup that caused a bad inertial nav solution. Resolve by restarting the vehicle and minimising movement for the first 5 seconds.

3) That makes sense. The problem is that with the emergence of multiple types of external aiding, the concept of GPs lock is less important than it used to be and the ekf can have a useful nav solution without a GPS lock or a un-usable one with GPS lock. We need to start using the concept of navigation quality rather than GPS lock. A common complaint i see is that the user sees that the GPs has a 3D fix, but the lights appear to be saying there is no GPS lock.

hamishwillee:
Closing this due to oddly missing branch in my repo. Continued in https://github.com/PX4/Devguide/pull/510

PR #501, opened 2018-04-03
Not use root for reading output of /dev/video0

kikislater:


en/qgc/video_streaming.md
```diff
@@ -191,7 +191,10 @@ If you cannnot start the stream on the Odroid with the uvch264s plugin, you can
 ```sh
  gst-launch-1.0 v4l2src device=/dev/video0 ! video/x-h264,width=1920,height=1080,framerate=24/1 ! h264parse ! rtph264pay ! udpsink host=xxx.xxx.xxx.xxx port=5000
 ```
-Where `xxx.xxx.xxx.xxx` is the IP address where QGC is running. If you get the system error: `Permission denied`, you might need to prepend `sudo` to the  command above. 
+Where `xxx.xxx.xxx.xxx` is the IP address where QGC is running. If you get the system error: `Permission denied`, you might need to prepend `sudo` to the  command above. If you don't want to type `sudo` you should be member of `video` group. To do that simply type this command :
```
ndepal, line 194: It might be worth mentioning that this only takes effect after logging out.

hamishwillee: Ageed. I'll merge this and make that change. Thanks @kikislater !

hamishwillee:
@ndepal FYI Added note about login/logout as part of https://github.com/PX4/Devguide/pull/502

en/qgc/video_streaming.md
kikislater: Yes sorry, forgot to mention logout/login !

PR #503, opened 2018-04-04
Update PX4 Porting Guide organization for clarity

joshuawhitehead:
Attempted to group information in a more consistent (and easily expandable) manner. Updated language in various locations to add clarity and be more consistent within the porting guide and with respect to other documentation files.  Added links and references in a few locations to other PX4 documentation as appropriate.  Corrected multiple dead links.

### DIFF
--- a/en/debug/porting-guide.md
+++ b/en/debug/porting-guide.md
@@ -2,55 +2,67 @@
 
 This topic is for developers who want to port PX4 to work with *new* flight controller hardware.
 
-## Architecture
+## PX4 Architecture
 
-PX4 consists of two main layers: The board support and middleware layer on top of the host OS (NuttX, Linux or any other POSIX platform like Mac OS). And the applications (Flight Stack in [src/modules](https://github.com/PX4/Firmware/tree/master/src/modules)\).
+PX4 consists of two main layers: The [board support and middleware layer](https://dev.px4.io/en/middleware/) on top of the host OS (NuttX, Linux or any other POSIX platform like Mac OS), and the applications (Flight Stack in [src/modules](https://github.com/PX4/Firmware/tree/master/src/modules)\).  Please reference the [PX4 Architectural Overview](../concept/architecture.md) for more information.
 
-This guide is focused only on the middleware as the applications will run on any board target.
+This guide is focused only on the host OS and middleware as the applications/flight stack will run on any board target.
 
-## NuttX Boards
+## Flight Controller Configuration File Layout
 
-The location of the main files for a NuttX board are:
+In addition to the host operating system specific configuration files described below, there are several groups of configuration files for each board located throughout the code base:
+* Board startup and configuration files are located in: [src/drivers/boards](https://github.com/PX4/Firmware/tree/master/src/drivers/boards).
+  * This folder contains bus mappings, GPIO mappings, and the initialization code for each board.
+  * FMUv5 example: [src/drivers/boards/px4fmu-v5](https://github.com/PX4/Firmware/tree/master/src/drivers/boards/px4fmu-v5).
+* The boot file system (startup script) is located in: [ROMFS/px4fmu\_common](https://github.com/PX4/Firmware/tree/master/ROMFS/px4fmu_common)
+* The board specific build configurations are located in: [cmake/configs/](https://github.com/PX4/Firmware/blob/master/cmake/configs/).
+* Driver files are located in: [src/drivers](https://github.com/PX4/Firmware/tree/master/src/drivers).
 
-* Board startup and PX4 board configuration: [src/drivers/boards](https://github.com/PX4/Firmware/tree/master/src/drivers/boards). It contains bus and GPIO mappings and the board initialization code. 
-  * FMUv5 example: [src/drivers/boards/px4fmu-v5](https://github.com/PX4/Firmware/tree/master/src/drivers/boards/px4fmu-v5). 
-* NuttX board config: [nuttx-configs](https://github.com/PX4/Firmware/tree/master/nuttx-configs). The OS gets loaded as part of the application build. 
-  * FMUv5 example: [nuttx-configs/px4fmu-v5/include/board.h](https://github.com/PX4/Firmware/blob/master/nuttx-configs/px4fmu-v5/include/board.h).
-* NuttX OS config (created with menuconfig): [nuttx-configs/px4fmu-v5/nsh/defconfig](https://github.com/PX4/Firmware/blob/master/nuttx-configs/px4fmu-v5/nsh/defconfig).
-* Linker scripts and other required settings: [nuttx-configs](https://github.com/PX4/Firmware/tree/master/nuttx-configs). 
-  * FMUv5 example: [nuttx-configs/px4fmu-v5](https://github.com/PX4/Firmware/tree/master/nuttx-configs/px4fmu-v5).
-* Boot file system (startup script): [ROMFS/px4fmu\_common](https://github.com/PX4/Firmware/tree/master/ROMFS/px4fmu_common)
-* Board build configuration: [cmake/configs/nuttx\_px4fmu-v5\_default.cmake](https://github.com/PX4/Firmware/blob/master/cmake/configs/nuttx_px4fmu-v5_default.cmake).
-* Drivers: [src/drivers](https://github.com/PX4/Firmware/tree/master/src/drivers).
-* Reference config: Running `make px4fmu-v4_default` builds the FMUv4 config, which is the current NuttX reference configuration.
+## Host Operating System Configuration
 
+This section describes the purpose and location of the configuration files required for each supported host operating system to port them to new flight controller hardware.
 
-### NuttX Menuconfig
+### NuttX
 
-If you need to modify the NuttX [menuconfig](https://bitbucket.org/nuttx/nuttx) you can do this using the PX4 shortcuts:
+In order to port PX4 on NuttX to a new hardware target, that hardware target must be supported by NuttX.  The NuttX project maintains an excellent [porting guide](http://www.nuttx.org/Documentation/NuttxPortingGuide.html) for porting NuttX to a new computing platform.
+
+For all NuttX based flight controllers (e.g. the Pixhawk series) the OS is loaded as part of the application build.
+
+The configuration files for NuttX based boards, including linker scripts and other required settings are located under [platforms/nuttx/nuttx-configs](https://github.com/PX4/Firmware/tree/master/platforms/nuttx/nuttx-configs).
+
+The following example uses FMUv5 as it is the current [reference configuration](./reference-design.md) for NuttX based flight controllers:
+* Running `make px4fmu-v5_default` from the `src/Firmware` directory will build the FMUv5 config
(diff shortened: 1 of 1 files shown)


en/debug/porting-guide.md
```diff
@@ -2,55 +2,67 @@
 
 This topic is for developers who want to port PX4 to work with *new* flight controller hardware.
 
-## Architecture
+## PX4 Architecture
 
-PX4 consists of two main layers: The board support and middleware layer on top of the host OS (NuttX, Linux or any other POSIX platform like Mac OS). And the applications (Flight Stack in [src/modules](https://github.com/PX4/Firmware/tree/master/src/modules)\).
+PX4 consists of two main layers: The [board support and middleware layer](https://dev.px4.io/en/middleware/) on top of the host OS (NuttX, Linux or any other POSIX platform like Mac OS), and the applications (Flight Stack in [src/modules](https://github.com/PX4/Firmware/tree/master/src/modules)\).  Please reference the [PX4 Architectural Overview](../concept/architecture.md) for more information.
```
hamishwillee: Links within the same doc should be relative. So for example this `https://dev.px4.io/en/middleware/` should be `../middleware/README.md`

en/debug/porting-guide.md
```diff
@@ -2,55 +2,67 @@
...
+For all NuttX based flight controllers (e.g. the Pixhawk series) the OS is loaded as part of the application build.
 
-* The start script is located in [posix-configs/](https://github.com/PX4/Firmware/tree/master/posix-configs).
-* The OS configuration is part of the default Linux image (TODO: Provide location of LINUX IMAGE and flash instructions).
-* The PX4 middleware configuration is located in [src/drivers/boards](https://github.com/PX4/Firmware/tree/master/src/drivers/boards). TODO: ADD BUS CONFIG
-* Drivers: [DriverFramework](https://github.com/px4/DriverFramework).
-* Reference config: Running `make qurt_eagle_release` builds the Snapdragon Flight reference config.
+The configuration files for NuttX based boards, including linker scripts and other required settings are located under [platforms/nuttx/nuttx-configs](https://github.com/PX4/Firmware/tree/master/platforms/nuttx/nuttx-configs).
+
+The following example uses FMUv5 as it is the current [reference configuration](./reference-design.md) for NuttX based flight controllers:
```
hamishwillee: This link will work, as will `(reference-design.md)`. I have a slight preference for the full relative path `(../debug/reference-design.md)` as it reminds me this file needs to be moved to somewhere with a better name!

en/debug/porting-guide.md
```diff
@@ -2,55 +2,67 @@
...
+## Middleware Components and Configuration
 
+This section describes the various middleware components, and the configuration file updates needed to port them to new flight controller hardware.
 
+### QuRT / Hexagon
 
+* The start script is located in [posix-configs/](https://github.com/PX4/Firmware/tree/master/posix-configs).
+* The OS configuration is part of the default Linux image (TODO: Provide location of LINUX IMAGE and flash instructions).
+* The PX4 middleware configuration is located in [src/drivers/boards](https://github.com/PX4/Firmware/tree/master/src/drivers/boards). TODO: ADD BUS CONFIG
+* Drivers: [DriverFramework](https://github.com/px4/DriverFramework).
```
hamishwillee, line 64: This is true, though in future these are expected to use same drivers as NuttX

en/debug/porting-guide.md
hamishwillee: Calling FMUv5 the current reference config for NuttX based flight controllers feels a little odd because historically the variants are not necessarily sequential - ie FMU versions only really infer binary compatibility and a lower number could be the same level of advanced. I guess it is the most recent and the one we would hope manufacturers would target so leave as is.

hamishwillee:
@joshuawhitehead Yes, this is much much better.

@dagar Is there anything in your recent reorganisational work that invalidates any of this, or which should be captured (in addition)? If not, I say we merge.

en/debug/porting-guide.md
joshuawhitehead: Ah, ok, so anything that's in the "Devguide" repo should use a relative link, I should only use a full URL if I need to reference something else, e.g. the User Guide?  I was working through the GitHub interface to make this update and was a little unsure of how it handled things.  I will update this.

en/debug/porting-guide.md
joshuawhitehead: Agreed, that makes sense, I will update accordingly.

en/debug/porting-guide.md
joshuawhitehead: Would it be worth adding a "Future Plans" subsection to add things like that? In the chance you pique someone's interest to help make it happen :-)

en/debug/porting-guide.md
joshuawhitehead: Understood about the binary compatibility/feature set, I had read that in other places as well, I had intended it only as you mentioned- If someone is porting to a brand new board, FMUv5 would be the best place to start.  I thought I had seen a similar sentence in the reference-design.md doc, but now I'm not... I will add a note here to remind people that the versions/generations to not correlate with capabilities, simply that v5, being the most recent version, reflects the most up to date methods for hardware and code structure.

joshuawhitehead:
Thank you for all the feedback @hamishwillee !

To add some context: I've been chatting with @dagar on the PX4 slack channel as well, and I am currently working through doing a port for a Cortex-R5 based FMU. My plan is to add as many of my "lessons learned" here as I am able. If you are happy with the changes so far, and are willing to merge them as is, I will wait until that merge happens to make the updates I mentioned in my replies to your comments above.

My plan is to approach the porting guide from the perspective of "Ok, you have a NuttX `hello_world` booting on the board/chip you want to use, now what?", mostly because that's how I'm doing it... If you have any thoughts on a superior approach, I'd be happy to hear them. Some things I would like to see in the guide are:
  * A list of the minimal drivers required for a "working" PX4 port
    * Perhaps even a table showing the capabilities enabled by a particular driver beyond the minimum set as this would allow developers to target their efforts
  * While the cMake files are somewhat "self documenting", a high-level overview of the build process would be nice, particularly to call out any "gotchas" that a developer might encounter, e.g. PX4 does not (currently) use the entire NuttX build system, it only uses NuttX to build some pieces/libraries and does the final linking with its own cMake files
  * ~If you'd like to use a back-up coprocessor (e.g. the STM32F103 on the PixHawk), where does that fit into the whole build process, and what are its requirements?~ (Found this)

Just a few ideas I had at the moment, there will likely be more, and I plan to answer as many of them myself as I am able while working on this port.  Any feedback is welcome, I'm monitoring the PX4 slack channel as well if that would be an easier place to have a conversation.

en/debug/porting-guide.md
```diff
@@ -2,55 +2,67 @@
...
+* Reference config: Running `make qurt_eagle_release` builds the Snapdragon Flight reference config.
```
dagar: This target isn't valid.

en/debug/porting-guide.md
hamishwillee: Thanks!

en/debug/porting-guide.md
hamishwillee: That is correct. You've done it right throughout pretty much - except for this one.

en/debug/porting-guide.md
hamishwillee: I actually wouldn't add a note here (maybe back in the reference-design doc though!). 
What I'd do is change as shown:

FROM
> The following example uses FMUv5 as it is the current [reference configuration](./reference-design.md) for NuttX based flight controllers:

TO

> The following example uses FMUv5 as it is a recent [reference configuration](./reference-design.md) for NuttX based flight controllers:

en/debug/porting-guide.md
hamishwillee: Not for now. Daniel Agar is going to do this particular task and I don't know of other future plans. If you know of enough to justify them, then definitely.

en/debug/porting-guide.md
hamishwillee: @joshuawhitehead Can you fix this up too  - it should just be `make eagle_default` (according to [this](https://dev.px4.io/en/setup/building_px4.html#qurt--snapdragon-based-boards).

hamishwillee:
Hi @joshuawhitehead 

> Thank you for all the feedback @hamishwillee !

You're very welcome. I should be thanking you. I've been trying to get to this document for ages, but 
* I simply do not know enough about porting to take a decent shot at it. 
* Still haven't managed to arrange a discussion with the people who do. 

> If you are happy with the changes so far, and are willing to merge them as is, I will wait until that merge happens to make the updates I mentioned in my replies to your comments above.

Can you please make the minor and specific modifications suggested - I'm then happy to merge (if there are any in the current set of suggestions you think are significant they could wait).

> My plan is to approach the porting guide from the perspective of "Ok, you have a NuttX hello_world booting on the board/chip you want to use, now what?", mostly because that's how I'm doing it... If you have any thoughts on a superior approach, I'd be happy to hear them. 

I think this is a good way to approach the problem - at least you should be able to capture most the issues and build iteratively. At some point you may create enough material that a different organisation is worth considering, but we're a long way from that. 

YOu might start even earlier discussing what sorts of board porting "makes sense" - ie what chips you can use, where to modifications "make sense" (as I said, I haven't had a discussion with a board engineer yet to know what things are doable/reasonable, so this might just be silly).

> Some things I would like to see in the guide are:
> A list of the minimal drivers required for a "working" PX4 port
> Perhaps even a table showing the capabilities enabled by a particular driver beyond the minimum set as this would allow developers to target their efforts

Sounds good. There is lots of work going on with drivers at the moment to capture much of the copy paste used currently into base classes. @dagar is doing most of that, and I presume a lot of it will inject into the device driver docs. 

With respect to driver docs/tables, at the moment we auto-generate module docs for a small subset of [drivers](https://dev.px4.io/en/middleware/modules_driver.html) - a little more on this in https://dev.px4.io/en/apps/module_template.html

What I would like to do is autogenerate more complex user  documentation about what is supported (from source). For example, at the moment this is handcrafted: https://docs.px4.io/en/sensor/rangefinders.html - but we could generate it from in source docs using only minor script enhancements of the module documentation. Anyway, just FYI - on my list as drivers get updated. 

> While the cMake files are somewhat "self documenting", a high-level overview of the build process would be nice, particularly to call out any "gotchas" that a developer might encounter, e.g. PX4 does not (currently) use the entire NuttX build system, it only uses NuttX to build some pieces/libraries and does the final linking with its own cMake files

Could not agree more. The current docs assume you already know how the bits fall together. 

> If you'd like to use a back-up coprocessor (e.g. the STM32F103 on the PixHawk), where does that fit into the whole build process, and what are its requirements? 

Yes. So that to me is part of that "what can you do/what should you do/what options are there" discussion that is currently missing. 

Thanks again.

en/debug/porting-guide.md
joshuawhitehead: Will do.  This was copy-paste from the current version of the porting guide, so the target may have changed since that was originally written.

hamishwillee (approved):

hamishwillee:
Thanks @joshuawhitehead - a clear improvement over previous text.

PR #504, opened 2018-04-08
Add how to use simulation tools on Windows

xuhao1:
Shared how to use simulation tools on Windows.


hamishwillee:
Thanks @xuhao1 This is very cool, but it won't work out of the box with the installation script provided, because that does not include jmavsim dependencies. Can you provide the update to windows_bash_nuttx.sh (see https://dev.px4.io/en/setup/dev_env_windows.html#bash-on-windows-new ) that you used to additionally install jMAVSim (and also separately .... Gazebo)?

EDIT: IGNORE THE ABOVE - I see it does work :-)

hamishwillee:
@MaEtUgR @LorenzMeier FYI, this shows how to use Ubuntu Bash for Windows with XMing to provide JMAVSim UI. It works out of the box for JMAVSim. 
1. The frame rate is same as in VM so nowhere near as good as @MaEtUgR environment
2. QGC 3.3.3 connected but couldn't be used to drive it due to some sort of parameter problem (have you seen this before?):
    image

@xuhao1 Can you provide the setup instructions you used for installing Gazebo dependencies. In the meantime I might merge this without the Gazebo bits.

xuhao1:
My qgroundcontrol works very well. 
I have use it to upload waypoints and also arm/disarm and also start  and end a mission.
@hamishwillee your issue seems like something wrong with your local loop network, are you using the latest updated Windows?
As for gazebo, I found  ubuntu_sim_ros_gazebo.sh or install ros-kinect-desktop-full manual can just works fun.
(I updated my gazebo version to 9.0 so that don't work before, gazebo 8.4 in Ubuntu repo works.)
But gazebo works really really slow on WSL on my Surface Book 2. I don't know its speed on Linux.

hamishwillee:
Thanks @xuhao1 . I will try see if I can get QGC working properly and get back to you. Given that jMAVSim is fairly slow via this mechanism I would expect Gazebo to be slow too. I am tempted to simply say that the method works but is not particularly recommended. Hope to try this tomorrow, but still have a few tasks to do.

MaEtUgR:
@xuhao1 Interesting to see! I should add all my instructions and the download link to the setup as well.
Just a small comment: The **second screenshot** shows some personal information of your computer, don't you want to replace **crop** it a bit in your own interest?

hamishwillee:
Thanks @xuhao1 - excellent contribution. I have merged this to make it easier to update. I have tried gazebo and to my mind it is not usable so I will tidy that section and cross-link.

hamishwillee:
FYI, I tidied up the structure and instructions in https://github.com/PX4/Devguide/pull/506

This removes the "detail" for Gazebo (including image) since my test on a fairly performant computer indicates that it isn't really a usable solution. I still say that it is possible.

PR #510, opened 2018-04-16
Tidy up EKF warnings v2

hamishwillee:
This tidies up the doc so that:
1. Descriptions of parameters in text are linked to "official" documentation (duplicated content removed)
1. Bullets are indented 2 spaces. This merely tidies the rendering in github - things always worked in final gitbook version
1. Headings have space between text and heading marker (just improves rendering).
1. Added definition for PREFLIGHT FAIL: EKF INTERNAL CHECKS from @priseborough

### DIFF
--- a/en/tutorials/pre_flight_checks.md
+++ b/en/tutorials/pre_flight_checks.md
@@ -1,59 +1,70 @@
 # Preflight Sensor and EKF Checks
-The commander module performs a number of preflight sensor quality and EKF checks which are controlled by the COM_ARM<> parameters. If these checks fail, the motors are prevented from arming and the following error messages are produced:
+
+The commander module performs a number of preflight sensor quality and EKF checks which are controlled by the [COM_ARM_](../advanced/parameter_reference.md#commander) parameters. If these checks fail, the motors are prevented from arming and the following error messages are produced:
 
 * PREFLIGHT FAIL: EKF HGT ERROR
- * This error is produced when the IMU and height measurement data are inconsistent.
- * Perform an accel and gyro calibration and restart the vehicle. If the error persists, check the height sensor data for problems.
- * The check is controlled by the COM_ARM_EKF_HGT parameter.
+  * This error is produced when the IMU and height measurement data are inconsistent.
+  * Perform an accel and gyro calibration and restart the vehicle. If the error persists, check the height sensor data for problems.
+  * The check is controlled by the [COM_ARM_EKF_HGT](#COM_ARM_EKF_HGT) parameter.
 * PREFLIGHT FAIL: EKF VEL ERROR
- * This error is produced when the IMU and GPS velocity measurement data are inconsistent. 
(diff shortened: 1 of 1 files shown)


hamishwillee:
@priseborough This is updated docs following your feedback in https://github.com/PX4/Devguide/pull/500

1. After fixing up links in top section to point to official docs (and deleting duplicates) I was left with these two that are not linked above: COM_ARM_WO_GPS, COM_ARM_EKF_YAW - should these be mentioned in the descriptions, or removed as not relevant?
1. I will move this to the user guide after we finish the discussion. My thought is that all I need to change is the first line.
   From
  > The commander module performs a number of preflight sensor quality and EKF checks which are controlled by the COMARM parameters. 

   TO 

   > PX4 performs a number of preflight sensor quality and EKF checks to determine if there is a good enough position fix to ARM/Fly. These checks are controlled by the COM/ARM parameters. 

Re

3. >The problem is that with the emergence of multiple types of external aiding, the concept of GPs lock is less important than it used to be and the ekf can have a useful nav solution without a GPS lock or a un-usable one with GPS lock. We need to start using the concept of navigation quality rather than GPS lock. A common complaint i see is that the user sees that the GPs has a 3D fix, but the lights appear to be saying there is no GPS lock.

That is a problem, but does it need to affect this document? My thought here is that we should make the light reflect the position lock (I say position lock to mean "navigation quality" - is there a better way of referring to the state "good enough to arm/takeoff" than "position lock"?

hamishwillee:
@priseborough  Ping!

priseborough:
It's just semantics, but how about using the term 'position estimate' instead of position fix'? Historically, use of the word 'fix' tends to be associated with GPS. I'm not hard over on it, just a suggestion.

hamishwillee:
Thanks. So what about this part @priseborough ...

After fixing up links in top section to point to official docs (and deleting duplicates) I was left with these two that are not linked above: COM_ARM_WO_GPS, COM_ARM_EKF_YAW - should these be mentioned in the descriptions, or removed as not relevant?

priseborough:
COM_ARM_EKF_YAW determines the maximum difference (in radians) between the navigation yaw angle and magnetic yaw angle allowed before preflight checks fail. It can fail if the yaw gyro has a large offset or if the vehicle is moved or rotated in the presence of a bad magnetic interference or magnetometer calibration.

According to the documentation COM_ARM_WO_GPS can be used to prevent preflight checks passing until GPS data is present. This is not very specific, but inspection of the code shows that it actually controls which EKF checks run so I will check further.

priseborough:
I've checked the code and setting COM_ARM_WO_GPS = 0 (instead of the default value of 1) has the following effects:

1) The EKF must be providing a global position estimate before the vehicle can be armed.
2) The EKF GPS quality checks must be passing before the vehicle can be armed.

hamishwillee:
Thanks Paul. Further minor changes in https://github.com/PX4/px4_user_guide/pull/210

PR #513, opened 2018-04-20
clarification added for adding airframe in QGC UI

barzanisar:


hamishwillee:
Thanks @barzanisar . THis is good. I merged and did some tidy up in #514. :FYI - what I did/do
1. I use bold to mark up button presses - e.g. **OK**. Improves readability.
1. If things need to be done in order then use numbered lists
1. If someone has to press a button and showing it in a screenshot, I tend to circle the button in red - easy to add and makes button easier to see. 
1. Images should be indented to the same point as text if they belong to a bullet.

PR #548, opened 2018-06-23
jenkins: use same agent for build and deploy

lamping7:
This follows the same logic as https://github.com/PX4/px4_user_guide/pull/275

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,5 +1,9 @@
 pipeline {
-  agent any
+  agent {
+    docker {
+      image 'px4io/px4-docs:2018-06-14'
+    }
+  }
   stages {
 
     stage('Build') {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -1,52 +1,47 @@
...
           reportName: 'PX4 Dev Guide'
-        ]
+        ])
       }
 
     } // Build
 
     stage('Deploy') {
       environment {
+        GIT_AUTHOR_EMAIL = "bot@pixhawk.org"
```
lamping7, line 36: This is needed because there is no checkout stage, to prevent git errors not knowing who the author is.

Jenkinsfile
```diff
@@ -1,52 +1,47 @@
...
-          image 'px4io/px4-docs:2018-06-14'
-        }
-      }
-
       steps {
-        sh 'export'
-        sh 'gitbook install'
-        sh 'gitbook build'
-        stash includes: '_book/', name: 'gitbook'
+        sh('export')
```
lamping7, line 15: using parens was added to be consistent

dagar (approved):

PR #551, opened 2018-07-01
Highlight dev targets

hamishwillee:
This better highlights the dev targets you can build with MacOS - improve type of issues raised by https://github.com/PX4/Firmware/issues/9787

### DIFF
--- a/en/setup/dev_env_mac.md
+++ b/en/setup/dev_env_mac.md
@@ -1,6 +1,11 @@
 # Development Environment on Mac
 
-Mac OS X is the main development platform for PX4. The following instructions explain how to set up a development environment for building NuttX-based hardware (Pixhawk, etc.) and Simulation (jMAVSim/Gazebo8) targets. For other targets see: [Toolchain Installation > Supported Targets](../setup/dev_env.md#supported-targets).
+Mac OS X is the main development platform for PX4. The following instructions set up a development environment for building:
+* NuttX-based hardware (Pixhawk, etc.)
+* jMAVSim Smulation
+* Gazebo 8 Simulation
+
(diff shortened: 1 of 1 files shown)


en/setup/dev_env_mac.md
```diff
@@ -1,6 +1,11 @@
 # Development Environment on Mac
 
-Mac OS X is the main development platform for PX4. The following instructions explain how to set up a development environment for building NuttX-based hardware (Pixhawk, etc.) and Simulation (jMAVSim/Gazebo8) targets. For other targets see: [Toolchain Installation > Supported Targets](../setup/dev_env.md#supported-targets).
+Mac OS X is the main development platform for PX4. The following instructions set up a development environment for building:
```
ndepal, line 3: > Mac OS X is the main development platform for PX4

This seems a bit contradictory to the statement on the "Toolchain Installation" page:
> We recommend Ubuntu Linux LTS edition as this enables building all PX4 targets, and using most simulators and ROS.

What makes Mac the main dev platform / what is the reader supposed to gather from this statement?

hamishwillee: Yes, I'm aware. The statement predates me. I'm not sure on what it was based. @bkueng What would you say defines the "main dev platform". I've always though Linux makes sense as a free OS that builds more targets, but you could argue that MacOS fits under criteria of "easiest setup".

bkueng: Well, I cannot give you an unbiased answer :)
I also say it's Linux, since as you say it's free, builds all targets and is best maintained.

hamishwillee: Fair enough. I have fixed this in https://github.com/PX4/Devguide/pull/554 - we have no "main" dev env, just ones that we support.

PR #559, opened 2018-07-19
Add info about singlewire support

hamishwillee:


en/hardware/porting_guide.md
```diff
@@ -72,3 +72,13 @@ This section describes the various middleware components, and the configuration
...
+
+
+## RC UART Wiring Recommendations (Singlewire)
+
+Boards using STMF7 (or other microcontrollers that support singlewire) should connect UARTs using the "normal" connector wiring - i.e. TX (OUT) of the F7 to RX (IN) of device and RX (IN) on the F7 to TX (OUT) of that device. 
+PX4 will ensure that the hardware sets the single wire on the SoC's RX pin to prevent any contention. 
+
+If the STMF7 RC UART RX & TX are connected together, *singlewire mode* must to be used.
+
+For more information, see [px4fmu-v5/board_config.h](https://github.com/PX4/Firmware/blob/master/src/drivers/boards/px4fmu-v5/board_config.h)
```
hamishwillee: @bkueng This is the only interesting/relevant change in this PR.

I have tried to word it as you suggested. Can't help thinking I should have just created an issue and assigned it to you :-)

ANyway, this last paragraph concerns me. The link to `px4fmu-v5/board_config.h` isn't particularly helpful if you go there - ie not as useful as stating that you have to (for example) call method "X" in drivers that use UART "Y". Is this link OK or were you suggesting something else?

bkueng: How about this:

It is generally recommended to connect RC via separate RX and TX pins to the microcontroller. If however RX and TX are connected together, the UART has to be put into singlewire mode to prevent any contention. This is done via board config and manifest files. One example is [px4fmu-v5](https://github.com/PX4/Firmware/blob/master/src/drivers/boards/px4fmu-v5/manifest.c).

> ie not as useful as stating that you have to (for example) call method "X" in drivers that use UART "Y". Is this link OK or were you suggesting something else?

Because there is nothing that you need to call. It is one level deeper than the drivers, you only configure things and provide board-specific implementations for certain functionality.

hamishwillee: @bkueng - I had completely missed that this was below driver level. Updated as suggested and merged. Thanks.

PR #560, opened 2018-07-19
system_console: add link to mavlink shell in QGC

### LINKED ISSUE #557: Hint to QGC console is missing in console guide
@bkueng Could you do a quick pass?

https://dev.px4.io/en/debug/system_console.html

bkueng:
Fixes  #557

### DIFF
--- a/en/debug/system_console.md
+++ b/en/debug/system_console.md
@@ -2,16 +2,21 @@
 
 The system console allows low-level access to the system, debug output and analysis of the system boot process. The most convenient way to connect it is by using a [Dronecode probe](http://nicadrone.com/index.php?id_product=61&controller=product), but a plain FTDI cable can be used as well.
 
-## System Console vs. Shell
+## System Console vs. Shells
 
-There are multiple shells, but only one console: The system console is the location where all boot output (and applications auto-started on boot) is printed.
+There is just one *System Console*, which runs on one specific UART (the debug port, as configured in NuttX), and is commonly attached via FTDI cable.
+- Used for *low-level debugging/development*: bootup, NuttX, startup scripts, board bringup, development on central parts of PX4 (e.g. uORB).
+- In particular, is the only place where all boot output (including information about applications auto-started on boot) is printed.
 
-  * System console (first shell): Hardware serial port
-  * Additional shells: Pixhawk on USB (e.g. lists as /dev/tty.usbmodem1 on Mac OS)
+Shells provide higher-level access to the system:
+- Used for basic module testing/running commands.
+- Only display the output of modules you start (and therefore cannot debug the boot process).
+- Cannot display the output of tasks running on the work queue.
 
-> **Info**
-> USB shell: To just run a few quick commands or test an application connecting to the USB
-> shell is sufficient. The MAVLink shell can be used for this, see [below](#mavlink_shell).
+There are several shells:
+* USB shell (e.g. Pixhawk on USB lists as `/dev/tty.usbmodem1` on Mac OS)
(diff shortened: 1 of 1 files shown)


en/debug/system_console.md
```diff
@@ -11,9 +11,7 @@ There are multiple shells, but only one console: The system console is the locat
 
 > **Info**
 > USB shell: To just run a few quick commands or test an application connecting to the USB
-> shell is sufficient. The Mavlink shell can be used for this, see below.
-> The hardware serial console is only needed for boot debugging or when USB should be used
-> for MAVLink to connect a [GCS](../qgc/README.md).
+> shell is sufficient. The Mavlink shell can be used for this, see [below](#mavlink_shell).
```
hamishwillee: MAVLink

bkueng: Done

hamishwillee: It predates your change, but this tip, and indeed the whole section here (System Console vs. Shell) is unclear.

For the tip,it says "The Mavlink shell can be used for this", but "this" is not defined. I think you are saying that the MAVLink shell can also be used for quick tests?

For the system console vs shell thing I think this is saying that a console is a type of output that is accessed through a shell. You can get this from several different shells - USB, MAVlink whatever. It is also called the NSH console. Is that correct?

If so, would bullet point all the consoles, including MAVLink, then have info at the end explaining what they are good for.

bkueng: > I think you are saying that the MAVLink shell can also be used for quick tests?

Correct.

There are a few differences between the system console and the shell.
System console:
- runs on one specific UART (debug port, as configured in NuttX), used for deeper debugging/devel (bootup, NuttX, startup scripts, board bringup, development on central parts of PX4, such as uORB).
- commonly attached via FTDI cable
- you see the boot output (and all the auto-started tasks)

Whereas the shell is more high-level and can also be used through MAVLink (when there was no mavlink shell yet, it was used directly via USB for example). You'll see the output of what you started, but not of what's already running (but you won't see the output of a task running on the work queue).

Do you want to make a suggestion, so it's understandable for you?

hamishwillee: I'll commit a change for you to check.

hamishwillee: Done!

en/debug/system_console.md
```diff
@@ -2,18 +2,21 @@
...
 
-  * System console (first shell): Hardware serial port
-  * Additional shells: Pixhawk on USB (e.g. lists as /dev/tty.usbmodem1 on Mac OS)
+Shells provide higher-level access to the system:
+- Used for basic module testing/running commands.
+- Only display the output of modules you start (and therefore cannot debug the boot process).
+- Cannot display the output of tasks running on the work queue.
+
+There are several shells:
+* USB shell (e.g. Pixhawk on USB lists as `/dev/tty.usbmodem1` on Mac OS)
```
bkueng: This also runs mavlink now. We can write that the shell can be used on any UART (including USB), but running mavlink is more powerful and thus we only use the shell via mavlink today.

bkueng (commented):
Thanks @hamishwillee

en/debug/system_console.md
hamishwillee: Can you update appropriately? What is wrong with this (as it is now) is that is isn't "really" clear how you would use it - ie I know how to connect up to use with QGC, but how do I use the USB option?

bkueng: Done

hamishwillee:
Thanks @bkueng ! Merged.

PR #565, opened 2018-07-27
Tidy up Windows PX4 toolchain support

hamishwillee:
- Remove warnings about Windows development not being supported.
- Highlight Cygwin solution is best at top level and in the other solutions
- Restructure Cygwin to make the main installation path direct - ie push features and the other installation options under sub heading. Otherwise this doc looks scary.


hamishwillee:
@MaEtUgR This is FYI, changes I made to make the Cygwin docs more user friendly. I will merge, but hope you are OK with them.

One question, is whether the repo should move to PX4 org.

MaEtUgR:
Perfect! I can always change if I see the need but generally you provide a lot more reasonable solutions than I could come up with. Thanks a lot @hamishwillee !

PR #576, opened 2018-08-09
RPi: update PX4 start instructions

bkueng:
Needed after https://github.com/PX4/Firmware/pull/10173 was merged.

### DIFF
--- a/en/setup/building_px4.md
+++ b/en/setup/building_px4.md
@@ -172,7 +172,7 @@ make posix_rpi_cross upload # for cross-compiler build
 Then, connect over ssh and run it with (as root):
 
 ```sh
-sudo ./px4 px4.config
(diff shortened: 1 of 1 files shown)


hamishwillee:
Looks good. Do we need to keep "retrospective" information - ie for people who might want to use older builds? If not, this is good/approved.

bkueng:
I thought that the dev-guide information is generally in sync with master. But I'm fine with adding a note if you wish.
I would advise someone using RPi to stick to master as we now have the scripting support and thus can add the same functionality as on Pixhawks.
I'm going ahead and merge it then.

hamishwillee:
> I thought that the dev-guide information is generally in sync with master. But I'm fine with adding a note if you wish.

You are correct, so up to you. I wish we did have time and energy to version some of this stuff though.

PR #589, opened 2018-08-27
Add JMAVSim multi-vehicle simulation

### LINKED ISSUE #585: Multi-Vehicle Simulation with jmavsim
Add the new multi-vehicle JMAVSim solution.
Possibly make this one a "parent": [Multi-Vehicle Simulation · PX4 Developer Guide](http://dev.px4.io/en/simulation/multi-vehicle-simulation.html).

The info reproduced from [here](https://github.com/Dronecode/DronecodeSDK/issues/494#issuecomment-413602890):

---

Multiple jMAVSim vehicles (with latest master):

Build PX4
```
make posix_sitl_default
```
Start multiple instances

```
Tools/sitl_multiple_run.sh 2
```

Then start the first instance:
```
 ./Tools/jmavsim_run.sh
```
Start subsequent instances, specifying the UDP port (expected to be 14560+i for i in [0, N-1]
```
./Tools/jmavsim_run.sh -p 14561
```

To have two drones connect on the same port, do this change:
```
diff --git a/ROMFS/px4fmu_common/init.d-posix/rcS b/ROMFS/px4fmu_common/init.d-posix/rcS
index 91c9cb7c8f..f276357d90 100644
--- a/ROMFS/px4fmu_common/init.d-posix/rcS
+++ b/ROMFS/px4fmu_common/init.d-posix/rcS
@@ -83,7 +83,7 @@ fi
 param set MAV_SYS_ID $((1+px4_instance))
 simulator_udp_port=$((14560+px4_instance))
 udp_offboard_port_local=$((14557+px4_instance))
-udp_offboard_port_remote=$((14540+px4_instance))
+udp_offboard_port_remote=$((14540))
 udp_gcs_port_local=$((14556+px4_instance))
 
 if [ $AUTOCNF == yes ]
```

hamishwillee:
Fixes #585

Note, I have not tested this as it currently can't be done on Windows (due to bugs in the Cygwin and VM-based simulation). I don't have time to do so on Linux yet.


en/simulation/multi_vehicle_jmavsim.md
```diff
@@ -0,0 +1,52 @@
...
+   ```
+   ./Tools/jmavsim_run.sh
+  ```
+1. Start subsequent instances, specifying the UDP port for the instance:
+   ```
+   ./Tools/jmavsim_run.sh -p 14561
+   ```
+   The port should be set to `14560+i` for `i` in `[0, N-1]`.
+
+## Instances on Same Port
```
bkueng: I'm not sure if we should add this. It's generally not something that you would want, and should there be a need for it, we can add a simpler way to achieve this.

en/simulation/multi_vehicle_jmavsim.md
```diff
@@ -0,0 +1,52 @@
+# Multi-Vehicle Simulation with JMAVSim
+
+This topic explains how to simulate multiple UAV (multicopter) vehicles using JMAVSim and SITL.
```
bkueng, line 3: Should we mention that this runs multiple independent simulators and all vehicles start at the same position? Meaning if you want to do swarm simulation (or similar) you're better off using gazebo.

bkueng (commented):
I have 2 remarks to discuss, otherwise looks good.

en/simulation/multi_vehicle_jmavsim.md
hamishwillee: Thanks @bkueng .

Can you clarify a bit more on why Gazebo is better for swarm? I ASSUME it is because 
* You can choose different vehicle start locations (?)
* The simulation runs a single instance of Gazebo with multiple SITL? (I presume you are saying that for JMAVSim `runs multiple independent simulators` means that you're running multiple SITL and multiple JMAVSim, and that therefore this is more increasingly resource consuming than Gazebo as you scale up.

Is that correct?

So presumably this is for when you have a small number of device - e.g. a 3-device survey and you don't care about the starting position - right?
At what number of instances would you guess that Gazebo becomes more reasonable?

en/simulation/multi_vehicle_jmavsim.md
hamishwillee: We should find a method for this ideally. The use case is multiple devices attached to a real Wifi network. The test case is Dronecode SDK, which in theory supports this.

en/simulation/multi_vehicle_jmavsim.md
bkueng: > You can choose different vehicle start locations

correct

> The simulation runs a single instance of Gazebo with multiple SITL

correct

> and that therefore this is more increasingly resource consuming than Gazebo as you scale up.

Yes that too. But also in terms of interactions: crashing into another drone, avoidance of other drones, etc.

> At what number of instances would you guess that Gazebo becomes more reasonable?

It's more about the use-case. If you want to test the SDK, or multi-vehicle support in QGC, you're probably fine with only jMAVSim. But if you want to do vision, swarm simulations with many vehicles or similar, gazebo is better.

hamishwillee:
Thanks @bkueng . I've updated as you suggested - OK now?

en/simulation/multi_vehicle_jmavsim.md
```diff
@@ -0,0 +1,52 @@
...
+   ```
+   make posix_sitl_default
+   ```
+1. Run **sitl_multiple_run.sh**, specifying the number of instances to start (e.g. 2):
+   ```
+   ./Tools/sitl_multiple_run.sh 2
+   ```
+1. Start the first instance:
+   ```
+   ./Tools/jmavsim_run.sh
```
hamishwillee, line 28: @bkueng According to http://dev.px4.io/en/simulation/#default-px4-mavlink-udp-ports port 14560 is used for connecting PX4 to simulators - ie the SITL / jMAVSim connection. 

Just to confirm:
1. Is the default port (for first instance) 14560?
2. When you set to 14561 is this the simulator port, or the port to which you connect a GCS?

bkueng: > Is the default port (for first instance) 14560?

Yes

> When you set to 14561 is this the simulator port, or the port to which you connect a GCS?

Simulator port

hamishwillee: So what does a GCS connect on then? ie if you set port 14561 , does QGC try to connect to 14541 or something else?

bkueng: All run on 14540 as usual.

hamishwillee: Thank you. FYI I updated doc with this information in https://github.com/PX4/Devguide/commit/360a30ecba13069349fbba97bfc2b1058e8da44f

PR #591, opened 2018-08-28
mc_03_auto_manual.mix.md

Tony3dr:
@hamishwillee @mrpollo let me know what you think.

### DIFF
--- a/en/test_cards/MC_03 - Auto Manual mix
+++ b/en/test_cards/MC_03 - Auto Manual mix
@@ -1,3 +1,5 @@
+MC_03 - Auto Manual mix
+
 ##Create and upload mission
 
 ❏ Mission Criteria
--- a/en/test_cards/MC_03 - Auto Manual mix
+++ b/en/test_cards/MC_03 - Auto Manual mix
@@ -1,0 +1,37 @@
+Create and upload mission
+
+❏ Mission Criteria
+
+    ❏ Changes in Altitude throughout the mission
+
+    ❏ Mission should end in the air and NOT Land/RTL
+
+    ❏ Duration of 3 to 4 minutes
+
+❏ Upload mission to vehicle using QGroundControl
+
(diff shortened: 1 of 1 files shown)


en/test_cards/MC_03 - Auto Manual mix
```diff
@@ -0,0 +1,37 @@
+Create and upload mission
+
+❏ Mission Criteria
```
mrpollo: @hamishwillee we can also use real read-only checkboxes ere, it would look like this

```
- [ ] Mission Criteria
```

- [ ]  Mission Criteria

What do you think? @hamishwillee

hamishwillee: That doesn't work on gitbook - they will just render as `[ ] Stabilized`. It is a perfectly reasonable change, but I kind of like the boxes.

hamishwillee:
Changes please:
1. Rename file to mc_03_auto_manual.mix.md . The bit that matters here is the file extension **.md** because that tells both github and gitbook how the file should be interpretted. I also like the all lower case underscore separated format, but that is taste.
1. You will need to add the new file to /en/SUMMARY.md below the other ones (and following the same pattern). If you don't do this then gitbook will not know where to render the file in the sidebar.
1. Please add link to this in en/test_and_ci/test_flights.md as I did in #590 - this provides a better link in the body of the parent doc.
1. The new doc lacks markup for headings that you should be able to see in #590. Specifically:
   - At top, put a heading:
     ```
     # Test MC_03 - Auto Manual mix 
     ```
   - Precede each of the following lines with `##` (second level heading): Create and Upload Mission, Flight, Expected Results. Note I have first-letter capitalised.
1. In Expected results, having the lines like that means they will just render as one paragraph. In front of each point (beginning of line) put a `*` to make them render as bullets.

That's it :-)

PR #593, opened 2018-08-30
Add info about how to do in-source docs

hamishwillee:
This follows on from https://github.com/PX4/Firmware/pull/10127

@bkueng @mcsauder - appreciate your review.


en/apps/module_template.md
```diff
@@ -14,7 +14,7 @@ The template demonstrates the following additional features/aspects that are req
   [startup script](../advanced/system_startup.md).
 - Command-line argument parsing.
 - Documentation: the `PRINT_MODULE_*` methods serve two purposes (the API is
-  documented [in the source code](https://github.com/PX4/Firmware/blob/master/src/platforms/px4_module.h#L387)):
+  documented [in the source code](https://github.com/PX4/Firmware/blob/v1.8.0/src/platforms/px4_module.h#L381)):
```
hamishwillee, line 17: FYI only, I fixed the reference because the docs moved in master.

en/middleware/modules_main.md
```diff
@@ -1,20 +1,16 @@
...
-The following pages document the PX4 modules, drivers and commands. They
-describe the provided functionality, high-level implementation overview and how
-to use the command-line interface.
 
-> **Note** **This is auto-generated from the source code** and contains the
-> most recent modules documentation.
+The following pages document the PX4 modules, drivers and commands. 
+They describe the provided functionality, high-level implementation overview and how to use the command-line interface.
+
+> **Note** **This is auto-generated from the source code** and contains the most recent modules documentation.
```
hamishwillee, line 6: FYI, main change here is to point to the docs API. I think that is useful.

en/contribute/README.md
```diff
@@ -112,6 +112,24 @@ PX4 uses [astyle](http://astyle.sourceforge.net/) for code formatting. Valid ver
 
 Once installed, formatting can be checked with `./Tools/astyle/check_code_style_all.sh`. The output should be `Format checks passed` on a clean master. If that worked, `make format` can be used in the future to check and format all files automatically.
 
+### In-Source Documentation
```
hamishwillee, line 115: This bit is what needs review.

en/contribute/README.md
```diff
@@ -112,6 +112,24 @@ PX4 uses [astyle](http://astyle.sourceforge.net/) for code formatting. Valid ver
...
+
+> **Note** Source-code documentation standards are not enforced, and the code is currently inconsistently documented.
+  We'd like to do better!
+  
+Currently we have two types of source-based documentation:
+- `PRINT_MODULE_*` methods are used for both module run time usage instructions and this guide.
+  - The API is documented [in the source code here](https://github.com/PX4/Firmware/blob/v1.8.0/src/platforms/px4_module.h#L381).
+  - The docs are published in [Modules & Commands Reference](../middleware/modules_main.md). 
+  - Good examples of usage include the [Application/Module Template](../apps/module_template.md) and the files linked from the modules reference.
+* [Doxgyen](http://www.doxygen.nl/) commenting is [published here](https://px4.github.io/Firmware-Doxygen/index.html)
```
bkueng: Not sure if we should link to `https://px4.github.io/Firmware-Doxygen`

en/contribute/README.md
```diff
@@ -112,6 +112,24 @@ PX4 uses [astyle](http://astyle.sourceforge.net/) for code formatting. Valid ver
...
+  - A good example of usage is [src/modules/events/send_event.h](https://github.com/PX4/Firmware/blob/master/src/modules/events/send_event.h), which is [published here](https://px4.github.io/Firmware-Doxygen/d4/dc6/structevents_1_1_send_event.html)
+  - We encourage documentation of classes, files (for multi-class files), methods/functions, members etc., 
+    using tags: `@class`, `@file`, `@param`, `@return`, `@brief`, `@var`, `@see`, `@note`.
+  - Do not document default constructors or destructors.
```
bkueng: I'd like to see it written as something like: We encourage documentation of... where it adds value, for example variables for which the purpose is not obivous from the name (For example from `orb_advert_t _controller_status_pub` it's already clear that it's the `controller_status` uORB topic publication). Documentation should not be redundant with the code.
Which then means it can make sense to document a default constructor. For example for a class that accesses a UART device it could read: 'Constructor that initializes the UART device based on a board-specific default device path'.

Also corner cases and error handling of an API should generally be documented.

hamishwillee: @bkueng There is a school of thought that you should document everything to ensure that a)you miss nothing b) you think about documentation. As you clearly do not agree with that :-) I have changed as you suggested. Can you please re-check?

en/contribute/README.md
hamishwillee: I deleted for now. I do think that if we are able to improve our docs we should eventually point to this.

bkueng (approved):
Thanks for the update, looks good from my side.

en/contribute/README.md
bkueng: Yup, I do not think that such a rule leads to good quality.

en/contribute/README.md
mcsauder: I apologize that I am late on this, but this all looks great.  Tested on a new repo clone:
image

mcsauder: Thanks for your work here @hamishwillee

en/middleware/modules_main.md
mcsauder: Agreed.

en/contribute/README.md
hamishwillee: Thanks @mcsauder for review. Now we all just need to start gently suggesting this when we see docs that could use some additional information.

PR #602, opened 2018-09-06
Translated config_initial.md into Korean

ghost:
The existing Korean file seems an old version. The content was updated and translated. Finished.

### DIFF
--- a/kr/setup/config_initial.md
+++ b/kr/setup/config_initial.md
@@ -1,21 +1,30 @@
-# 초기 설정
+# 초기 설정 및 구성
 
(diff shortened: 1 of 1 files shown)


hamishwillee:
Thanks @konjoon . Thank you. I will merge.

Note however that we have moved to the crowdin translation service, which will make translation much easier. We'd love your help with this, as we do not have many Korean translators. To join, you would create a login on https://crowdin.com/ and then use https://crowdin.com/project/px4-developer-guide/invite?d=d5j626j6b51733e307p47313e3k4l4 to join this project and start translating.

PR #605, opened 2018-09-11
Add pwm escs and servo from old pixhawk.org site

hamishwillee:
@bkueng This adds the *PWM Servos and Motor Controllers* doc from old pixhawk.org. The current changes are very minor.
In the porting doc you wrote "Beat: RC manufacturers would probably not use the information from here (we would need to provide an in-depth technical document)"

I am not sure if this is the right place, or if bits can be split out.  It is far more about hardware, so it really should be on pixhawk.org ... but there is no place for it there. 

Would appreciate your advice on what should be kep, and what might be added to https://docs.px4.io/en/peripherals/esc_motors.html
We don't say anything about connecting the motors - is that reasonable? Without knowing anything about this, it sounds to me like the information on connecting up and powering ESCs and servos should be captured. 

Yes, I am effectively dumping the problem to you :-). This PR is to make sure that the info does not get lost before we can make a decision !

Thoughts?


en/hardware/pwm_escs_and_servo.md
```diff
@@ -0,0 +1,132 @@
...
+
+## ESC Power Connection
+
+Each PWM Electronic Speed Controller (ESC) has at least five main wires:
+- Power VBAT (usually thick and red)
+- Power GND (usually thick and black)
+
+And on the servo plug:
+- PWM signal (usually white or yellow)
+- +5V (usually red or orange)
```
bkueng: Not all ESCs have the 5V connection

en/hardware/pwm_escs_and_servo.md
```diff
@@ -0,0 +1,132 @@
...
+- GND (usually black or brown)
+
+> **Tip** Always connect Power VBAT and GND as well as the PWM signal and GND from the servo plug. 
+There is **no setup** that does not require signal ground to be connected!
+
+The +5V line on the servo plug has a special purpose, and how it should be connected very much depends on the particular ESC in use:
+
+- On a normal fixed wing ESC (which are often abused for multicopters), it provides the output of a BEC to power servos. Its safe to connect to Pixhawk.
```
bkueng: I think this these can go to the user-guide as it documents how to attach an ESC.

en/hardware/pwm_escs_and_servo.md
bkueng: Not all ESCs have a 5V wire

en/hardware/pwm_escs_and_servo.md
bkueng: These could go to the user-guide, as they describe how to attach an ESC

en/hardware/pwm_escs_and_servo.md
```diff
@@ -0,0 +1,132 @@
...
+> **Tip** Instead of cutting the +5V line you can gently lift of the locking tab of the plastic housing of the servo connector for that pin (e.g. using a cutter blade or small screw driver) and pull the pin out. 
+Isolate it with electrical isolation tape and tape it to the servo cable. 
+This allows to undo this later easily.
+
+## PWM Values
+
+The PWM value sent out depends on the flight stack and configuration used. 
+Most ESCs must see a low value before they will spin up a motor to higher values (to protects user who have the throttle stick in the middle position on power-up). 
+This low value barrier varies between roughly 1200 and 900 us. 
+The native stack therefore sends a value of 900 while being disarmed to ensure the ESC sees a valid low pulse.
```
bkueng: You can reference the PWM_DISARMED params

en/hardware/pwm_escs_and_servo.md
```diff
@@ -0,0 +1,132 @@
...
+PWM interfaces are not formally standardised, however, the utilised micro controllers use TTL or CMOS voltage levels. 
+TTL is defined as low < 0.8V and high > 2.0V with some manufacturers using > 2.4V for additional noise margin. 
+CMOS logic is defined with similar voltage levels. 
+5V levels are **never** required to successfully switch to an *on* state.
+
+> **Tip** Futaba, FrSky and Spektrum receivers output 3.3V or 3.0V voltage levels, as they are well above 2.4V. 
+  Pixhawk has adopted this common industry pattern and outputs 3.3V levels on recent boards.
+
+Note that on a normal quad rotor motors spin at 100-130 Hz. 
+This means that if the bus operates at 400 Hz the motor receives new commands three to four times during a single revolution, but it is physically impossible to change the blade speed (due to electric effects and blade inertia) that fast.
```
bkueng: This and the next sections can be dropped completely. While this is correct (and the next section partially), it does not really help, because what is relevant is the latency/delay. And a faster output rate helps reducing the latency (that's at least my understanding, and I've also seen this in practice).

en/hardware/pwm_escs_and_servo.md
```diff
@@ -0,0 +1,132 @@
...
+The image below shows how noisy the signal becomes if GND is not connected.
+
+![PWM without ground](../../assets/hardware/pwm_esc/pwm_without_gnd.png)
+
+
+## Signal Shape Comparison
+
+As the scope shots below show the Pixhawk signal arrives faster (at 400 Hz when configured for multicopters instead of 50-60 Hz), but the overall signal shape and voltage levels are exactly the same as for common RC receivers.
+
+### FrSky D4R-II
```
bkueng: I'm not sure if we need these comparisons. It might have been needed in the early days to show that a Pixhawk is compatible to the output of the RC receivers (where intelligent autopilots were not common yet and RC receivers were directly attached to an ESC to fly a plane fully manual - which is still the case today but we would need to explain that here), but today the signals are established and we don't need it anymore.

en/hardware/pwm_escs_and_servo.md
```diff
@@ -0,0 +1,132 @@
...
+### Spektrum AR 10000
+
+3.3V levels, 45 Hz
+
+
+![spektrum_ar_10000](../../assets/hardware/pwm_esc/spektrum_ar_10000.png)
+
+### Pixhawk (PX4FMU v2.4.3)
+
+3.3V levels, 400 Hz (**400 Hz ONLY if configured as multicopter, default is 50 Hz**)
```
bkueng: If you want to mention the 400Hz, we also need to mention OneShot, which runs in sync with the rate controller (currently 250Hz, but should soon be 1kHz).

en/hardware/pwm_escs_and_servo.md
```diff
@@ -0,0 +1,132 @@
...
+This is because for every positive pulse (the ESC signal) there needs to be an adjacent ground return path for a clean signal shape.
+
+### Power Connection / Opto-isolated ESCs
+
+Please see the first section of this page for a complete explanation on power connection considerations. 
+If you are using an opto-isolated ESC not providing a BEC / power output, please ensure that the ESC does not need its +5V line powered for the opto-isolator.
+
+### Missing Valid Minimum Value
+
+If the ESC requires to see a special low value pulse first before switching on, it is necessary that the flight control stack sends this value initially.
```
bkueng: This is a good point and I had cases where the ESC would not initialize until I slightly changed the PWM_DISARMED value.

en/hardware/pwm_escs_and_servo.md
```diff
@@ -0,0 +1,132 @@
...
+### Missing Valid Minimum Value
+
+If the ESC requires to see a special low value pulse first before switching on, it is necessary that the flight control stack sends this value initially. 
+The PX4 flight stack for example sends a 900 us pulse while being disarmed, ensuring all ESCs initialise correctly (and do not annoy the user because they beep at him).
+
+### Timeout
+
+Some ESCs might time out (do not allow to switch on the motor) after not having received a valid low pulse a few seconds after power on. 
+This again depends on the flight control stack. 
+As the PX4 flight stack sends a 900 us idle pulse right after power on, ESCs will not time out.
```
bkueng: As PX4 handles this, I'm not sure if we should mention it.

bkueng (commented):
> We don't say anything about connecting the motors - is that reasonable? Without knowing anything about this, it sounds to me like the information on connecting up and powering ESCs and servos should be captured.

You're right, would be good to have that. The ESC Power Connection section is a starting point for that and I think it can be moved to the user guide.

en/hardware/pwm_escs_and_servo.md
hamishwillee: I think we should, because I'll bet it sends PWM_DISARMED . Hence it is possible that an ESC won't work that does  not set this value properly. I'll update.

hamishwillee:
@bkueng Thanks for your help. I've updated the doc and it is ready for your re-review.

I have made it user facing, and moved the "theory of PWM escs" down into troubleshooting. In most cases this works OK. 

When you are happy with this I will close this PR and copy it to user guide, below https://docs.px4.io/en/peripherals/esc_motors.html

bkueng:
Did you push the changes?

hamishwillee:
@bkueng Yes, but with great skill, to the wrong repo. Fixed. Let us never speak of this again.

en/hardware/pwm_escs_and_servo.md
```diff
@@ -0,0 +1,99 @@
...
+> **Tip** There is **no setup** that does not require signal ground to be connected!
+  
+The servo plug *may also have a +5V wire* (usually red or orange).
+The purpose of this wire and how it is connected depends on the particular ESC in use:
+
+- On a fixed wing ESC it usually provides the output of a Battery Elimination Circuit (BEC). 
+  - This can be connected to the Pixhawk servo rail and used to power servos for flaps, ailerons etc.
+  - As a rule of thumb you should only connect the *output of only one BEC* to the Pixhawk servo rail. 
+    > **Tip** While it may be possible to connect multiple +5V outputs to the rail, this depends on the ESC model.
+- On a multicopter ESC this line might be not be present or (if present) may not be connected.
```
bkueng, line 26: might not be present

bkueng (approved):
> @bkueng Yes, but with great skill, to the wrong repo. Fixed. Let us never speak of this again.

Haha, I won't :rofl: 
Looks good.

hamishwillee:
THanks. Discussion moved to https://github.com/PX4/px4_user_guide/pull/337

PR #606, opened 2018-09-11
Add note that default O line can be omitted.

hamishwillee:
Note that I chose NOT to delete the line from the example that follows. Maybe I should (or put in a comment line?)


RicardoM17:
I think the line is still relevant in order to explain how it actually works (in case someone wants to use other than default settings).

bkueng (approved):
I think we can keep it in the example for completness reasons. The reason to remove it is to reduce FLASH usage - which is not a concern if you have a mixer on the SD card.

hamishwillee:
Great. Thanks guys!

PR #608, opened 2018-09-11
Add Dronecode governance notice

mrpollo:
The PX4 flight stack is hosted under the governance of the [Dronecode Project](https://www.dronecode.org/).


hamishwillee:
Thanks. Looks good.

PR #610, opened 2018-09-16
Add section on starting JMAVSIm/PX4 separately

hamishwillee:
@bkueng This originates from https://github.com/PX4/Firmware/pull/10481

Any particular reasons why you might want to do this?

### DIFF
--- a/en/simulation/jmavsim.md
+++ b/en/simulation/jmavsim.md
@@ -96,6 +96,14 @@ make broadcast jmavsim
 
 The simulator broadcasts its address on the local network as a real drone would do.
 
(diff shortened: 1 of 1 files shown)


bkueng:
We already have this here: https://github.com/PX4/Devguide/blame/ff9080c2570397b31201068feb45a96da6a54a9e/en/simulation/README.md#L119

It's useful for a faster testing cycle. Restarting jMAVSim simply takes quite a bit longer.

hamishwillee:
Thanks @bkueng . I think it is worth having in both places - because I'm fairly sure that many people reading the jMAVSim docs will choose not to read that more detailed doc.

PR #616, opened 2018-09-24
Update reference-design.md

ocularmagic:
FMUv3 pointed to FMUv2, updated the link to point to v3.


hamishwillee:
Thanks @ocularmagic - good catch! 

@dagar Jenkins build failing for some reason. It is not a fault with this PR.

PR #618, opened 2018-09-27
Mixer - Update link

bresch:
The link didn't redirect to the correct section in the page.
Here is the correct link: https://github.com/PX4/Devguide/blob/master/en/airframes/adding_a_new_frame.md#mixer-file-mixer-file


hamishwillee:
When gitbook builds the page, it sets the heading anchor using the syntax below. So this anchor will work in gitbook but not in github. Closing as invalid.

image

As an aside, you may ask "why not just remove the gitbook markup and use the github link - as this does work". The reason is that using this approach I can still change the heading name without changing the text. At some point I will have done that or this anchor would not exist (I don't put them in everywhere).

bresch:
Sorry, my bad, I think I was too tired xD

PR #622, opened 2018-10-02
Change bash to /bin/sh

bramsvs:
See: https://github.com/PX4/Firmware/pull/10582

### DIFF
--- a/en/concept/system_startup.md
+++ b/en/concept/system_startup.md
@@ -12,7 +12,7 @@ The following sections are split according to the operating system that PX4 runs
 
 ## Posix (Linux/MacOS)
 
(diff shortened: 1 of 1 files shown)


en/concept/system_startup.md
```diff
@@ -12,7 +12,7 @@ The following sections are split according to the operating system that PX4 runs
 
 ## Posix (Linux/MacOS)
 
-On Posix, the system shell is used as script interpreter (e.g. bash).
+On Posix, the system shell is used as script interpreter (e.g. /bin/sh).
```
bkueng: Can you change this to something like `(/bin/sh which is dash on Ubuntu)`?

bkueng (approved):
Thanks.

PR #626, opened 2018-10-09
land detector: remove LNDMC_THR_RANGE

bkueng:
Got removed from the Firmware, see https://github.com/PX4/Firmware/pull/10665.


hamishwillee:
Whoops - merged too early. Sorry @bkueng - this is also mentioned lower in the [land detector document](http://dev.px4.io/en/tutorials/land_detector.html#maybe-landed) under "Maybe landed"

bkueng:
Fix pushed: https://github.com/PX4/Devguide/commit/0b47542bab7ed2cd3dddacfcb530428dcebcd2fd

PR #631, opened 2018-10-16
fixed class info and added image for Plane V-Tail

fredowski:
I added the image for the Plane V-Tail and fixed a problem due
to the autogenerated class information from the PX4/Firmware source.

PR PX4/Firmware#10414 added the new X-UAV Mini Talon Airframe and the Plane V-Tail group.

This PR adds the image for the new group. The class information "Plane" was malformed because there was additional comment. The malformed class info is also addressed in this commit. 

The original comment which resulted in this problem of malformed class info is changed with PX4/Firmware#10699 .


hamishwillee:
Hi @fredowski ,

I'm closing this because the changes to airframe reference are supposed to come from buildbot. I've fixed the script, so that should happen when that PR gets accepted. 

I've separately added the image for user guide and devguide (see https://github.com/PX4/Devguide/pull/634
and https://github.com/PX4/px4_user_guide/pull/369)

Cheers
H

fredowski:
O.k. so the image information is hardcoded in the bot script. Thanks for clarification!

hamishwillee:
> O.k. so the image information is hardcoded in the bot script. Thanks for clarification!

Yes, but only for the group, not the specific airframe. If no specific group-> image mapping is provided a placeholder is inserted (this was broken, but is now fixed)

PR #635, opened 2018-10-19
added how to introduce airframe group with image

fredowski:
Introducing a new airframe group with a new image requires
to introduce the image name in a botscript and to checkin
the image itself at qgc, devguide and userguide. Added the
procedure how to do this.

### DIFF
--- a/en/airframes/adding_a_new_frame.md
+++ b/en/airframes/adding_a_new_frame.md
@@ -192,6 +192,27 @@ S: 0 3      0  20000 -10000 -10000  10000
 
 ```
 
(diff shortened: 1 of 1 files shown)


hamishwillee:
Thanks very much @fredowski  I've re-written it slightly to my taste: https://github.com/PX4/Devguide/commit/e0c2e29cc6971fea4dc915922e2a0be9e80b37b9

(Really appreciate you doing this - it is much easier to improve something that someone has written than motivate to capture information in the first place).

fredowski:
@hamishwillee Thanks for looking at the patch making it even nicer!

PR #640, opened 2018-10-31
px4_ros_com docs

TSC21:
This is the ongoing work on the `px4_ros_com` package documentation, which is to be release by the end of this week. The release is dependent of having good documentation about the package, so any feedback on the ongoing work is important. In the mean time, here's the list of things that need to be addressed on the docs:

- [X] Architectural overview of how the ROS side interacts with the current existent micro-RTPS client/agent architecture (theoretical)
- [x] Same as the above, but in a diagram that clarifies the architecture
- [X] Simple build structure explanation and instructions
- [x] Explain the code used on the example nodes added on the package
- others?...


en/middleware/micrortps.md
```diff
@@ -1,24 +1,26 @@
 # RTPS/ROS2 Interface: PX4-FastRTPS Bridge
 
-The *PX4-FastRTPS Bridge* adds a Real Time Publish Subscribe (RTPS) interface to PX4, enabling the exchange of [uORB messages](../middleware/uorb.md) between PX4 components and (offboard) *FastRTPS* applications.
+The *PX4-FastRTPS Bridge* adds a Real Time Publish Subscribe (RTPS) interface to PX4, enabling the exchange of [uORB messages](../middleware/uorb.md) between PX4 components and (offboard) *Fast-RTPS* applications, including ones built over the ROS2/ROS frameworks.
```
hamishwillee: eProsima call this *Fast RTPS* so we should probably use that rather than Fast-RTPS (or FastRTPS - which is what the original authors used).

hamishwillee: > applications, including ones built over the ROS2/ROS frameworks.

TO

> applications (including those built over the ROS2/ROS frameworks).

en/middleware/micrortps.md
```diff
@@ -1,24 +1,26 @@
...
+> **Note** RTPS is the underlying protocol of the Object Management Group's (OMG) Data Distribution Service (DDS) standard. It aims to enable scalable, real-time, dependable, high-performance and interoperable data communication using the publish/subscribe pattern. *Fast-RTPS* is a very lightweight cross-platform implementation of the latest version of the RTPS protocol and a minimum DDS API.
 
 RTPS has been adopted as the middleware for the ROS2 (Robot Operating System). The bridge allows us to better integrate with ROS2, making it easy to share sensor values, commands, and other vehicle information.
 
-This topic describes the bridge architecture, how it is compiled, and how to write a simple FastRTSP application to subscribe to PX4 changes.
+This topic describes the bridge architecture, how it is compiled, and how to:
+1. Write a simple FastRTSP application to subscribe to PX4 changes;
```
hamishwillee: *Fast RTPS* - note that this says RTSP currently :-(

en/middleware/micrortps.md
```diff
@@ -1,24 +1,26 @@
...
+2. Or, if your application is on the ROS2/ROS1 side, understand how the `px4_ros_com` package works and how to use it to bridge your ROS nodes with PX4.
```
hamishwillee: FYI Markdown allows you to use `1.` for every bullet. That is useful if you might later move bullets around.

Suggest we change this to:
```
1. For ROS2/ROS1 applications, understand how the `px4_ros_com` package works and how to use it to bridge ROS nodes with PX4.
```

en/middleware/micrortps.md
```diff
@@ -30,58 +32,84 @@ The main elements of the architecture are the client and agent processes shown i
...
+
+
+## Architectural overview for a ROS2/ROS application pipeline
+
+<!--![basic example flow](../../assets/middleware/micrortps/architecture_ros.png)-->
+
+ROS2 has been developed on top of the DDS/RTPS, which is what composes its middleware. This same middleware serves as the end-to-end architecture for plugging different applications that rely on distributed discovery, serialization and QoS control over the transportation layer.
+
+So one is able to publish and subscribe to uORB data from PX4 using ROS2 (and ROS1 if using the [`ros1_bridge`](https://github.com/ros2/ros1_bridge), it just needs to create it's own listener our advertiser nodes in the respective frameworks. The only consideration to take into account is that the msg types, headers and source files being used on both client and agent side (and consequently, on the ROS side) need to be the same and generated from the same IDL (Interface Description Language).
```
hamishwillee: Everywhere remove the back ticks (`) from the link text shown below. It doesn't need the additional highlighting.
```
[`ros1_bridge`] 
```

en/middleware/micrortps.md
```diff
@@ -30,58 +32,84 @@ The main elements of the architecture are the client and agent processes shown i
...
-- The *Agent* and any *FastRTPS* applications are connected via UDP, and may be on the same or another device. In a typical configuration they will both be on the same system (e.g. a development computer, Linux companion computer or compute board), connected to the *Client* over a Wifi link or via USB.
+- The *Agent* and any *Fast-RTPS* applications are connected via UDP, and may be on the same or another device. In a typical configuration they will both be on the same system (e.g. a development computer, Linux companion computer or compute board), connected to the *Client* over a Wifi link or via USB.
+
+
+## Architectural overview for a ROS2/ROS application pipeline
+
+<!--![basic example flow](../../assets/middleware/micrortps/architecture_ros.png)-->
+
+ROS2 has been developed on top of the DDS/RTPS, which is what composes its middleware. This same middleware serves as the end-to-end architecture for plugging different applications that rely on distributed discovery, serialization and QoS control over the transportation layer.
```
hamishwillee: I'm not sure why you have this paragraph / why is it needed? 

What I would expect here is a sentence that says something about why we're talking about ROS2 here in a topic about RTPS. Something like  "RTPS is the ROS2 native communications middleware, which makes it easy to integrate with PX4".

en/middleware/micrortps.md
hamishwillee: This paragraph is confusing
- the brackets don't match up, and without a diagram I can't see how it "should" be. 
- Typo: "it just needs to create it's own listener our advertiser nodes in the respective frameworks." - I think you mean "and advertiser nodes"
- I can't see where ros1_bridge fits in. It sounds like you only need this if working with ROS1. I would pull ROS 1 out into its own paragraph. 
- We say "e only consideration to take into account". Isn't this always the same consideration? 
- `it just needs to create it's own listener our advertiser nodes in the respective frameworks.` - what is "IT" here? ROS2 needs to create its own nodes? What do you mean by "listener and advertiser nodes"? Do you mean uOrb topics?

Here is a _guess_ at what I think you want to say in this section:

```
RTPS is the ROS2 native communications middleware, which makes it easy to integrate with PX4.

In order to publish and subscribe to PX4 uORB data from ROS2 (and visa versa) you will need to create matching uORB topics in PX4 and listener/advertiser in ROS2. The msg types, headers and source files used on both client and agent side (and consequently, on the ROS side) must be generated from the same IDL (Interface Description Language) definition.

ROS1 integration with is also supported via the [ros1_bridge](https://github.com/ros2/ros1_bridge)  <NEED DIAGRAM TO SEE WHERE THAT FITS>
```

en/middleware/micrortps.md
```diff
@@ -30,58 +32,84 @@ The main elements of the architecture are the client and agent processes shown i
...
+
+So one is able to publish and subscribe to uORB data from PX4 using ROS2 (and ROS1 if using the [`ros1_bridge`](https://github.com/ros2/ros1_bridge), it just needs to create it's own listener our advertiser nodes in the respective frameworks. The only consideration to take into account is that the msg types, headers and source files being used on both client and agent side (and consequently, on the ROS side) need to be the same and generated from the same IDL (Interface Description Language).
 
 
 ## Code generation
 
-All the code needed to create, build and use the bridge is automatically generated when the PX4 Firmware is compiled. 
+### Code generation for ROS-independent applications
```
hamishwillee: Up to you, but I might remove "Code generation for " in this heading and the next one.

en/middleware/micrortps.md
```diff
@@ -30,58 +32,84 @@ The main elements of the architecture are the client and agent processes shown i
...
+All the code needed to create, build and use the bridge is automatically generated when the PX4 Firmware is compiled.
 
 The *Client* application is also compiled and built into the firmware as part of the normal build process. The *Agent* must be separately/manually compiled for the target computer.
 
 > **Note** [Fast RTPS must be installed](../setup/fast-rtps-installation.md) in order to generate the required code!
 
 <span></span>
-> **Tip** The bridge code can also be [manually generated](micrortps_manual_code_generation.md). Most users will not need to do so, but the linked topic provides a more detailed overview of the build process and can be useful for troubleshooting.
+> **Tip** The bridge code can also be [manually generated](micrortps_manual_code_generation.md). Most users will not need to do
+so, but the linked topic provides a more detailed overview of the build process and can be useful for troubleshooting.
```
hamishwillee: This probably doesn't have enough indentation. I would put the break on "Most users" and indent 2 spaces.

en/middleware/micrortps.md
```diff
@@ -30,58 +32,84 @@ The main elements of the architecture are the client and agent processes shown i
...
 
+### Code generation for ROS2/ROS applications - `px4_ros_com`
 
-## Supported uORB messages
-
-The generated bridge code will enable a specified subset of uORB topics to be published/subscribed via RTPS.
+The [`px4_ros_com`](https://github.com/PX4/px4_ros_com) package allows the developer to generate all the required ROS2/ROS message headers and sources, but also the agent application which will bridge the client with the RTPS middleware on the ROS2 side. The package is basically composed by two separate branches, where the `master` branch generates all the required ROS2 messages and IDL files, besides some listener and advertiser example nodes, in order to bridge PX4 with ROS2 nodes, and a `ros1` branch, which similarly to the `master` branch, generates the ROS(1) message headers and source files, besides the ROS nodes examples, that allow, by using the `ros1_bridge`, share data between PX4 and ROS.
```
hamishwillee: - I get an HTTP404 on this link https://github.com/PX4/px4_ros_com
- when you say "besides some listener" do you mean "this also has some example nodes"?

hamishwillee: How about:
```
The [px4_ros_com](https://github.com/PX4/px4_ros_com) package allows the developer to generate all the required ROS2/ROS message headers and sources, and the agent application that will bridge the client with the RTPS middleware on the ROS2 side. 

The package is has two separate branches:
- `master` is used with ROS2. It contains code to generate all the required *ROS2* messages and IDL files to bridge PX4 with ROS2 nodes. 
- `ros1` is used with ROS1. It contains code to generate the ROS(1) message headers and source files, which can be used *with* the `ros1_bridge` to share data between PX4 and ROS.

Both branches additionally include some example listener and advertiser example nodes. <!-- perhaps we add to this: "advertiser example nodes that can be used for X, Y, Z" 
ie it isn't clear if these examples are required or just interesting -->
```

en/middleware/micrortps.md
```diff
@@ -113,12 +147,12 @@ cmake ..
 make
 ```
 
-> **Note** To cross-compile for the *Qualcomm Snapdragon Flight* platform see [this link](https://github.com/eProsima/PX4-FastRTPS-PoC-Snapdragon-UDP#how-to-use).
+> **Note** To cross-compile for the *Qualcomm Snapdragon Flight* platform see [this link](https://github.com/eProsima/PX4-Fast-RTPS-PoC-Snapdragon-UDP#how-to-use).
```
hamishwillee: This is a URL. Revert.

en/middleware/micrortps.md
```diff
@@ -30,58 +32,84 @@ The main elements of the architecture are the client and agent processes shown i
...
+The [`px4_ros_com`](https://github.com/PX4/px4_ros_com) package allows the developer to generate all the required ROS2/ROS message headers and sources, but also the agent application which will bridge the client with the RTPS middleware on the ROS2 side. The package is basically composed by two separate branches, where the `master` branch generates all the required ROS2 messages and IDL files, besides some listener and advertiser example nodes, in order to bridge PX4 with ROS2 nodes, and a `ros1` branch, which similarly to the `master` branch, generates the ROS(1) message headers and source files, besides the ROS nodes examples, that allow, by using the `ros1_bridge`, share data between PX4 and ROS.
 
-For *automatic code generation* (via the normal PX4 firmware build process) this set must be listed in the **/path/to/PX4/Firmware/src/modules/micrortps_bridge/CMakeLists.txt** file.
+The `px4_ros_com` package build structure is structured in a way that, when one builds the package, it automatically generates all the required components of the bridge that allow the developer to use the generated messages on its own nodes, being it on the ROS2 or in the ROS side.
```
hamishwillee: > The `px4_ros_com` package build structure is structured in a way that, when one builds

TO

> The `px4_ros_com` package is structured such that when one builds the package

hamishwillee: > it automatically generates all the required components of the bridge that allow the developer to use the generated messages on its own nodes, being it on the ROS2 or in the ROS side."

Again, not clear whether you need this bridge for just ROS1 or also for ROS2. Para here seems to indicate for both. 

This talks about generated messages being used for ROS1?ROS2 nodes, but it isn't clear what components or messages are built with px4_ros_com that a developer might use. Are they the agent messages (from the IDL)?

en/middleware/micrortps.md
```diff
@@ -30,58 +32,84 @@ The main elements of the architecture are the client and agent processes shown i
...
 ## Client (PX4 Firmware)
 
-The *Client* source code is generated, compiled and built into the PX4 firmware as part of the normal build process. 
+The *Client* source code is generated, compiled and built into the PX4 firmware as part of the normal build process.
 
-To build and upload the firmware for NuttX/Pixhawk flight controllers:
+To build the firmware for NuttX/Pixhawk flight controllers:
 ```sh
-make px4fmu-v4_default upload
+make px4fmu-v4_rtps
```
hamishwillee, line 117: Is this change to the tooling in? 
This seems odd - does this mean that RTPS is not in firmware by default?

en/middleware/micrortps.md
```diff
@@ -30,58 +32,84 @@ The main elements of the architecture are the client and agent processes shown i
...
 
 ## Client (PX4 Firmware)
 
-The *Client* source code is generated, compiled and built into the PX4 firmware as part of the normal build process. 
+The *Client* source code is generated, compiled and built into the PX4 firmware as part of the normal build process.
 
-To build and upload the firmware for NuttX/Pixhawk flight controllers:
+To build the firmware for NuttX/Pixhawk flight controllers:
 ```sh
-make px4fmu-v4_default upload
```
hamishwillee, line 77: >  To build the firmware for NuttX/Pixhawk flight controllers:

Maybe to

> To build the firmware for NuttX/Pixhawk flight controllers use the _rtps feature in the configuration target. 
For example, to build RTPS for px4fmu-v4:

en/middleware/micrortps.md
```diff
@@ -30,58 +32,84 @@ The main elements of the architecture are the client and agent processes shown i
...
+    id: 5
+    send: true
+  - msg: battery_status
+    id: 6
+    send: true
+  - msg: camera_capture
+    id: 7
+  - msg: camera_trigger
+    id: 8
+    receive: true
```
hamishwillee, line 97: Can a message be both sent and received? If so, show example of this.
What is the default? (ie the first example does not have either)

en/middleware/micrortps.md
```diff
@@ -30,58 +32,84 @@ The main elements of the architecture are the client and agent processes shown i
...
 
-set(config_rtps_receive_topics
-   sensor_baro
-   # Add new topic...
-   )
+The generated bridge code will enable a specified subset of uORB topics to be published/subscribed via RTPS. This applicable on both ROS or non-ROS applications.
+
+For *automatic code generation* (via the normal PX4 Firmware build process or, in the case of the `px4_ros_com`, through the build process of the package), there's a yaml definition file on the PX4 Firmware directory called `uorb_rtps_message_ids.yaml`, found in **/path/to/PX4/Firmware/msg/tools/** dir, where all the uORB messages (RTPS) ID's are set and where is also set which messages should be set to be sent, received, or both, in the RTPS stream.
+
+> **Caution** All new uORB messages which one adds to the Firmware side should have a respective RTPS ID set when one wants the data to be published and/or subscribed in the RTPS stream.
```
hamishwillee: Are you saying that the uORB definition in firmware source code must include an RTPS ID field? If so, we should show an example. 
Or are you saying that an ID needs to be in the YAML as shown below.
Or are you saying both - you define a field in firmware and have to match it up here in the YAML?

What is the strategy for allocating ids?

en/middleware/micrortps.md
```diff
@@ -30,58 +32,84 @@ The main elements of the architecture are the client and agent processes shown i
...
-   )
+## Supported uORB messages
 
-set(config_rtps_receive_topics
-   sensor_baro
-   # Add new topic...
-   )
+The generated bridge code will enable a specified subset of uORB topics to be published/subscribed via RTPS. This applicable on both ROS or non-ROS applications.
+
+For *automatic code generation* (via the normal PX4 Firmware build process or, in the case of the `px4_ros_com`, through the build process of the package), there's a yaml definition file on the PX4 Firmware directory called `uorb_rtps_message_ids.yaml`, found in **/path/to/PX4/Firmware/msg/tools/** dir, where all the uORB messages (RTPS) ID's are set and where is also set which messages should be set to be sent, received, or both, in the RTPS stream.
```
hamishwillee: This yaml file is used to generated the IDL file(s) from the uOrb message definitions right?

I think you are saying:
```
For *automatic code generation* there's a yaml definition file in the PX4 **Firmware/msg/tools/** directory called **uorb_rtps_message_ids.yaml**.
This file defines the set of uORB messages to be used with RTPS, whether the messages are to be sent, received or both, and the RTPS ID for the message to be used in RTPS nodes.
```

en/middleware/micrortps.md
```diff
@@ -30,58 +32,84 @@ The main elements of the architecture are the client and agent processes shown i
...
+    send: true
+  - msg: camera_capture
+    id: 7
+  - msg: camera_trigger
+    id: 8
+    receive: true
 ```
 
-> **Caution** At time of writing (August 2017), only the small set of uORB topics listed above are included in our cmake files: **posix_sitl_default.cmake**, **nuttx_px4fmu-v4_default.cmake**, **posix_sdflight_default.cmake**. It is likely you will need to edit your *cmake* file and add additional uORB topics. In future we hope to define a larger standard set. 
+In the case of `px4_ros_com`, the `uorb_rtps_message_ids.yaml` is transformed in a way that the message names become PascalCased and there's a swap between the received and the sent messages. The logic behind this is that if a message is sent from the client side, then it's received on the agent side, and vice-versa. The naming of the message is indifferent on the client-agent communication, but it is critical on the ROS2 side, since the message naming has to follow the the PascalCase convention, which means that the generated IDL files also need to follow this same convention.
```
hamishwillee: This information doesn't seem to match the topic "## Supported uORB messages".
- It is additional/supporting so perhaps should be in  a note, a different section or a sub section.
- Does this transformation only happen when you build using `px4_ros_com`? Or are the ROS IDl files built whenever you build firmware. 

I'll suggest best path once you tell me the answer to the second point. Something like (if this is correct!)

```
The yaml file is used to create IDL files for specified uORB messages as part of the build process.

Separate IDL files are build for use by ROS2 nodes and other RTPS nodes. 

> **Note** ROS2 requires PascalCased node names. The `px4_ros_com` build process automatically creates appropriate IDL files for use with ROS2 nodes. 
  These are mapped by the agent to the equivalent messages PX4-side, using the RTPS ID.
```

en/middleware/micrortps.md
```diff
@@ -98,9 +126,15 @@ The *Client* application can be launched from [NuttShell/System Console](../debu
   -s <sending port>       UDP port for sending. Default 2020
 ```
 
-> **Note** By default the *Client* runs as a daemon, but you will need to start it manually. The PX4 Firmware initialisation code may in future automatically start the *Client* as a permanent daemon process.<!-- at that point, most of this section would move into the "manual generation" doc: https://github.com/PX4/Firmware/pull/7663#issuecomment-317928506 -->
+> **Note** By default the *Client* runs as a daemon, but you will need to start it manually. The PX4 Firmware initialization code may in future automatically start the *Client* as a permanent daemon process.
```
hamishwillee: Just FYI, there is no standard for US or internationalized English. Upshot, generally you find me using "s", but I am consistent within a document.

en/middleware/micrortps.md
```diff
@@ -98,9 +126,15 @@ The *Client* application can be launched from [NuttShell/System Console](../debu
...
 
-## Agent (Off Board FastRTPS Interface)
+For example, in order to run the *Client* daemon with SITL, and since the data is being shared through UDP, it's required that the daemon is started using the UDP as the transport protocol:
+
+```sh
+micrortps_client start -t UDP
```
hamishwillee, line 145: It would be MUCH better if this was set through a user-side parameter.
@bkueng just did some great stuff with Telemetry and sensors along these lines.

en/middleware/micrortps.md
```diff
@@ -131,16 +165,137 @@ $ ./micrortps_agent [options]
...
 
+As an example, to start the *micrortps_agent* with connection through UDP, issue:
+
+```sh
+./micrortps_agent -t UDP
+```
+
+## Agent interfacing with a ROS2 middleware
+
+`px4_ros_com` build structure allows to automatically generate and build the agent application. Since it is also installed using the [`colcon`](http://design.ros2.org/articles/build_tool.html) build tools, running it works exactly the same way as the above. Check the **Building the `px4_ros_com` package** for details about the build structure.
```
hamishwillee: > `px4_ros_com` build structure allows to automatically generate and build the agent application. Since it is also installed using the [`colcon`]

TO

> Building `px4_ros_com` automatically generates and builds the agent application.

en/middleware/micrortps.md
```diff
@@ -131,16 +165,137 @@ $ ./micrortps_agent [options]
...
+
+
+## Building the `px4_ros_com` package
```
hamishwillee, line 190: Add anchor by appending to end of heading line: `{#px4_ros_com}`

en/middleware/micrortps.md
hamishwillee: > Since it is also installed using the [`colcon`](http://design.ros2.org/articles/build_tool.html) build tools, running it works exactly the same way as the above. 

This doesn't follow - how does installing with colcon mean that you start the agent in the same way as without ROS?

> Check the **Building the `px4_ros_com` package** for details about the build structure.

Link to the section "Building the `px4_ros_com` package" below. YOu can use anchor `#building_px4_ros_com` - I will add a note about that below.

en/middleware/micrortps.md
```diff
@@ -131,16 +165,137 @@ $ ./micrortps_agent [options]
...
+
+As aforementioned, the `px4_ros_com` comes bundled with two branches, where one links the ROS2 with the PX4-agent/client bridge and the other allows, through `ros1_bridge`, to link the ROS framework with the ROS2 framework through the same set of messages. Therefore, both branches need to be cloned separately so the build process can happen correctly. Before that can happen, one requires to install and setup both ROS2 and ROS environments on its machine.
```
hamishwillee, line 192: Change to:

```
Before building `px4_ros_com` you will first need to clone the PX4 Firmware repository
(this is normally placed at the same tree level as the ROS workspaces).
Then install and setup both ROS2 and ROS environments on your development machine
and separately clone the `px4_ros_com`  repo for both the `master` and `ros1` branches (see [above for more information](#insert_anchor_up_to_where_this_is)).
```

This text is not needed here, but should make sure that it is clear above:

> where one links the ROS2 with the PX4-agent/client bridge and the other allows, through `ros1_bridge`, to link the ROS framework with the ROS2 framework through the same set of messages." is useful. 

Specifically, this provide a bit more clarity about the architecture, and possibly could live up in that section.(though again, without diagram, still not sure how this works)

en/middleware/micrortps.md
```diff
@@ -131,16 +165,137 @@ $ ./micrortps_agent [options]
...
+
+> **Note** This can be taken as the step 0 to build the package: the package relies on the PX4 Firmware directory, which in the build process it tries to find using a cmake module (`FindPX4Firmware.cmake`). It will try to find the package in several known directories and if it doesn't find it, it will fail the build! The most common place to put the Firmware directory is at the same tree level of the ROS workspaces.
```
hamishwillee, line 194: Suggested:
```
> **Note** The package build relies on the PX4 Firmware directory. 
  The build process tries to find it using a *cmake* module (`FindPX4Firmware.cmake`).
  If the directory cannot be found the build will fail.
```

en/middleware/micrortps.md
```diff
@@ -131,16 +165,137 @@ $ ./micrortps_agent [options]
...
+
+### Installing ROS and ROS2 and respective dependencies
+
+In order to install ROS Melodic and ROS2 Bouncy on an Ubuntu 18.04 machine, follow the links bellow, respectively:
```
hamishwillee, line 198: below, not bellow

hamishwillee: Does Firmware build on Ubuntu 18.04. FYI we don't regularly test on that.

en/middleware/micrortps.md
```diff
@@ -131,16 +165,137 @@ $ ./micrortps_agent [options]
...
+1. [Install ROS Melodic](http://wiki.ros.org/melodic/Installation/Ubuntu)
+2. [Install ROS2 Bouncy](https://index.ros.org/doc/ros2/Linux-Install-Debians/)
+
+For the ROS2 case, so the package properly generates the IDL files, the following should be installed:
+
+```sh
+sudo apt install ros-bouncy-rmw-opensplice-cpp
+```
+
+The install process should also install the `colcon` build tools, but in case that doesn't happen, you can install it manually:
```
hamishwillee, line 209: The ROS install process should ...

en/middleware/micrortps.md
```diff
@@ -131,16 +165,137 @@ $ ./micrortps_agent [options]
...
+
+```sh
+sudo apt install python3-colcon-common-extensions
+```
+
+> **Note** This install and build guide is also aplicable in an environment with Ubuntu 16.04, ROS Kinetic and ROS2 Ardent installed.
```
hamishwillee, line 215: aplicable -> applicable.

en/middleware/micrortps.md
```diff
@@ -131,16 +165,137 @@ $ ./micrortps_agent [options]
...
+
+> **Note** ROS2 requires Python3, which will be install on your system. Be aware of any incompatibilities you may find with other packages that were installed, for example by pip, using Python2 and you may want to install using `pip3` so to solve these incompatibilities.
```
hamishwillee, line 217: There shouldn't be any dependency issues unless you are installing dependencies, in which case you MUSt install pip3. SO suggest you either delete this if not needed by build process or instructions - or do something like:

```
> **Note** ROS2 requires Python3, which is already installed on Ubuntu 18.04.
  You should install *pip3* in order to be able to install dependencies. <!-- HERE, show how -->
```

en/middleware/micrortps.md
```diff
@@ -131,16 +165,137 @@ $ ./micrortps_agent [options]
...
+
+> **Caution** Do not install the `ros1_bridge` package through the deb repository. The package has to be built by source.
```
hamishwillee, line 219: TO
```
> **Caution** Do not install the `ros1_bridge` package from the deb repository. 
  The package must be built from source.
```

en/middleware/micrortps.md
```diff
@@ -131,16 +165,137 @@ $ ./micrortps_agent [options]
...
+
+### Setting up the workspaces
+
+Since the ROS2 and ROS require different environments to be set, there should exist two different workspaces for each ROS version. As an example:
```
hamishwillee, line 223: " there should exist two different workspaces for each ROS version."

That says that you need 4 workspaces in total. Or do you mean "there should exist a different workspace for each ROS version."

en/middleware/micrortps.md
```diff
@@ -131,16 +165,137 @@ $ ./micrortps_agent [options]
...
+
+Then, clone the respective ROS2 (`master`) branch to the `/src` directory:
+
+```sh
+$ git clone https://github.com/PX4/px4_ros_com.git ~/px4_ros_com_ros1/src/px4_ros_com # clones the 'ros1' branch
+```
+
+### Building the workspaces
+
+For building the workspaces, there's already a script available on the `px4_ros_com` package that can be used to automate the build process. But, for a matter of understanding the process, bellow are the steps to manually build the packages:
```
hamishwillee, line 251: TO

```
To build the workspace, just run `build_ros2_side.bash` under `px4_ros_com/scripts`.

The steps below show how to *manually* build the packages (provided for your information only):
```

en/middleware/micrortps.md
```diff
@@ -131,16 +165,137 @@ $ ./micrortps_agent [options]
...
+
+> **Note** If you want to skip a step-by-step build of the package, just run `build_ros2_side.bash` under `px4_ros_com/scripts`.
```
hamishwillee, line 253: Delete, see above.

en/middleware/micrortps.md
```diff
@@ -131,16 +165,137 @@ $ ./micrortps_agent [options]
...
+
+1. `cd` into `px4_ros_com_ros2` dir and source the ROS2 environment. Don't mind if it tells you that a previous workspace was set before:
```
hamishwillee, line 255: Don't mind if it tells you that a previous workspace was set before:

TO

Ignore warnings that a workspace has been set before.

en/middleware/micrortps.md
```diff
@@ -131,16 +165,137 @@ $ ./micrortps_agent [options]
...
+
+```sh
```
hamishwillee, line 257: All the lines of a code block should be indented three spaces to become part of the preceding list item. This true for all cases below.

en/middleware/micrortps.md
```diff
@@ -131,16 +165,137 @@ $ ./micrortps_agent [options]
...
+git clone https://github.com/ros2/ros1_bridge.git ~/px4_ros_com_ros2/src/ros1_bridge
+```
+
+3. Build the `px4_ros_com` package, excluding the `ros1_bridge` package:
+
+```sh
+colcon build --symlink-install --packages-skip ros1_bridge --event-handlers console_direct+
+```
+
+> **Note** `--event-handlers console_direct+` only serves the purpose of adding verbosity to the `colcon` build process and can be removed if one wants a more "quite" build.
```
hamishwillee, line 273: This should also be indented 3 spaces

en/middleware/micrortps.md
```diff
@@ -131,16 +165,137 @@ $ ./micrortps_agent [options]
...
+
+4. Then, follows the process of building the ROS(1) packages side. For that, one requires to source the environments so when the `ros1_bridge` is built with support for any messages that are on PATH and have an associated mapping between ROS1 and ROS2:
```
hamishwillee, line 275: Note, you can number all bullets as `1.` - gitbook and github will auto number. It isn't a big deal, but can help if you ever want to re-number.

to:
```
Then build the ROS(1) packages. 
This requires you to source the environments so that the `ros1_bridge` is built with support for any messages that are on PATH and have an associated mapping between ROS1 and ROS2:
```

en/middleware/micrortps.md
```diff
@@ -131,16 +165,137 @@ $ ./micrortps_agent [options]
...
+
+```sh
+source /opt/ros/melodic/setup.bash
+source /opt/ros/bouncy/setup.bash
+```
+
+5. Build the `px4_ros_com` package on the ROS side:
```
hamishwillee, line 282: "on the ROS side" is an odd expression. I presume you mean on the ROS computer?

en/middleware/micrortps.md
```diff
@@ -131,16 +165,137 @@ $ ./micrortps_agent [options]
...
+
+```sh
+cd ~/px4_ros_com_ros1 && colcon build --symlink-install --event-handlers console_direct+
+```
+
+6. Complementing 4., one also needs to source the workspaces after those are built:
```
hamishwillee, line 288: NOte also, you're allowed to refer to the developer as "you". "One" is fine too, but is very formal.

hamishwillee: Also "Complementing 4., " really necessary? You could just say. "Then source the ..."

en/middleware/micrortps.md
```diff
@@ -131,16 +165,137 @@ $ ./micrortps_agent [options]
...
+
+```sh
+source ~/px4_ros_com_ros1/install/setup.bash
+source ~/px4_ros_com_ros2/install/setup.bash
+```
+
+7. At last, build the `ros1_bridge`. Note that the build process may consume a lot of memory resources, so if one is using a resource limited machine, reduce the number of jobs being processed in parallel, by for example, setting the environmental variable `MAKEFLAGS=-j1`. For more details on the build process, one can consult the build instructions on the [`ros1_bridge`](https://github.com/ros2/ros1_bridge) package page.
```
hamishwillee: At last, build TO Finally, build ...

en/middleware/micrortps.md
```diff
@@ -217,13 +372,198 @@ baro_temp_celcius: 43.93
 
 > **Note** If the *Listener application* does not print anything, make sure the *Client* is running.
 
+## Creating a ROS2 listener
+
+With the `px4_ros_com` built successfully, one can now take advantage of the generated *micro-RTPS* agent app and also from the generated sources and headers of the ROS2 msgs, which represent a one-to-one matching with the uORB counterparts.
+
+To create a listener node on ROS2, lets take as an example the `sensor_combined_listener.cpp` node under `px4_ros_com/src/listeners`:
```
hamishwillee, line 390: lets -> let's  (here you're actually saying let us)

en/middleware/micrortps.md
```diff
@@ -217,13 +372,198 @@ baro_temp_celcius: 43.93
...
+		subscription_ = this->create_subscription<px4_ros_com::msg::SensorCombined>("SensorCombined_topic", callback);
+	}
+
+private:
+	rclcpp::Subscription<px4_ros_com::msg::SensorCombined>::SharedPtr subscription_;
+};
+```
 
-## Examples/tests
+The above create a subscription to the `sensor_combined_topic` which can be matched with one or more compatible ROS publishers.
```
hamishwillee, line 441: above createS

hamishwillee: which can be > that can be

en/middleware/micrortps.md
```diff
@@ -217,13 +372,198 @@ baro_temp_celcius: 43.93
...
+
+	rclcpp::shutdown();
+	return 0;
+}
+```
+
+The instantion of the `SensorCombinedListener` class as a ROS node is done on the `main` function.
+
+## Creating a ROS2 advertiser
+
```
hamishwillee, line 459: Is it worth having one sentence saying what an advertiser is?

en/middleware/micrortps.md
```diff
@@ -217,13 +372,198 @@ baro_temp_celcius: 43.93
...
+## Examples/tests of ROS-independent apps
 
 The following examples provide additional real-world demonstrations of how to use the features described in this topic.
 
 * [Throughput test](../middleware/micrortps_throughput_test.md): A simple test to measure the throughput of the bridge.
 
+## Testing the PX4-FastRPTS bridge with ROS2 and ROS
+
+Bellow it is presented a fast way of testing the package, using PX4 SITL with Gazebo:
```
hamishwillee, line 532: Bellow -> Below

en/middleware/micrortps.md
```diff
@@ -250,18 +590,18 @@ export FASTRTPSGEN_DIR=/path/to/fastrtps/install/folder/bin
...
-  groups pi
-  sudo usermod -a -G dialout pi
-  ```
+```sh
+groups pi
+sudo usermod -a -G dialout pi
+```
 
-2. You need to stop the GPIO serial console that is using the port:
+2. For the Raspberry Pi in specific, you need to stop the GPIO serial console that is using the port:
```
hamishwillee, line 649: in specific TO in particular

hamishwillee:
Hi @TSC21 

There are lots of minor grammatical errors, which are easy enough to fix. I've provided some fixes, and I can do another round once this is in.

The real concerns are all in the initial section - architecture. I don't quite see how ROS1 interacts. 

The how to instructions are pretty good. Might be worth at some point pulling them out into another document - but I wouldn't do it now.

Cheerio
H

en/middleware/micrortps.md
TSC21: Well they also have `Fast-RTPS` on their git repo name (https://github.com/eProsima/Fast-RTPS). But I am ok with using `FastRTPS`.

en/middleware/micrortps.md
TSC21: yeah that needs a fix. Thanks

en/middleware/micrortps.md
TSC21: Ok thanks

en/middleware/micrortps.md
TSC21: OK sure. It's just that I am used to use the back ticks to identify code/repository related syntax. I will simplify it then.

en/middleware/micrortps.md
TSC21: Well the idea of this paragraph was to make people understand what's underneath ROS2. I guess we can add your sentence to what I have stated? Or should I replace it completely?

en/middleware/micrortps.md
TSC21: Well actually it's the other way around: you don't mess with the the uORB topics. It's the `px4_ros_com` package that takes the available uORB topic messages and generates the respective sources and headers for the ROS2/ROS messages and also the IDL files required to generate *micrortps_agent*. I will redo this paragraph to make it more clear.

en/middleware/micrortps.md
TSC21: sure

en/middleware/micrortps.md
TSC21: I will take you suggestion. Much cleaner! Thanks!

> when you say "besides some listener" do you mean "this also has some example nodes"?

I mean the node code that you find on the package are just examples of how people can write the nodes to interact with PX4. So they are just "interesting", not required.

en/middleware/micrortps.md
TSC21: ups! Yeah sorry

en/middleware/micrortps.md
TSC21: `px4_ros_com` generates:
- on the `master` branch, the ROS2 messages sources and headers, the IDL files required for the `micrortps_agent` and also builds the `micrortps_agent` with the capability to bridge these message topics between the client and ROS2;
- on the `ros1` branch, the ROS(1) messages sources and headers.

en/middleware/micrortps.md
TSC21: It's not a default feature. You need to build the `micrortps_client`. It should become default soon.

en/middleware/micrortps.md
TSC21: right

en/middleware/micrortps.md
TSC21: Yes it can. I didn't add any example because currently there's none on that same file. The default is what you currently have on the `uorb_rtps_message_ids.yaml` file.

en/middleware/micrortps.md
TSC21: You just need to add the RTPS ID to the respecrive msg/topic on the `uorb_rtps_message_ids.yaml` file. And that's it.

en/middleware/micrortps.md
TSC21: > It is additional/supporting so perhaps should be in a note, a different section or a sub section.

Yeah I should add it as a note.

> Does this transformation only happen when you build using `px4_ros_com`? Or are the ROS IDl files built whenever you build firmware.

This just happens when you build the `px4_ros_com` package.

Actually this needs to be clarified in a different matter. The PX4 Firmware only generates the IDL files required for a bridge between the client-agent and a standard listener app. For the ROS2 side, it's actually the `rosidl_generator` that takes care of that, not PX4. I will make this clear.

en/middleware/micrortps.md
TSC21: Ok I will keep it with an S.

en/middleware/micrortps.md
TSC21: This is not standard yet. This would be also something that you will need to set for each airframe config. For the POSIX case,s that would be using UDP, but for the Nuttx and other targets, it would need to be serial. This is something to address further.

en/middleware/micrortps.md
TSC21: ok

en/middleware/micrortps.md
TSC21: right makes sense.

TSC21:
> There are lots of minor grammatical errors, which are easy enough to fix. I've provided some fixes, and I can do another round once this is in.

Thanks I will address those. I was more concerned with getting everything in place than with the actual form that some things passed through me. I will fix those. Thanks!

> The real concerns are all in the initial section - architecture. I don't quite see how ROS1 interacts.

Yeah I will make that clear. ROS2 interacts with ROS1 through the `ros1_bridge`. The details about the bridge can be found on the bridge repo docs, so I didn't want to be too specific about how they actually interact. I think it is enough to say that for ROS1 to interact with PX4, it needs to `ros1_bridge` so that the data on the topics from the ROS1 side get into the DDS/RTPS middleware on the ROS2 side, which then, by the microrps_agent, get published into the UDP/serial stream to the client (PX4) side.

> The how to instructions are pretty good. Might be worth at some point pulling them out into another document - but I wouldn't do it now.

Yeah I agree. Let's keep it consistent here first and then decide to move it in a further period.

en/middleware/micrortps.md
hamishwillee: No, I am proposing *Fast RTPS* (with the italicisation). On their site they just call it Fast RTPS. It is a trivial change I can do as a post process.

en/middleware/micrortps.md
TSC21: I wasn't actually able to make this work. Can be addressed later I guess.

TSC21:
@hamishwillee I think I have addressed most of your comments. I will add the diagram later today. If all is ok please do merge it. The plan is to release the package by the 2nd of November, so the docs should be there so people know how to use and test the package.

en/middleware/micrortps.md
TSC21: Yes it does build on 18.04.

en/middleware/micrortps.md
TSC21: I think we can just remove this note.

en/middleware/micrortps.md
TSC21: yeah, a workspace for each ROS version.

en/middleware/micrortps.md
TSC21: Well, I mean the ROS(1) end.

en/middleware/micrortps.md
TSC21: not necessary.

en/middleware/micrortps.md
TSC21: Well we can say something as "An adverstiser node is a ROS node which publishes data into the DDS/RTPS layer."

hamishwillee:
Merging now (to allow further subediting)

en/middleware/micrortps.md
hamishwillee: I'll update this in my PR and you can decide if it is OK.

en/middleware/micrortps.md
hamishwillee: Done in new update

en/middleware/micrortps.md
hamishwillee: Fixed in next version

en/middleware/micrortps.md
hamishwillee: I have done this in my updated version

en/middleware/micrortps.md
hamishwillee: Done.

en/middleware/micrortps.md
hamishwillee: Fixed in my update

en/middleware/micrortps.md
hamishwillee: Cool. I will update to "An ROS advertiser node publishes data into the DDS/RTPS network (and hence to PX4)."

hamishwillee:
See https://github.com/PX4/Devguide/pull/642 for updates

PR #642, opened 2018-11-02
Corrections to RTPS docs

hamishwillee:
This fixes the issues discussed in the previous document. In particular note that I have split the ROS architecture docs into 2 so that it is clear how ROS1 case works.

There are some open things the might be further clarified:
1. What `px4_ros_com` actually does - I say: *package provides some of the needed infrastructure for generating messages and headers needed by ROS2.*
2. Need to confirm what special needs to happen to build the client side firmware for use with ROS.

I'm happy to merge this whenever. @TSC21 ping me when you have reviewed.


en/middleware/micrortps.md
```diff
@@ -1,34 +1,39 @@
...
+RTPS has been adopted as the middleware for the ROS2 (Robot Operating System).
 The bridge allows us to better integrate with ROS2, making it easy to share sensor values, commands, and other vehicle information.
 
-This topic describes the bridge architecture, how it is compiled, and how to:
-1. Write a simple *Fast RTPS* application to subscribe to PX4 changes;
-1. For ROS2/ROS1 applications, understand how the `px4_ros_com` package works and how to use it to bridge ROS nodes with PX4.
+This topic describes the RTPS bridge architecture (and how it is used within the ROS2/ROS application pipeline).
+It also shows how to compile needed code to:
+1. Write a simple *Fast RTPS* application to subscribe to PX4 changes
+1. Connect ROS2 and ROS1 nodes with PX4 (via the RTPS Bridge, and using the `px4_ros_com` package)
```
hamishwillee, line 14: Note, this changed. Basically you can now read both points separately and they both make sense.

en/middleware/micrortps.md
```diff
@@ -45,18 +50,22 @@ The main elements of the architecture are the client and agent processes shown i
   In a typical configuration they will both be on the same system (e.g. a development computer, Linux companion computer or compute board), connected to the *Client* over a Wifi link or via USB.
 
 
-## Architectural overview for a ROS2/ROS application pipeline
+### ROS2/ROS application pipeline
+
```
hamishwillee, line 54: FYI, this had major change to make it clear what the pipeline looks like in both cases.

en/middleware/micrortps.md
```diff
@@ -45,18 +50,22 @@ The main elements of the architecture are the client and agent processes shown i
...
+The application pipeline for ROS2 is very straightforward! 
+Because ROS2 uses DDS/RTPS as its native communications middleware, you can create a ROS2 listener or advertiser node to publish and subscribe to uORB data on PX4, via the *PX4 Fast RTPS Bridge*. This is shown below.
+
+> **Note** You do need to make sure that the message types, headers and source files used on both client and agent side (and consequently, on the ROS nodes) are generated from the same Interface Description Language (IDL) files. The `px4_ros_com` package provides some of the needed infrastructure for generating messages and headers needed by ROS2.
```
hamishwillee: @TSC21 This needs review in particular. Ie confirm that this is what px4_ros_com does

en/middleware/micrortps.md
```diff
@@ -73,25 +82,24 @@ The *Agent* must be separately/manually compiled for the target computer.
...
+The [px4_ros_com](https://github.com/PX4/px4_ros_com) package, when built, generates everything needed to access PX4 uORB messages from a ROS2 node (for ROS1 you also need [ros1_bridge](https://github.com/ros2/ros1_bridge)).
+This includes all the required components of the *PX4 RTPS bridge, including the IDL files (required by the `micrortps_agent`), the `micrortps_agent` itself and the sources and headers of the ROS messages.
 
 The package has two separate branches:
 - a `master` branch, used with ROS2. It contains code to generate all the required ROS2 messages and IDL files to bridge PX4 with ROS2 nodes.
-- `ros1` is used with ROS(1). It contains code to generate the ROS message headers and source files, which can be used *with* the `ros1_bridge` to share data between PX4 and ROS.
+- a `ros1` branch, used with ROS(1). It contains code to generate the ROS message headers and source files, which can be used *with* the `ros1_bridge` to share data between PX4 and ROS.
```
hamishwillee, line 92: After this might be good to add a note at the stating what about building the client for ROS?

en/middleware/micrortps.md
```diff
@@ -230,13 +238,13 @@ In order to install ROS Melodic and ROS2 Bouncy on a Ubuntu 18.04 machine, follo
 1. [Install ROS Melodic](http://wiki.ros.org/melodic/Installation/Ubuntu)
 2. [Install ROS2 Bouncy](https://index.ros.org/doc/ros2/Linux-Install-Debians/)
 
-For the ROS2 case, so the package properly generates the IDL files, the following should be installed:
+Install the following component to ensure that the package properly generates the IDL files:
```
hamishwillee, line 241: FYI, I removed for "ROS2" case because this should also be done for ROS1 right?

en/middleware/micrortps.md
```diff
@@ -215,13 +225,11 @@ Check the **Building the `px4_ros_com` package** for details about the build str
 
 ## Building the `px4_ros_com` package
 
-As aforementioned, the `px4_ros_com` comes bundled with two branches, where one links the ROS2 with the PX4-agent/client bridge and the other allows, through `ros1_bridge`, to link the ROS framework with the ROS2 framework through the same set of messages. 
-Therefore, both branches need to be cloned separately so the build process can happen correctly. 
-Before that can happen, one requires to install and setup both ROS2 and ROS environments on its machine.
+Before building `px4_ros_com` you will first need to clone the PX4 Firmware repository
+(this is normally placed at the same tree level as the ROS workspaces).
+Then install and setup both ROS2 and ROS environments on your development machine
```
hamishwillee, line 230: @TSC21 Do we need to install the ros1 branch if we just want to use ROS2?

en/middleware/micrortps.md
TSC21: The client side is built the same way for a standart RTPS app or for a ROS2 app.

TSC21: Also, maybe it makes sense to have a sinlr way of identifying ROS 1. There's ROS1 in some places, ROSv1 on others. Maybe initially we should state that ROS v1 is from now on, called only ROS (because that's how it is still called, even now that you have ROS2).

en/middleware/micrortps.md
TSC21: No we don't need the ROS1 branch if you are just going to use the ROS2 middleware.

en/middleware/micrortps.md
TSC21: Well not exactly. The ROS2 side is the only one that needs the IDL's generated (note it generates IDL's not only for fast-RTPS but also for other RTPS protocols as openSplice). And the reason is that the micrortps_agent requires those. The only thing that you require generating for ROS1 is the msg headers and sources. Though of course you still need ROS2 (and ros1_bridge) to be able to bridge with the PX4 client.

TSC21 (commented):
Changes addressed. Looks really good thanks! Just one or two details and it's good to go!
Oh and great job splitting the diagram :)

en/middleware/micrortps.md
hamishwillee: Thanks for clarification. The text is still IMO better, so not planning to change.

en/middleware/micrortps.md
hamishwillee: It seems to me that you really want to just call it ROS, so that is what I have done (what I would like to have done is just always referred to it as ROS1).

PR #645, opened 2018-11-12
Add a page to document how to identify the most common issues on Flight Review

bkueng:
This documents the most common issues that I have come across when looking at different logs. It explains how to identify the problems by explaining how a plot should look like and providing examples, and how to fix the issue(s).

@hamishwillee can you have a look, I'm sure you have some good input here.

I'm sure we can add more, but I think this is a good start.
I will probably directly link to this page from Flight Review.


hamishwillee:
@bkueng I'm reviewing/updating this now. It is EXCELLENT and much needed. But it should be in the User Guide.

I'm also going to try organise by plot with example, rather than by good, bad, really bad for all the charts. The reason is that people need to understand each plot to see what it is doing. On the other hand, it might not work if you need to use multiple charts to analyse what is going on with vibration?

en/log/flight_review.md
```diff
@@ -0,0 +1,332 @@
...
+- Mouse scrolling on a particular plot axis zooms that axis (horizontally or vertically).
+- Mouse scrolling inside the plot zooms both axes.
+
+## PID Tracking Performance {#tracking}
+
+Depending on the flight mode, the vehicle controllers may attempt to track position, velocity, altitude or rate setpoints (the tracked setpoints depend on the mode, e.g.: in Stabilized mode there is no velocity setpoint).
+
+The **Estimated** line (red) should closely match with the **Setpoint** (green).
+If they do not, in most cases the PID gains of that controller need to be tuned.
+
```
hamishwillee, line 26: It would be good to have one or two graph images here. Maybe a "good track" and an unacceptable overshoot. Just to give a flavour. You could reuse image from the tuning guide.

en/log/flight_review.md
```diff
@@ -0,0 +1,332 @@
...
+This graph shows a frequency plot for the roll, pitch and yaw axis based on the actuator controls signal (the PID output from the rate controller). 
+It helps to identify frequency peaks and configuring the software filters.
+There should only be a single peak at the lowest end up to a few Hz, the rest should be low and flat.
+
+Note that the y-axis scaling is different for different vehicles, but logs from the same vehicle can be directly compared to each other.
+
+#### Examples: Good Vibration
+
+[QAV-R 5" Racer](https://docs.px4.io/en/frames_multicopter/qav_r_5_kiss_esc_racer.html) frame (excellent vibration).
+Note that this allows us to considerably increase the cutoff frequency of the software filters
```
hamishwillee: This comment was in original version. I have just put it on first mention. 

What does this mean? When do we have to increase the frequency and when decrease it. I would have assumed we would want to decrease the frequency cutoff to the lowest frequency reasonable. We could also put the param that does this here as a link.

en/log/flight_review.md
```diff
@@ -0,0 +1,332 @@
...
+> **Tip** It is worth looking at multiple charts when analysing vibration (different charts can better highlight some issues).
+
+
+### Actuator Controls FFT
+
+> **Note** You need to enable the high-rate logging profile ([SDLOG_PROFILE](../advanced/parameter_reference.md#SDLOG_PROFILE)) to see this plot.
+
+This graph shows a frequency plot for the roll, pitch and yaw axis based on the actuator controls signal (the PID output from the rate controller). 
+It helps to identify frequency peaks and configuring the software filters.
+There should only be a single peak at the lowest end up to a few Hz, the rest should be low and flat.
```
hamishwillee: > lowest end up to a few Hz,

What is a few Hz? The last image in "good" is 20Hz, so is that reasonable? Should we say or give a reasonable range?

en/log/flight_review.md
```diff
@@ -0,0 +1,332 @@
...
+
+Note that the y-axis scaling is different for different vehicles, but logs from the same vehicle can be directly compared to each other.
+
+#### Examples: Good Vibration
+
+[QAV-R 5" Racer](https://docs.px4.io/en/frames_multicopter/qav_r_5_kiss_esc_racer.html) frame (excellent vibration).
+Note that this allows us to considerably increase the cutoff frequency of the software filters
+
+![Low vibration QAV-R 5 Racer - FFT plot](../../assets/flight_log_analysis/flight_review/vibrations_good_actuator_controls_fft.png)
```
hamishwillee, line 68: MC_DTERM_CUTOFF higher here than on the F450. Why?

en/log/flight_review.md
```diff
@@ -0,0 +1,332 @@
...
+
+DJI F450 frame (good vibration).
+Note the blade passing frequency of the propellers around 100 Hz.
```
hamishwillee: I've noted it but I don't know what it means. Does  this mean we don't need to worry about it?

en/log/flight_review.md
```diff
@@ -0,0 +1,332 @@
...
+
+![Low vibration DJI F450 - FFT plot](../../assets/flight_log_analysis/flight_review/vibrations_f450_actuator_controls_fft.png)
+
+S500 frame:
+
+![Low vibration S500 actuator controls - FFT plot](../../assets/flight_log_analysis/flight_review/vibrations_s500_actuator_controls_fft.png)
```
hamishwillee, line 78: Do we need to think about the roll/pitch/yaw mean values in images?

en/log/flight_review.md
```diff
@@ -0,0 +1,332 @@
...
+
+> **Note** While the plot above looks good, the [Raw Acceleration graph for the same flight](#raw_acc_s500) shows that the vibration levels are a bit high for x and y.
+  This is a good example of why it is worth checking several graphs!
+
+#### Examples: Bad Vibration
+
+This example shows a peak in frequency close to 50 Hz. 
+In this case the cause was a landing gear that had a bit of play.
```
hamishwillee: Does this mean that the landing gear wasn't attached firmly?

en/log/flight_review.md
```diff
@@ -0,0 +1,332 @@
...
+S500 frame: <!-- the Actuator Controls FFT plot looks good, whereas the vibration levels are a bit high for x and y in the Raw Acceleration plot (which is typical for an S500 airframe). This is at the limit where it starts to negatively affect flight performance.
+GOOD or bad?
+-->
+
+![Vibration S500 - spectral density plot](../../assets/flight_log_analysis/flight_review/vibrations_s500_spectral.png)
+
+
+#### Examples: Bad Vibration
+
+? <!-- In this log the vibration levels are too high, apparent in the Raw Acceleration plot - the graph of the z-axis overlaps with the x/y-axis graph: -->
```
hamishwillee: What can we say here? Just that it has quite a lot of yellow across a large range of frequencies?

en/log/flight_review.md
```diff
@@ -0,0 +1,332 @@
...
+
+
+![High vibration in spectral density plot](../../assets/flight_log_analysis/flight_review/vibrations_too_high_spectral.png)
+
+This example shows a peak in frequency close to 50 Hz.
```
hamishwillee: So is it the fact that this is up around 50Hz that makes it a problem? Why is this a problem but not the 100Hz propeller stuff you pointed to above?

en/log/flight_review.md
```diff
@@ -0,0 +1,331 @@
...
+[QAV-R 5" Racer](https://docs.px4.io/en/frames_multicopter/qav_r_5_kiss_esc_racer.html) frame (excellent vibration).
+
+![Low vibration QAV-R 5 Racer - spectral density plot](../../assets/flight_log_analysis/flight_review/vibrations_good_spectral.png)
+<!-- https://logs.px4.io/plot_app?log=cd88b091-ec89-457c-85f6-e63e4fa0f51d -->
+
+DJI F450 frame (good vibration).
+Note the blade passing frequency of the propellers around 100 Hz.
+![Low vibration DJI F450 - spectral density plot](../../assets/flight_log_analysis/flight_review/vibrations_f450_spectral.png)
+
+S500 frame: <!-- the Actuator Controls FFT plot looks good, whereas the vibration levels are a bit high for x and y in the Raw Acceleration plot (which is typical for an S500 airframe). This is at the limit where it starts to negatively affect flight performance.
```
hamishwillee: NOte commented out text. Anything special we should say here?

en/log/flight_review.md
```diff
@@ -0,0 +1,331 @@
...
+Solutions and steps to reduce vibrations include:
+- Make sure everything is firmly attached on the vehicle (landing gear, GPS mast, etc.)
+- Use balanced propellers.
+- Make sure to use high-quality components for the propellers, motors, ESC and airframe.
+  Each of these components can make a big difference.
+- Use a vibration-isolation method to mount the autopilot.
+- As a *last* measure, adjust the software filters (see [here](https://docs.px4.io/en/config_mc/racer_setup.html#filters)).
+  It is better to reduce the source of vibrations, rather than filtering them out in software.
+
+<!-- TODO: write a separate vibration setup page in more depth, move some of this there and link to it from here -->
```
hamishwillee, line 199: Great. Should link to / from https://docs.px4.io/en/assembly/mount_and_orient_controller.html#vibration-damping

en/log/flight_review.md
```diff
@@ -0,0 +1,331 @@
...
+
+
+## Actuator Outputs
+
+The *Actuator Outputs* graph shows the signals that are sent to the individual actuators (motors/servos).
+Generally it is in the range between the minimum and maximum configured PWM values (e.g. from 1000 to 2000).
+
+This is an example for a quadrotor where everything is OK (all of the signals are within the range, approximately overlap each other, and are not too noisy):
```
hamishwillee, line 207: I updated this. Is this a good description of why this is a good plot?

en/log/flight_review.md
```diff
@@ -0,0 +1,331 @@
...
+![Good actuator outputs](../../assets/flight_log_analysis/flight_review/actuator_outputs_good.png)
+
+The plot can help to identify different problems:
+- If one or more of the signals is at the maximum over a longer time, it means the controller runs into **saturation**.
```
hamishwillee, line 211: Image of bad case might be nice.

hamishwillee:
@bkueng OK, ready for your next update/review. I think the restructure of vibration section by chart works - makes it much easier to compare good to bad for a particular chart. I made sure there are notes that state you probably should look at all the charts that are useful for vibration, not just this one.

When we're happy, lets move this to the user guide. Probably in https://docs.px4.io/en/advanced_config/? My logic is that it isn't just people developing new hardware that might want to do this - it is anyone building a frame. 

Generally I'm trying to get as much as reasonable from dev guide into user guide. At some point might even make the developer guide a sub-section of the user guide.

en/log/flight_review.md
```diff
@@ -0,0 +1,331 @@
+# Flight Review 101
+
+The [Flight Review](http://logs.px4.io) plots can be used to analyze vehicle condition in a specific flight.
```
bkueng: I was trying to clarify 2 point here:
- a log provides information about the specific flight
- but it also shows important information about the general vehicle condition (independent from that flight)

Not sure how important that seems to you or how to phrase it best.

en/log/flight_review.md
bkueng: We can do that, but I wanted to avoid duplication and explaining the same thing twice.

en/log/flight_review.md
bkueng: Yes 20 is reasonable, and we can write that.

en/log/flight_review.md
bkueng: See https://docs.px4.io/en/config_mc/racer_setup.html#filters. We want to increase the cutoff frequency as much as possible.

en/log/flight_review.md
bkueng: higher=better, see above.
On the F450, the defaults are used, and there was no filter tuning done.

en/log/flight_review.md
bkueng: I've put it there to explain what we see in the plot. It's generally not an issue, but it can be if they are too strong.
If remarked, it should be next to the Spectral plot.

en/log/flight_review.md
bkueng: If you want we can explain it: they should be a help when comparing logs with different filter settings.

en/log/flight_review.md
bkueng: Exactly.

en/log/flight_review.md
bkueng: No, this looks good, comparable to the one from the F450.

en/log/flight_review.md
bkueng: Yes, it's what I wrote above: the strong yellow lines around 100 Hz can be an issue, but they don't have to be. In that case they are, but it's very difficult to tell from that plot alone.

en/log/flight_review.md
bkueng: The problems are:
- it's a single frequency, and it's very strong
- the problem is that it's quite low, close to the vehicle dynamics: with the default filter settings of 80 Hz, the 100 Hz are filtered out (or at least dampened), but not vibrations at 50 Hz.

en/log/flight_review.md
bkueng: Ok, that section is very minimal though.

en/log/flight_review.md
bkueng: :+1:

en/log/flight_review.md
bkueng: Agreed, but I didn't find one :/

en/log/flight_review.md
```diff
@@ -0,0 +1,331 @@
...
+![Estimator watchdog with magnetometer problems](../../assets/flight_log_analysis/flight_review/estimator_watchdog_mag_problem.png)
+
+
+## Sampling Regularity of Sensor Data
+
+The sampling regularity plot provides insights into problems with the logging system and scheduling.
+
+The amount of **logging dropouts** start to increase if the log buffer is too small, the logging rate is too high or a low-quality SD card is used.
+
+> **Note** Occasional dropouts *are* expected.
```
bkueng: Good SD cards don't have them (or at least only in very rare cases), so I'd change it back to 'can be expected'.

bkueng (commented):
Thanks @hamishwillee for the review.

> I'm also going to try organise by plot with example, rather than by good, bad, really bad for all the charts. The reason is that people need to understand each plot to see what it is doing. On the other hand, it might not work if you need to use multiple charts to analyse what is going on with vibration?

This looks good and it makes sense. But as you write it separates plots coming from the same log/example. That's why I put them together initially, because often it helps looking at all of the plots to get the full picture.
We can't have both I guess, and I'm not sure which one is better. Maybe numbering them would help?

> My logic is that it isn't just people developing new hardware that might want to do this - it is anyone building a frame.

Right, that's good. I've put it here because Flight Log Analysis is already here.

en/log/flight_review.md
hamishwillee: Because of the context I interpreted  "can be expected" reads as "any card will have some dropouts". I was slightly unsure though, because you could have actually meant to say "Occasional dropouts are acceptable"

I've changed this to (OK?):

> **Note** Occasional dropouts can be expected on medium quality cards.

hamishwillee:
>> I'm also going to try organise by plot with example, rather than by good, bad, really bad for all the charts. The reason is that people need to understand each plot to see what it is doing. On the other hand, it might not work if you need to use multiple charts to analyse what is going on with vibration?

> This looks good and it makes sense. But as you write it separates plots coming from the same log/example. That's why I put them together initially, because often it helps looking at all of the plots to get the full picture.

I'm much happier with the current organisation - it makes it much easier for me to work out what is relevant in a particular chart than the previous case. 
I also added several notes that it is "good to view all the charts together" too.

> We can't have both I guess, and I'm not sure which one is better. Maybe numbering them would help?

Sure we can. As an experiment I have added a section at the end with the graphs sorted by flight. Note that you will need to check the text, but I hope you can put your analysis in this structured format (ie "what does each plot show" in order with bulletpoints.

en/log/flight_review.md
hamishwillee: Yes, we probably need a heap more than that. This was really a note to me.

en/log/flight_review.md
hamishwillee: That sounds like a useful thing to say. I have changed to:

> The strong yellow lines at around 100Hz indicate a potential issue that requires further investigation (starting with a review of the other charts).

en/log/flight_review.md
hamishwillee: The fact that a log provides information about a specific flight is IMO obvious. 
A slight change better captures the other idea though. How about:

> The [Flight Review](http://logs.px4.io) plots for a flight be used to analyze general vehicle condition.

hamishwillee: (I changed it like this already)

en/log/flight_review.md
hamishwillee: Fair enough. Personally I'm OK with 1 or 2 images here - purpose isn't to explain so much as position people in the right charts "mentally".

I'll set resolved and you decide if you want to add or not.

en/log/flight_review.md
hamishwillee: Awesome. I changed to:

> There should only be a single peak at the lowest end (below around 20 Hz), the rest should be low and flat.

en/log/flight_review.md
hamishwillee: Thanks for the explanation. I can see that we want to increase the filter cutoff frequency to reduce the control latency. We don't want to increase too far because this results in more noise.

But the statement is still confusing. See below - what is it about  "THIS" plot that means we can increase the cutoff frequency? Is it just that the initial bump is slow low, or is that the rest is flat, or that the signal isn't noisy?

> Note that **this** allows us to considerably increase the cutoff frequency of the software filters

en/log/flight_review.md
hamishwillee: OK, thanks. I don't think we need to mention this then because defaults were used. If this had been tuned and the higher MC_DTERM_CUTOFF was because of better vibration behaviour then it would have been.

en/log/flight_review.md
hamishwillee: Cheers. I have removed from here and just put near the spectral plot:

> **Note** Above you can see the blade passing frequency of the propellers at around 100 Hz.

It would be good if we could explain how to tell if it is too strong. But assume not needed. Setting resolved.

en/log/flight_review.md
hamishwillee: Entirely up to you - I don't know how useful this is.

en/log/flight_review.md
hamishwillee: Thanks, I slightly reworded those "(in this case due to "loose" landing gear)."

en/log/flight_review.md
hamishwillee: Thanks. I've been a dumbass, reading filter as "High pass" rather than "low pass". All the explanations make a lot more sense now ([insert palm smack emoticon]).

I have added a Tip below this:

> **Tip** This indicates a possible problem because it is a strong single low frequency that is close to the vehicle dynamics.
  With the default filter settings of 80 Hz vibrations at 50 Hz will not be filtered.

hamishwillee: Note, this is generally what you want:
- Look, see what the plot shows
- What might cause this
- Why is it a problem

hamishwillee:
>> My logic is that it isn't just people developing new hardware that might want to do this - it is anyone building a frame.

> Right, that's good. I've put it here because Flight Log Analysis is already here.

Move that too then - to me this looks useful to all users. Maybe put in a folder analysis, and actually also have a top level Flight Analysis entry point.

In Devguide, have a near empty topic that links to these two.

I can do this if you want?

Anyway, done on this for today.

en/log/flight_review.md
bkueng: The reason is the low vibrations of the complete setup. The consequence is that the rest is flat in this plot.
From that plot you can essentially only see if the cutoff frequency is too high, because if you have the filter settings too low, either there are no vibrations or they are filtered out.

en/log/flight_review.md
bkueng: It's a bit less useful than I initially thought. I don't think we need to mention it here, maybe I'll drop it from Flight Review as well.

en/log/flight_review.md
bkueng: > I've been a dumbass, reading filter as "High pass" rather than "low pass". All the explanations make a lot more sense now ([insert palm smack emoticon]).

Haha. No it's just that this is an advanced topic.

bkueng:
> Sure we can. As an experiment I have added a section at the end with the graphs sorted by flight. Note that you will need to check the text, but I hope you can put your analysis in this structured format (ie "what does each plot show" in order with bulletpoints.

Cool, I'll need to look at this tomorrow, I'm away today.

> I can do this if you want?

If you want to do it today, please go ahead. Otherwise I can do it as well.

en/log/flight_review.md
hamishwillee: OK. Ignored :-)

hamishwillee:
> If you want to do it today, please go ahead. Otherwise I can do it as well.

@bkueng Done - see https://github.com/PX4/px4_user_guide/pull/379

hamishwillee:
Note, I'll clean up Devguide for moved Flight Analysis topic when the other issue has merged.

Issue #650, opened 2018-11-14
MAVLink Messaging - update to MAVLink 2


hamishwillee:
[MAVLink Messaging · PX4 Developer Guide](http://dev.px4.io/en/middleware/mavlink.html) needs update work.

From: https://github.com/PX4/Devguide/pull/649#issue-230708462

1. It is based on MAVLink 1.0 and needs updates to 2.0? I THINK that is just a matter of changing the links. 
2. This starts at the point you already have a library with the new message available. However it should start a bit earlier. Specifically, what is the process for me to add my own new generated library into a build? Is it just a matter of building the library and dumping it into Firmware/mavlink/include/mavlink/ ?
   - My reasoning is that if the message is custom, you aren't getting this into the standard library. It isn't necessarily obvious the "right" way to get the message into PX4 for testing.

@bkueng I'm assigning to you, but you can push back to me with a few notes on how this would be done if you like!

hamishwillee:
Also https://github.com/PX4/Devguide/issues/538

bkueng:
> It is based on MAVLink 1.0 and needs updates to 2.0? I THINK that is just a matter of changing the links.

Correct. And the generator needs to be configured to output mavlink 2.0.

> Specifically, what is the process for me to add my own new generated library into a build? Is it just a matter of building the library and dumping it into Firmware/mavlink/include/mavlink/ 

Correct, once you have generated the lib, just copy everything over to `Firmware/mavlink/include/mavlink/v2.0`.

>  My reasoning is that if the message is custom, you aren't getting this into the standard library. It isn't necessarily obvious the "right" way to get the message into PX4 for testing.

Making it available to others to test is a bit of a hassle. You would do:
- fork https://github.com/mavlink/c_library_v2 (for simplicity, it's just as valid to create an entirely new repo)
- add your generated lib there and push the changes
- let others know to update the submodule to the changed fork.

> @bkueng I'm assigning to you, but you can push back to me with a few notes on how this would be done if you like!

Going for the notes then.
Is it clear?

hamishwillee:
Thanks @bkueng . I've added this to #652

We could spell out how to update a submodule. Can we assume people know how to do this? I didn't until last week.
(I assume it is just go into the submodule directory, use git as usual to set remotes, fetch branch on remote, checkout branch on remote)

bkueng:
> (I assume it is just go into the submodule directory, use git as usual to set remotes, fetch branch on remote, checkout branch on remote)

Yes

PR #652, opened 2018-11-15
Testing custom MAVLink messages

hamishwillee:
@bkueng This adds note about testing custom uorb messages. Suggestion is that you can look in uLog and rebuild QGC and use mavlink inspector. I have not actually done these things, so can you:
- Confirm this is good advice, that you think is clear enough to follow
- Suggest better alternative

I presume best way to test an incoming message  can't really be documented because it will depend on the message. Or could we suggest look at uLog if there is one associated with the incoming message?

### DIFF
--- a/en/middleware/mavlink.md
+++ b/en/middleware/mavlink.md
@@ -111,9 +111,12 @@ MAVLink channel on UDP port 14556):
 mavlink stream -r 50 -s CA_TRAJECTORY -u 14556
 ```
 
-> **Tip** You can use uLog [Logging](../log/logging.md) to verify that your uORB message is being emitted correctly. 
-  In order to verify that the message is being sent "end to end" you will need to [Rebuild QGroundControl](https://dev.qgroundcontrol.com/en/getting_started/) with your MAVLink library, 
-  and then verify that the message is received using [MAVLink Inspector Widget](https://docs.qgroundcontrol.com/en/app_menu/mavlink_inspector.html) (or some other MAVLink tool).
+> **Tip** You can use the `uorb top [<message_name>]` command to verify in real-time that your message is published and the rate (see [uORB Messaging](../middleware/uorb.md#uorb-top-command)). 
+  This approach can also be used to test incoming messages that publish a uORB topic (for other messages you might use `printf` in your code and test in SITL).
+>
+> To see the message on *QGroundControl* you will need to [build it with your MAVLink library](https://dev.qgroundcontrol.com/en/getting_started/),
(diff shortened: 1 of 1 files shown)


hamishwillee:
@bkueng Slightly unrelated, but example shows how to set up code to stream a message. Is all outgoing traffic handled this way (ie via stream)? What about commands etc - are they sent out as part of a stream or ad-hoc. 
To put it another way, do you think the doc covers all the standard sending scenarios, or just one?

bkueng:
> Confirm this is good advice, that you think is clear enough to follow

I think it's better to not add ulog, as it's another component you need to configure.
It's much simpler to use `uorb top [<message_name>]`, then you see in real-time if your message is published, and even at what rate.

> I presume best way to test an incoming message can't really be documented because it will depend on the message.

Yes, but the most common case is a streamed message. If it is published as a uORB topic, you can also use `uorb top`. Otherwise via printf (and test in SITL first).

> @bkueng Slightly unrelated, but example shows how to set up code to stream a message. Is all outgoing traffic handled this way (ie via stream)? What about commands etc - are they sent out as part of a stream or ad-hoc.
To put it another way, do you think the doc covers all the standard sending scenarios, or just one?

I'd say streaming is the most important and most common one, so I'd leave it with that. Commands need to be acked, so they are not streamed but handled separately. You could explain that if you want to though. For a new command most of it is automatic, since it's already converted into an uORB topic `vehicle_command`.

hamishwillee:
Thanks @bkueng . I've updated with your advice to use `uorb top`. I left in comment about QGC because it might not be obvious if they work end to end that it also needs to be updated with new mavlink library.

Can you check it again please?

hamishwillee:
> I'd say streaming is the most important and most common one, so I'd leave it with that. Commands need to be acked, so they are not streamed but handled separately. You could explain that if you want to though. For a new command most of it is automatic, since it's already converted into an uORB topic vehicle_command.

I would if I "really" understood what needs to be done.

 I'm guessing from what you're saying any new command gets turned into a `vehicle_command` uorb. Presumably any system that wants to handle a particular command listens for that and processes the command. Lots of questions though - what if multiple parts of the system try to handle a command? How is the ack processed/sent by the code that is processing the message? etc etc.

hamishwillee:
... and presumably on that note, we should also have a section on handling new MAV_CMDs in missions.

hamishwillee:
@bkueng I've further added updated to MAVLink 2 as per https://github.com/PX4/Devguide/issues/650

bkueng (approved):
Ok, let's merge this now and do follow-ups as needed.

PR #657, opened 2018-11-20
Update parameter doc with API and metadata info

### LINKED ISSUE #532: [Feature Request] Guide to Adding Params
Hi PX4-devs,

Could you please add a guide to adding custom parameters within PX4?

From the parameter and configs page, the most I can intuit is that I can add new parameters in each module's or driver's *_params.c file (including meta data) using the format below.

image

Can the team confirm this? Could it please be added to the dev guide?

Thanks! - Marc

### LINKED ISSUE #644: Parameters & Configs  - Not enough to create my own
[Parameters & Configs](https://dev.px4.io/en/advanced/parameters_and_configurations.html) is supposed to explain how to work with parameters. IMO it is not enough to do so.

Specifically we need 
* [Parameter Metadata](https://dev.px4.io/en/advanced/parameters_and_configurations.html#parameter-meta-data)
  * What file should these be declared in - where should it be located, file extensions or naming conventions etc, should it be per module or per library or what?
  * What are naming conventions for individual parameters?
  * What types of params are supported - e.g. `PARAM_DEFINE_FLOAT`. Where is API for this?
  * HOw are the params imported into a module/modules?
* How can their values be accessed and set from C++ code. 

I also looked at the [Module Template](https://dev.px4.io/en/apps/module_template.html) which states that it shows "Accessing parameters and reacting to parameter updates.". HOwever I can't see the parameters [being defined with metadata](https://github.com/PX4/Firmware/blob/master/src/templates/module/module.h#L77).

I would consider this sufficiently documented if I could have just sent this page link as a response to this comment: 
https://github.com/PX4/Firmware/issues/10822#issuecomment-437673911

@bkueng Can you help me by answering some of the above? (or editing the page appropriately)?

hamishwillee:
This adds improved parameter docs, including info on the C and C++ APIs and the param metadata. 
Fixes #532
Fixes #644

@bkueng This is a good first draft - meaning I think it is correct, but I'm pretty tired so I will read it again tomorrow. It is worth your while checking for accuracy and omission.

### DIFF
--- a/en/advanced/parameters_and_configurations.md
+++ b/en/advanced/parameters_and_configurations.md
@@ -38,6 +38,8 @@ param show -c
 
 ### Exporting and Loading Parameters
 
+You can save any parameters that have been *touched* since all parameters were last reset to their firmware-defined defaults (this includes any parameters that have changed been changed, even if they have been changed back to their default).
+
 The standard `param save` command will store the parameters in the current default file:
 ```sh
 param save
@@ -48,23 +50,28 @@ If provided with an argument, it will store the parameters instead to this new l
 param save /fs/microsd/vtol_param_backup
 ```
 
-There are two different commands to load parameters: 
-- `param load` replaces all the current parameters, duplicating the state when the parameters were stored.
-   Parameters are not automatically saved by this operation (it is assumed that they are already saved).
-- `param import` only updates parameter values that have been *changed from the default*.
-  The parameters are automatically saved to the default location when loading.
-  
-  This can be used, for example, to just import calibration data without overwriting the rest of the system configuration (i.e. first calibrate a board that is otherwise unconfigured and save the parameters; this file can then be imported on another system to *just* update the calibration parameters).
-  
-Both loading options are shown below:
-```sh
-# Overwrite the current parameters
-param load /fs/microsd/vtol_param_backup
-param save  # Params not automatically saved on load
+There are two different commands to *load* parameters: 
+- `param load` first does a full reset of all parameters to their defaults, and then overwrites parameter values with any values stored in the file.
+- `param import` just overwrites parameter values with the values from the file and then saves the result (i.e. effectively calls `param save`).
 
-# Merge the current parameters with stored non-default parameters
+The `load` effectively resets the parameters to the state when the parameters were saved (we say "effectively" because any parameters saved in the file will be updated, but other parameters may have different firmware-defined default values than when the parameters file was created). 
+
+By contrast, `import` merges the parameters in the file with the current state of the vehicle.
+This can be used, for example, to just import a parameter file containing calibration data, without overwriting the rest of the system configuration.
+
+Examples for both cases are shown below:
+
+```sh
+# Reset the parameters to when file was saved
+param load /fs/microsd/vtol_param_backup
+# Optionally save params (not done automatically with load)
+param save
+```
+```sh
+# Merge the saved parameters with current parameters
 param import /fs/microsd/vtol_param_backup  
 ```
+
 
 ## Parameter Names
 
--- a/en/advanced/parameters_and_configurations.md
+++ b/en/advanced/parameters_and_configurations.md
@@ -1,23 +1,25 @@
 # Parameters & Configurations
 
-The PX4 platform uses the param subsystem (a flat table of float and int32_t values) and text files (for mixers and startup scripts) to store its configuration.
+PX4 uses the *param subsystem* (a flat table of `float` and `int32_t` values) and text files (for mixers and startup scripts) to store its configuration.
 
-The [system startup](../concept/system_startup.md) and how [airframe configurations](../airframes/adding_a_new_frame.md) work are detailed on other pages. This section discusses the param subsystem in detail
+This section discusses the *param* subsystem in detail.
+It covers how to list, save and load parameters, and how to define them.
 
-## Command Line usage
+> **Note** [System startup](../concept/system_startup.md) and the way that [airframe configurations](../airframes/adding_a_new_frame.md) work are detailed on other pages. 
 
-The PX4 [system console](../debug/system_console.md) offers the *param* tool, which allows to set parameters, read their value, save them and export and restore to and from files.
+
+## Command Line Usage
+
+The PX4 [system console](../debug/system_console.md) offers the [param](../middleware/modules_command.md#param) tool, which can be used to set parameters, read their value, save them, and export and restore to/from files.
 
 ### Getting and Setting Parameters
 
-The param show command lists all system parameters:
-
+The `param show` command lists all system parameters:
 ```sh
 param show
 ```
 
-To be more selective a partial parameter name with wildcard can be used:
-
+To be more selective, a partial parameter name with wildcard "*" can be used:
 ```sh
 nsh> param show RC_MAP_A*
 Symbols: x = used, + = saved, * = unsaved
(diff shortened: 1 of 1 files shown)


en/advanced/parameters_and_configurations.md
```diff
@@ -29,52 +31,189 @@ x   RC_MAP_ACRO_SW [375,514] : 0
...
 param save /fs/microsd/vtol_param_backup
 ```
 
-There are two different commands to load parameters: `param load` will load a file and replace the current parameters with what is in this file, resulting in a 1:1 copy of the state that was present when these parameters were stored. `param import` is more subtle: It will only change parameter values which have been changed from the default. This is great to e.g. initially calibrate a board (without configuring it further), then importing the calibration data without overwriting the rest of the system configuration.
-
-Overwrite the current parameters:
-
+There are two different commands to load parameters: 
+- `param load` replaces all the current parameters, duplicating the state when the parameters were stored.
+   parameters are not automatically saved by this operation (it is assumed they are already saved).
```
bkueng: **P**arameters are not...

en/advanced/parameters_and_configurations.md
```diff
@@ -29,52 +31,189 @@ x   RC_MAP_ACRO_SW [375,514] : 0
...
+- `param import` only updates parameter values that have been *changed from the default*.
```
bkueng: Here we should be more accurate: the `import` loads all params that it finds in the file, overwriting the current value of these params. This means even if a parameter is set to default, and stored in the file, it will be set to the default when importing the file.
A params file typically contains all parameters that an aifrframe config or the user touched/changed (also if changed back to default).

en/advanced/parameters_and_configurations.md
```diff
@@ -29,52 +31,189 @@ x   RC_MAP_ACRO_SW [375,514] : 0
...
-```sh
-param import /fs/microsd/vtol_param_backup
-```
 
 ## C / C++ API
 
-There is also a C and a separate C++ API which can be used to access parameter values.
+There are separate C and C++ APIs that can be used to access parameter values from within PX4 modules and drivers.
+
+The most important difference between the APIs is that the C++ version has a more efficient standardized mechanism to synchronize with changes to parameter values (i.e. from a GCS).
```
bkueng: The C++ version has also better type-safety and less overhead in terms of RAM. The drawback is that the param name must be known at compile-time, whereas the C API is more dynamic, allowing to pass a (dynamically created) param name as string.

en/advanced/parameters_and_configurations.md
```diff
@@ -29,52 +31,189 @@ x   RC_MAP_ACRO_SW [375,514] : 0
...
+`parameter_update_sub` returned by `orb_subscribe()` is a handle we can use to refer to this particular subscription.
+```cpp
+# Subscribe to parameter_update message
+int parameter_update_sub = orb_subscribe(ORB_ID(parameter_update));
+...
+# Unsubscribe to parameter_update messages
+orb_unsubscribe(parameter_update_sub);
+```
+
+Call `parameters_update(parameter_update_sub, true);` periodically in code to check if there has been an update (this is boilerplate):
```
bkueng:
```suggestion
Call `parameters_update(parameter_update_sub);` periodically in code to check if there has been an update (this is boilerplate):
```

en/advanced/parameters_and_configurations.md
```diff
@@ -29,52 +31,189 @@ x   RC_MAP_ACRO_SW [375,514] : 0
...
 
-PX4 uses an extensive parameter meta data system to drive the user-facing presentation of parameters. Correct meta data is critical for good user experience in the ground station.
+> **Note** If `PARAM_NAME` was declared in parameter metadata then its default value will be set, and the above call to find the parameter should always succeed. 
 
-A typical parameter metadata section will look like this:
+`param_find()` is an "expensive" operation, which returns a handle that can be used by `param_get()`. 
+If you're going to read the parameter multiple times, you may cache the handle and use it in `param_get()` when needed
+```cpp
+# Get the handle to the parameter
+int32_t my_param_handle = 0;
```
bkueng:
```suggestion
param_t my_param_handle = PARAM_INVALID;
```

bkueng (commented):
Looks good, I remarked a few things that I noticed.

en/advanced/parameters_and_configurations.md
hamishwillee: Hi @bkueng That comment makes no sense to me at all - ie I can't see how that differs from `load`.

Let me see if I an deconstruct it ...

> A params file typically contains all parameters that an aifrframe config or the user touched/changed (also if changed back to default).

So basically a param file just contains _changed_ parameters. Presumably this means there is something in the system that notices something has been changed (as opposed to just comparing to a stored default value).

So when this is imported it will only update the parameters it contains (things that have been changed, even if they were changed back to defaults) . 

So import replaces all values. If a value was changed it will be updated even if it was changed back to default. 
1. What implication does that have? If it was originally a default value and it gets set to that same default value isn't that the same thing as not having changed the value in the first place. And if not, why? 
2.  The possible implication I can see is if the default value in file differs from the default value on vehicle then that parameter will be updated. But presumably the value will not be marked as a "default" even though it was in the original
3. Still can't see how import and load actually differ!


How does that differ from `load`? The only thing

en/advanced/parameters_and_configurations.md
hamishwillee: Thanks, I'll add this!

en/advanced/parameters_and_configurations.md
hamishwillee: @bkueng This comes from the module template, which takes a force parameter. I assume that for simplicity you are proposing we don't support forcing the value.
Thanks!

hamishwillee:
HI @bkueng 

Thanks for review. I tidied things up as per your suggestions, including removing the [forced update option](https://github.com/PX4/Devguide/pull/657/commits/6b34e67e3db86deba1c6a9abc91da413bdec5556#diff-1f6473deb29efaa6accccf5d51963ab2R145) 

If that is OK then only open issue is that I still don't "get" param `load` vs `import`: https://github.com/PX4/Devguide/pull/657#discussion_r235227680

en/advanced/parameters_and_configurations.md
bkueng: > What implication does that have? If it was originally a default value and it gets set to that same default value isn't that the same thing as not having changed the value in the first place. And if not, why?

There's a subtle case where it's not the same: if the default value in the Firmware changes, you will keep your 'default' if you ever set the param, and if not, you will use the updated Firmware default when flashing the new version.

> The possible implication I can see is if the default value in file differs from the default value on vehicle then that parameter will be updated. But presumably the value will not be marked as a "default" even though it was in the original

Maybe this helps: a params file does not have any notion of default. It just stores a list of param names with an associated value. All default/untouched params are simply not stored in that file.

> Still can't see how import and load actually differ!

You can see it as following: a `load` does a full reset to default first for all params, and then imports the params.

bkueng:
> Thanks for review. I tidied things up as per your suggestions, including removing the forced update option

I did not mean to remove the entire force update, though that works too. The force update exists for the following reason: if you have other variables that you need to initialize based on parameter values you can do that in `Module::parameters_update`. Then on startup you call `parameters_update` once with force=true, so that your variables get initialized as well.

en/advanced/parameters_and_configurations.md
hamishwillee: @bkueng A real challenge, but I now fully get this. Please check the updated version (the whole section updated).

hamishwillee:
Hi @bkueng Also fixed the force=True thingy. Should all be good now.

bkueng (approved):
You got it :) Simple command but tricky semantics.

hamishwillee:
Thanks @bkueng . I'm pleased this is in. It has been niggling at me that some of these "core essentials" weren't clear.

PR #661, opened 2018-11-23
Add alternative wifi dongle for video streaming - TL-WN821N

hamishwillee:
Anecdotally this is better - proposed by @AlexisTM in slack. 

I have just added it as a note because the rest of the tutorial is very specific. @AlexisTM, can you confirm that nothing else had to be changed in the configuration? Looking at this the instructions should actually work for any dongle right?

### DIFF
--- a/en/qgc/video_streaming.md
+++ b/en/qgc/video_streaming.md
@@ -6,6 +6,7 @@ The whole hardware setup is shown in the figure below. It consists of the follow
 * Odroid C1
 * Logitech camera C920
 * WiFi module TP-LINK TL-WN722N
+  > **Note** The [TP-LINK WN821N](https://www.conrad-electronic.co.uk/ce/en/product/995310/TP-LINK-TL-WN821N-WiFi-dongle-USB-20-300-Mbps) can also be used (without modification to the instructions). This is smaller and has a better range.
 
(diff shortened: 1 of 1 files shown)


AlexisTM:
I will ensure that we can do an AP with it (following the tutorial)

AlexisTM:
Badly, this module does not have AP capability. 

I am using it for the offboard mode.

hamishwillee:
@AlexisTM Thanks. So just to be clear, what you're saying is that this module can't be used for this purpose and I should close the PR?

Presumably you're also saying that you can set up an alternative configuration that can use this module? If so, is it worth adding sections to capture that alternative?

AlexisTM:
Yes, I am sorry of that. 

The module does not have the AP/ad-hoc capability so it does not fit this use-case.

Yet it is a good choice for a multi-vehicle setup in offboard mode.

hamishwillee:
Hi @AlexisTM ,

No need to be sorry - this time didn't work out, but another will.

If you think it is a good idea, perhaps you could add a page or section to https://docs.px4.io/en/telemetry/telemetry_wifi.html - I presume that this is the sort of use case you mean?

Cheers
Hamish

AlexisTM:
Kind of, the WiFi is in my case connected to the Onboard computer, not the Pixhawk.
I'll write a page once I find time :)

Issue #656, opened 2018-11-20
Add docs for posix dynamic modules on Posix + FIFOs


hamishwillee:
Support for dynamic modules on Posix (and possibly nuttx, untested) added in https://github.com/PX4/Firmware/pull/10870#issuecomment-439932956

Needs to state what it can be used for and how it is used.
* [ ] Possibly add a section in  [System Startup](http://dev.px4.io/en/concept/system_startup.html) linked from a small note in the [app tutorial](http://dev.px4.io/en/apps/hello_sky.html).

* [x] In addition, the section in [System Startup](http://dev.px4.io/en/concept/system_startup.html) about client/daemon FIFOs is no longer true. 

Hoping that @m-ou-se will create PR :-)

m-ou-se:
Updated documentation on the client/daemon communication in #662.

Will submit a PR for the dynamic modules as soon as https://github.com/PX4/Firmware/pull/10870 is merged.

m-ou-se:
The code is merged. #663 updates the documentation.

PR #662, opened 2018-11-23
Update description of daemon communication in 'System Startup'.

m-ou-se:
We no longer use FIFOs, but instead use a single Unix socket.

(As mentioned in #656.)


hamishwillee:
Thanks Mara. @bkueng I'm very happy with this, but leaving to you for technical review.

bkueng (approved):

PR #663, opened 2018-11-24
Mention loadable modules in system startup and the app tutorial.

### LINKED ISSUE #656: Add docs for posix dynamic modules on Posix + FIFOs

m-ou-se:
Fixes #656

### DIFF
--- a/en/apps/hello_sky.md
+++ b/en/apps/hello_sky.md
@@ -135,6 +135,10 @@ In this section we create a *minimal application* that just prints out `Hello Sk
    
    > **Tip** The `px4_add_module()` format is documented in [Firmware/cmake/common/px4_base.cmake](https://github.com/PX4/Firmware/blob/master/cmake/common/px4_base.cmake).
    
+   > **Note**
+   > If you specify `DYNAMIC` as an option to `px4_add_module`, a *shared library* is created instead of a static library.
+   > Your app will not become a builtin command, but ends up in a separate file called `examples__px4_simple_app.px4mod`.
+   > You can then run your command by loading the file at runtime using the `dyn` command: `dyn ./examples__px4_simple_app.px4mod`
 
 ## Build the Application/Firmware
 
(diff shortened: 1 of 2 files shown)


hamishwillee:
Thanks @m-ou-se . 

You say this only works for Posix builds? 
   - When would you want to declare a (dynamic) module that can only be used on Posix?
   - Can you let me know a few cases where generally the ability to make dynamic modules will improve things?
Basically, I'm trying to understand why/when someone might do this.

Once that is answered I'll merge and do some minor tidy up to fix markdown issues that affect gitbook (but not github) and cross link between the sections.

hamishwillee:
Hi @m-ou-se ,

I'd like to merge this, and I could as-is, but it would be more complete if you could possibly answer https://github.com/PX4/Devguide/pull/663#issuecomment-441487906

If you're not going to be able please let me know.

m-ou-se:
Hey, Sorry for my very late reply.

To answer your questions:

- Most importantly: It makes it easier to make modules written in other languages than C or C++. Getting those to compile together with the rest of PX4 in the same build process would be a lot harder than compiling it as a separate module. For Rust, for example: https://docs.rs/px4/
- It makes it easier to swap optional modules in and out without having to re-compile PX4.
- It also allows for sharing a module for someone to use, without having to share the source code with them (useful for companies producing custom drivers with restrictions, for example).

It could probably be made available on NuttX as well, but it'll cost a lot of memory, as all linker symbols need to be available at run-time, which is not really a problem on Posix platforms.

hamishwillee:
Thanks @m-ou-se - what's 2 months between friends :-). I merged this and added the core info in https://github.com/PX4/Devguide/pull/807 - can you check that is correct/useful? It should be - so this is just a sanity check.

PR #667, opened 2018-11-26
Companion: fix baudrate param

bkueng:
@DonLakeFlyer fyi

### DIFF
--- a/en/companion_computer/pixhawk_companion.md
+++ b/en/companion_computer/pixhawk_companion.md
@@ -11,7 +11,7 @@ Enable MAVLink on any [configurable serial port](https://docs.px4.io/en/peripher
 To set up the default companion computer message stream on `TELEM 2`, set the following parameters:
 * [MAV_1_CONFIG](../advanced/parameter_reference.md#MAV_1_CONFIG) = `TELEM 2` (`MAV_1_CONFIG` is often used to map the `TELEM 2` port)
 * [MAV_1_MODE](../advanced/parameter_reference.md#MAV_1_MODE) = `Onboard`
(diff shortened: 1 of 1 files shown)


en/companion_computer/pixhawk_companion.md
```diff
@@ -11,7 +11,7 @@ Enable MAVLink on any [configurable serial port](https://docs.px4.io/en/peripher
 To set up the default companion computer message stream on `TELEM 2`, set the following parameters:
 * [MAV_1_CONFIG](../advanced/parameter_reference.md#MAV_1_CONFIG) = `TELEM 2` (`MAV_1_CONFIG` is often used to map the `TELEM 2` port)
 * [MAV_1_MODE](../advanced/parameter_reference.md#MAV_1_MODE) = `Onboard`
-* [MAV_X_RATE](../advanced/parameter_reference.md#MAV_X_RATE) = `921600` (or higher)
+* [SER_TEL2_BAUD](../advanced/parameter_reference.md#SER_TEL2_BAUD) = `921600` (or higher)
```
DonLakeFlyer, line 14: Yeah, much better. But isn't the (or higher) wrong here? For example I use 57600 for baud rate.

bkueng: It's a recommendation since typically a high bandwidth is required for a companion.
What about `(921600 or higher recommended for applications like log streaming or FastRTPS)`?

DonLakeFlyer: Any sort of wording which doesn't seem to say that 921600 is the lower bound is fine.

bkueng: Fixed in https://github.com/PX4/Devguide/commit/b3a1e4cb4e496b64c901e13709050cc7da2f1999

PR #669, opened 2018-11-26
Improve instructions for adding new airframe group

hamishwillee:
Add information about resource file updates, and also group the QGC actions together.


DonLakeFlyer:
Looks good

PR #671, opened 2018-11-27
Move build system info from simulation to building

hamishwillee:
Logically it makes sense to have the `make` docs in building PX4 rather than in the Simulation topic. This moves them and does some tidy up.


bkueng (approved):

hamishwillee:
Thanks @bkueng - I did a further minor update to show the mapping between vendor/model/variant to the file structure and simplify the syntax presented in the simulation section. Much happier now. Merging :-)

PR #573, opened 2018-08-06
Add instructions for mocap fusion using EKF2

mzahana:
This PR updates the instructions on setting up EKF2 to fuse MOCAP data using MAVROS.

PR #677, opened 2018-11-29
Computer vision overview page

### LINKED ISSUE #530: Computer Vision Landing Page
Easy to discover landing page for all things computer vision. Expectation is that you can go to the dev guide and have everything laid out about all components that can be leveraged.
- Incorporate or link to https://github.com/PX4/avoidance

This should also be linked from user guide as a concept. 

Link or move docs for  into Developer guide.

Other resources:
- https://www.youtube.com/watch?v=VqZkAWSl_U0&feature=push-sd&attr_tag=ymcTQRVV1-ppQPbr-6
- Possible cross linking or requirements blend with https://github.com/PX4/px4_user_guide/issues/251

hamishwillee:
Fixes #530

This adds a computer vision page with links to obstacle avoidance, vio, collision prevention, optical flow. Essentially it provides a link for people using the devguide to find the right resources.


hamishwillee:
@mrivi @baumanta FYI also.

hamishwillee:
Can anyone clarify the relationship between VIO, Optical Flow and MoCap/Vision? I have put these in separately because our docs are separate, but they seem very similar.

My guess from the docs is that 
- all these techniques are VIO (or possibly VO)
- Optical Flow is one particular implementation of VO ("no I") with uorb topics and MAVLink messages for position and distance. There is also a particular hardware PX4FLOW that we know provides the messages. 
- mocap does the same thing using a different set of messages. This is coupled to ROS for the implementations we use, and is intended to run with vision system on a companion computer.

So essentially we have two supported models for VIO using different APIs, and primarily intended for a standalone board in one case, and companion in the other. Is that about right? Are there any particular benefits/limitations of one over the other?

en/advanced/computer_vision.md
```diff
@@ -0,0 +1,32 @@
+# Computer Vision (VIO, Avoidance)
+
+[Computer vision](https://en.wikipedia.org/wiki/Computer_vision) techniques enable computers to use visual data to make sense of their environment. 
+
+PX4 uses computer vision systems (primarily running on [Companion Computers](../companion_computer/pixhawk_companion.md)) in order to support the following features:
+- [Optical Flow](https://docs.px4.io/en/sensor/optical_flow.html) uses a downward facing camera and a downward facing distance sensor for position estimation.
```
bkueng: It's used for 2D velocity estimation, to be more precise.

en/advanced/computer_vision.md
```diff
@@ -0,0 +1,32 @@
...
+- [Visual Inertial Odometry](#vio) is used for navigation without a global position source (e.g. indoors).
+- [Obstacle Avoidance](https://docs.px4.io/en/computer_vision/obstacle_avoidance.html) provides navigation around obstacles when flying a planned path (currently missions are supported). This uses [PX4/avoidance](https://github.com/PX4/avoidance) running on a companion computer.
+- [Collision Prevention](https://docs.px4.io/en/computer_vision/collision_prevention.html) is used to stop vehicles before they can crash into an obstacle (primarily when flying in manual modes).
+
+
+## Visual Inertial Odometry (VIO) {#vio}
+
+VIO is used for estimating the *pose* (position and orientation) of a moving vehicle relative to a *local* starting position.
+This can be used to navigate a vehicle in the absence of GPS (or another global position source), and is used, in particular, for indoor navigation.
```
bkueng: You can add that is also helps outdoors, in situations where you lose GPS, like under a bridge.

bkueng (commented):
> all these techniques are VIO (or possibly VO)

No, only what's called VIO is actually VIO (if that makes sense :). Optical flow uses different algorithms and provides a different output (2D velocity vs. 3D pose). So V(I)O is more powerful, but also more difficult to get to work robustly (and uses more resources).

> Optical Flow is one particular implementation of VO ("no I") with uorb topics and MAVLink messages for position and distance. There is also a particular hardware PX4FLOW that we know provides the messages.

No, it's separate, see above (distance is optional, it does not need to be combined in the same sensor/board).

> mocap does the same thing using a different set of messages. This is coupled to ROS for the implementations we use, and is intended to run with vision system on a companion computer.

I think it's using the same interface (not sure though). It stands for motion capture, and it provides the same information as VIO (a pose), but through different means: typically through external fixed installed cameras that track the vehicle and send the pose in real-time to the vehicle. So it's limited to a location. It can be used instead of VIO (to localize the vehicle) or to provide ground-truth information to validate another pose estimation method (like VIO).

hamishwillee:
Thanks @bkueng . I'm clear on the difference between VIO and Optical flow and have updated this appopriately. 

But what I think you are saying is that VIO and Motion Capture are different. The main difference being mocap provides a pose based on external cameras looking at the vehicle, while VIO provides pose and velocity based on cameras on the vehicle. Is that right?

Looking at the links under VIO: 
- These look like MoCap docs that cover exactly the same material and should be merged (right?)
  - [External Position Estimation](http://dev.px4.io/en/ros/external_position_estimation.html)
  - [Flying with Motion Capture (VICON, Optitrack)](http://dev.px4.io/en/tutorials/motion-capture-vicon-optitrack.html)
- This is doc on EKF tuning that probaby applies to both VIO and MoCap [EKF > External Vision System](../tutorials/tuning_the_ecl_ekf.md#external-vision-system)  (?)

or to put it another way, what docs would I have under Mocap and which under VIO?

bkueng:
> But what I think you are saying is that VIO and Motion Capture are different. The main difference being mocap provides a pose based on external cameras looking at the vehicle, while VIO provides pose and velocity based on cameras on the vehicle. Is that right?

Right, just that an external system does not necessarily need to use cameras (it could use liadar or UWB as well). Camera (infrared) is the most common though I think.

> These look like MoCap docs that cover exactly the same material and should be merged (right?) 

From the content it does look the same and I think it can be merged. Strictly from the headings 'Flying with Motion Capture (VICON, Optitrack)' would be a subpage of 'Using Vision or Motion Capture Systems'.

> This is doc on EKF tuning that probaby applies to both VIO and MoCap EKF > External Vision System (?)

Yes I think so.

hamishwillee:
Thanks @bkueng . I have separated out MoCap and VIO in the docs as discussed. Can you please re-review.

I have not handled the merge of the two topics on MoCap/VIO. There is an ongoing effort to get the ROS doc updated, and that should come first. See https://github.com/PX4/Devguide/pull/573#issuecomment-443563730

en/advanced/computer_vision.md
```diff
@@ -0,0 +1,49 @@
...
+
+
+## Visual Inertial Odometry {#vio}
+
+Visual Inertial Odometry (VIO) is used for estimating the 3D *pose* (position and orientation) of a moving vehicle relative to a *local* starting position.
+It is commonly used to navigate a vehicle in situations where GPS is absent (e.g. indoors) or unreliable (e.g. when flying under a bridge).
+
+VIO uses [Visual Odometry](https://en.wikipedia.org/wiki/Visual_odometry) to estimate vehicle *pose* from visual information, combined with inertial measurements from an IMU (to correct for errors associated with rapid vehicle movement resulting in poor image capture).
+
+> **Note** VIO uses some similar techniques to [MoCap](#mocap). One difference is that the VIO cameras/IMU are vehicle-based, and additionally provide velocity information.
```
bkueng: Can you clarify this? I don't see how the techniques are similar.

hamishwillee: That was my interpretation of your comments - ie that they both use image analysis to get position/pose. Because I don't understand the detail I used the "weasel words" *some similar* :-). 
I've updated this so that it just refers to the on/off vehicle camera difference.

bkueng (approved):

hamishwillee:
Thanks for your help (and patience) on this one @bkueng. I hope this will be helpful.

PR #679, opened 2018-12-03
Add build troubleshooting section

hamishwillee:


mhkabir:
@dagar

mhkabir (approved):

mhkabir:
This is a fine first step, but I suspect that we'll need : 
1. A way to permanently set the file limit.
2. Better visibility for this solution (maybe on the toolchain/devenv page?), since potentially _everyone_ on macOS will run into this.

hamishwillee:
Good point. I'll add a section here: http://dev.px4.io/en/setup/dev_env_mac.html#homebrew-installation

Note, this will have to be removed when you fix this issue. Got a PR I can use to track this?

hamishwillee:
@mhkabir @dagar Were the issues that this was documenting fixed? if not, is there anything I need to do before merging (other than fixing up the conflicts).

hamishwillee:
I'm merging this, as the content is still relevant.

PR #683, opened 2018-12-04
mavlink: fixed boilerplate code

potaito:
Reference for `get_id()`:
https://github.com/PX4/Firmware/blob/5ba56c8c2cb7b48b6987ac06a503fd8841eb9925/src/modules/mavlink/mavlink_messages.cpp#L311-L314

Reference for `send()`:
https://github.com/PX4/Firmware/blob/5ba56c8c2cb7b48b6987ac06a503fd8841eb9925/src/modules/mavlink/mavlink_messages.cpp#L343-L362

@hamishwillee please take a look :)

### DIFF
--- a/en/middleware/mavlink.md
+++ b/en/middleware/mavlink.md
@@ -55,7 +55,7 @@ public:
     {
         return "CA_TRAJECTORY";
     }
-    uint8_t get_id()
+    uint16_t get_id()
     {
         return MAVLINK_MSG_ID_CA_TRAJECTORY;
     }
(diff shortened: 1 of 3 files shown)


potaito:
Actually there is another error regarding 

```
_mavlink->send_message(MAVLINK_MSG_ID_CA_TRAJECTORY, &_msg_ca_trajectory);
```

which is now different: https://github.com/PX4/Firmware/blob/5ba56c8c2cb7b48b6987ac06a503fd8841eb9925/src/modules/mavlink/mavlink_messages.cpp#L358-L359

PR #688, opened 2018-12-10
Point at github releases for cygwin msi

hamishwillee:


hamishwillee:
@MaEtUgR Would be good if the S3 link had a "latest" URL we could use. Is S3 link needed or best just to point to Github releases page?

MaEtUgR (approved):
Makes sense!

The only reason why I kept both places is dowload performance. On GitHub the hosting space for releases is free but they'll throttle your download without hesitation if the servers are already loaded. On amazon we pay for space and traffic (not a lot) and the result is consistent high download speeds. Given the file has a significant size I thought I still need to find a solution. e.g. automatically mirror to S3 and have the link in the same place if that's possible.

hamishwillee:
@MaEtUgR Yes, we keep S3 version. I will merge for now but will update later if you can provide a link to "latest" on S3

PR #689, opened 2018-12-11
Add vim-common for xxd

### LINKED ISSUE #685: add xxd via vim as a dep for ubuntu?
Put ```apt-get install vim``` in ubuntu_sim_nuttx.sh

like with  PR #301 

I was building on ~~docker~~ **a stock ubuntu:16.04 docker image that I set up with  ubuntu_sim_nuttx.sh** and got error that xxd was not installed:

```bash
make px4fmu-v5_default
-- PX4 VERSION: v1.8.0-1490-gf637f6a
# BLAH BLAH BAH ...
-- ROMFS: px4fmu_common
CMake Error at ROMFS/CMakeLists.txt:157 (message):
  xxd not found
```

as the msg implies, xxd is used to do something in ROMFS/CMakeLists.txt

```CMake
# create nsh_romfsimg.c
find_program(XXD xxd)
if(NOT XXD)
    message(FATAL_ERROR "xxd not found")
endif()
add_custom_command(OUTPUT nsh_romfsimg.c
        COMMAND ${CMAKE_COMMAND} -E remove -f nsh_romfsimg.c
        COMMAND ${XXD} -i romfs.img nsh_romfsimg.c
        COMMAND sed 's/unsigned/const unsigned/g' nsh_romfsimg.c > nsh_romfsimg.c.tmp && ${CMAKE_COMMAND} -E rename nsh_romfsimg.c.tmp nsh_romfsimg.c
        DEPENDS romfs.img
        )
```

So after a google search, I find out it is distributed with apt in ```vim-common```.

I installed vim (```apt install vim```) and now it compiles.

Another solution would be to switch to something more common like ```od```?

Anyway, maybe this issue (/ pr if someone makes one) might help someone someday

hamishwillee:
Fixes #685

### DIFF
--- a/en/setup/dev_env_linux_centos.md
+++ b/en/setup/dev_env_linux_centos.md
@@ -20,7 +20,7 @@ easy_install pexpect
 easy_install toml
 easy_install pyyaml
 easy_install cerberus
-yum install openocd libftdi-devel libftdi-python python-argparse flex bison-devel ncurses-devel ncurses-libs autoconf texinfo libtool zlib-devel cmake
+yum install openocd libftdi-devel libftdi-python python-argparse flex bison-devel ncurses-devel ncurses-libs autoconf texinfo libtool zlib-devel cmake 
 ```
 
 > **Note** You may want to also install  python-pip and screen
--- a/en/setup/dev_env_linux_centos.md
+++ b/en/setup/dev_env_linux_centos.md
@@ -20,7 +20,7 @@ easy_install pexpect
 easy_install toml
 easy_install pyyaml
 easy_install cerberus
-yum install openocd libftdi-devel libftdi-python python-argparse flex bison-devel ncurses-devel ncurses-libs autoconf texinfo libtool zlib-devel cmake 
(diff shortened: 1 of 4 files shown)


en/setup/dev_env_linux_centos.md
```diff
@@ -20,7 +20,7 @@ easy_install pexpect
 easy_install toml
 easy_install pyyaml
 easy_install cerberus
-yum install openocd libftdi-devel libftdi-python python-argparse flex bison-devel ncurses-devel ncurses-libs autoconf texinfo libtool zlib-devel cmake
+yum install openocd libftdi-devel libftdi-python python-argparse flex bison-devel ncurses-devel ncurses-libs autoconf texinfo libtool zlib-devel cmake
```
hamishwillee: @dagar How would I install vim-common on centos? Just add to this list

dagar: I believe it's `vim-common` in centos as well, but the easiest thing to do would be jump into a fresh centos container and give this a try.

hamishwillee: That's not easy. I've updated this to add vim-common because it is, according to a few sources on the internet, the correct thing to do.

hamishwillee:
Thanks @dagar . I am assuming this is otherwise good and merging now.

PR #690, opened 2018-12-11
Fixing broken link on documentation

RicardoM17:
@hamishwillee


hamishwillee:
Hi Richard, 

Unfortunately I read my emails/slack out of order and submitted/merged the same fix a few minutes ago. Thanks very much for doing the right thing.

Cheerio
H

PR #698, opened 2018-12-26
Fix a Markdown syntex error in pwm_limit.md

bys1123:
It's look like this now, I think maybe it's not the desire format.
image


hamishwillee:
Thanks @bys1123 ! Merged.

PR #699, opened 2018-12-27
Use new name for Dronecode SDK

bys1123:


hamishwillee:
Awesome thanks. Generally speaking we don't need to update the translated versions. The round-trip through crowdin should do that for us. It doesn't make much difference except that sometimes I end up having to force merge.

bys1123:
Thanks for your suggestion. I didn't notice that I changed translated versions until I created the PR, will do better next time.

PR #703, opened 2018-12-30
Adding a section about runnig simulation on a remote server

kaklik:
The new section discussing the options of connection in case of simulation is run on multiple computers. The SSH tunneling example and a link to connection bash script were introduced.

### DIFF
--- a/en/simulation/README.md
+++ b/en/simulation/README.md
@@ -140,3 +140,32 @@ The simulated camera is a gazebo plugin that implements the [MAVLink Camera Prot
 
 The same approach can be used by other simulators to implement camera support.
 
+## Running simulation on the remote server 
+
+As had been mentioned at the beginning the simulation environment can be run on multiple computers on the same network. Unfortunately, it is a slightly complicated situation, because of the out of box configuration of Ubuntu computer does not broadcast the PX4 UDP packets to external interfaces and packets are routed internally by default. 
+A solution is to modify system routes to broadcast UDP packets to the network or use a tunnel to connect computers together. 
+
+Using the tunnel is a more flexible option because the computers are not required to sit on the same network and remote powerful simulation server can be used for example. 
+
+One, probably the easiest way to create the tunnel is the use of SSH tunneling options.  The tunnel itself could be created easily by running the following command on localhost.
+
+    ssh -C -fR 14551:localhost:14551 remote.local
+
+Where "remote.local" is the name of a remote computer. 
+
+Unfortunately, the SSH itself cannot route the UDP packets.  Therefore the UDP packets need to be translated to TCP packets. By netcat utility separately on the local and remote side of the tunnel. 
+Local side of UDP packet translation of QGC could be implemented by running following commands.
+
+     mkfifo /tmp/tcp2udp
+     netcat -lvp 14551 < /tmp/tcp2udp | netcat -u localhost 14550 > /tmp/tcp2udp
+
+For the remote side of the tunnel, the command differs. 
+
+    mkfifo /tmp/udp2tcp
+    netcat -lvup 14550 < /tmp/udp2tcp | netcat localhost 14551 > /tmp/udp2tcp
+
+It is necessary to have QGC running before executing the netcat. 
+The tunnel could run infinitely, but netcat connections may need a restart in case of improper communication state occurs. The port number 14550 is valid for QGC software connection and should be adjusted for other possible communication channels.  
+
+The automated [bash connection script](https://raw.githubusercontent.com/ThunderFly-aerospace/sitl_gazebo/autogyro-sitl/scripts/QGC_remote_connect.bash) is prepared for automation of QGC to simulation server running the PX4 stack. 
+
--- a/en/simulation/README.md
+++ b/en/simulation/README.md
@@ -140,7 +140,7 @@ The simulated camera is a gazebo plugin that implements the [MAVLink Camera Prot
 
 The same approach can be used by other simulators to implement camera support.
 
-## Running simulation on the remote server 
+## Running Simulation On The Remote Server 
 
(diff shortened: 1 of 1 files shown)


en/simulation/README.md
```diff
@@ -140,32 +140,32 @@ The simulated camera is a gazebo plugin that implements the [MAVLink Camera Prot
 
 The same approach can be used by other simulators to implement camera support.
 
-## Running simulation on the remote server 
+## Running Simulation On The Remote Server
```
hamishwillee: Just FYI, in this kind of capitalisation it is usual to not capitalise small connecting words - e.g. "on the"

hamishwillee:
Thanks very much @kaklik.

This is a pretty cool approach if you need to bridge networks, but fairly complicated.

It seems like this should not be a Ubuntu limitation, and that if it is, then other approaches like using https://github.com/intel/mavlink-router might be much easier for people to set up to connect on the same network.  @bkueng Have you tried this/got knowledge of it? Would this be disproved if QGC on one computer could connect to Simulator running on another using the other computer's IP address port 14540?

kaklik:
@hamishwillee  Yes I know about mavlink-router, but it is a different approach because this utility needs to have server ports opened. Therefore it is complicated to use it on controlled networks like academy networks. 
Moreover, the mavlink-router is undocumented in the Devguide. 

The statement that routing is Ubuntu limitation is my mistake I recently discovered that PX4 has the parameter MAV_BROADCAST, which controls broadcasting to the local network. Therefore in case, computers are directly connected and network allows connection on all ports. My script could be avoided by enabling MAV_BROADCAST. 

But in the case the network is limited and only several ports are opened, like SSH, my script is useful. 

I will correct it in the Devguide text asap.

hamishwillee:
@kaklik Thank you very much. I don't question the value of the script at all - it is cool. Just the "positioning". 

When you say MAVLink router is undocumented in DevGuide, do you mean "that it exists" or "how it is used". If the former, then I added this following your post of this issue: http://dev.px4.io/en/companion_computer/pixhawk_companion.html#companion-computer-setup

Are there other places you think it should be linked?

kaklik:
@hamishwillee Ok, I have fixed the paragraph and mentioned the MAV_BROADCAST parameter. 

I think there are missing a section in DevGuide summarising an advanced setup of a  ground control station. Such section should contain paragraphs like wind measurement, DGPS,  landing beacons, tracking antenna systems, connection to modems (MAVLink router), in-flight debugging tools etc.  The most similar section which already exists is this: http://dev.px4.io/en/qgc/ But it contains QGC setup only.

hamishwillee:
@kaklik Thanks for that. I'm going to wait on @bkueng before reviewing this in detail, because has a lot more experience of this stuff. ON first scan though I will probably pull it apart slightly so that the tunnel is presented as an option under its own heading.

You're certainly right that we don't have an advanced setup of QGC document. The docs tend to be organised by functionality, so if you're interested in precision landing or more accurate GPS you'd look at https://docs.px4.io/en/advanced_features/ (user guide). That is somewhat a product of the fact that features get added over time, but also because what is an interesting advanced setup for one person is quite different for another.

That said, it should be easy/possible to find the information you need, and if it isn't then this does need to be revisited. What we probably need is a better feature list with links to associated information.

en/simulation/README.md
```diff
@@ -140,3 +140,32 @@ The simulated camera is a gazebo plugin that implements the [MAVLink Camera Prot
 
 The same approach can be used by other simulators to implement camera support.
 
+## Running Simulation on the Remote Server 
+
+As had been mentioned at the beginning the simulation environment can be run on multiple computers on the same network. Unfortunately, it is a slightly complicated, because of the out of box configuration does not broadcast the PX4 UDP packets to external interfaces and packets are routed internally by default.
```
bkueng, line 145: Note: the internal default routing is on purpose to avoid spamming the network and to avoid different simulations interfering with each other.

bkueng (commented):
Looks good.
Some alternatives:
- If you're not in the same network, a VPN is also an option.
- If you're on the same network, the stream can also be sent to a certain IP directly (via UDP), instead of broadcasting.

en/simulation/README.md
hamishwillee: So just to spell it out for me, the internal default routing just means that we route to localhost?

> If you're on the same network, the stream can also be sent to a certain IP directly (via UDP), instead of broadcasting.

Is this the -t option in http://dev.px4.io/en/middleware/modules_communication.html#mavlink_usage ?
WOuld I be correct in assuming that this means that to route to an external port you would need to change the calls in code that launch mavlink and talk to 14550 (say) to use the -t and an IP address?

> If you're not in the same network, a VPN is also an option.

Is there any setup needed to discuss for this? Or is it as simple as setting your system up to use VPN for external traffic and updating to use external network address as per previous point above?

Sorry if these are dumb questions - I just don't want to write "use VPN" if that then means someone has to do a whole lot of working out of stuff.

bkueng: > So just to spell it out for me, the internal default routing just means that we route to localhost?

Right.

> Is this the -t option in http://dev.px4.io/en/middleware/modules_communication.html#mavlink_usage ?
WOuld I be correct in assuming that this means that to route to an external port you would need to change the calls in code that launch mavlink and talk to 14550 (say) to use the -t and an IP address?

Correct.

> Is there any setup needed to discuss for this? Or is it as simple as setting your system up to use VPN for external traffic and updating to use external network address as per previous point above?

Nothing specifically. If you have the VPN, you're essentially back to the local network case (assuming the VPN routes all ports, which is usually the default).

> Sorry if these are dumb questions - I just don't want to write "use VPN" if that then means someone has to do a whole lot of working out of stuff.

Not dumb questions! VPN is a large topic in itself and there are many variations, and therefore lots of things that can go wrong. But I would not go into that here, I think someone setting up a VPN for that needs to have some networking background anyway. What's important is to have good documentation which ports we use.

kaklik:
Is there something which I should do for merging this pull-request?

hamishwillee:
@kaklik 

> Is there something which I should do for merging this pull-request?

No thanks. I'll merge it shortly.

hamishwillee:
Thanks very much. I've merged and done the subedit in #716 - for your review.

PR #706, opened 2019-01-01
Gitbook versioning

dagar:
@hamishwillee any opinion on granularity? Either match releases 1-1 or get away with v1.8 docs until a point release version is necessary.

I think we might as well go with branches instead of tags (helps with CI as well).

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -44,16 +44,19 @@ pipeline {
         unstash('gitbook')
         withCredentials([usernamePassword(credentialsId: 'px4buildbot_github_personal_token', passwordVariable: 'GIT_PASS', usernameVariable: 'GIT_USER')]) {
           sh('git clone https://${GIT_USER}:${GIT_PASS}@github.com/PX4/dev.px4.io.git')
-          sh('rm -rf dev.px4.io/*')
-          sh('cp -r _book/* dev.px4.io/')
-          sh('cd dev.px4.io; git add .; git commit -a -m "gitbook build update `date`"')
+          sh('rm -rf dev.px4.io/${BRANCH_NAME}')
+          sh('mkdir -p dev.px4.io/${BRANCH_NAME}')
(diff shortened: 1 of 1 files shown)


dagar:
Seems to be working. Note the pr-jenkins directory in https://github.com/PX4/dev.px4.io (we can remove later). 

Once this is merged we'll have a `master` subdirectory in dev.px4.io, then we can cherry-pick this change to the `v1.8` branch (or `v1.8.3` if desired).

Then in the dev.px4.io repo we delete most of the top level contents and add a simple redirect to the current stable version.

The process for a new stable PX4 release will then be - 
 1. branch Devguide
 2. update gitbook versions
 3. change dev.px4.io index.html

hamishwillee:
Branches are better, because we *know* that sometimes we're going to end up with a point revision that has stuff isn't in that release - at least with a branch we can choose to edit the content. 

> any opinion on granularity? Either match releases 1-1 or get away with v1.8 docs until a point release version is necessary.

I _think_ we match releases 1:1. We probably rename the https://github.com/PX4/Devguide/tree/v1.8 branch to 1.8.0. 
We then have the choice of deleting old point revisions when we get to a new major release. E.g. at 1.9.0 we migh rename 1.8.3 (say) to v 1.8. 

My reasoning is that if we were to release 1.8.3 now it would differ significantly from either the v1.8 branch or the 1.8.2 branch because of this posix change. While we can cherry pick, I'd prefer to minimise that and try to just use latest snapshot at release time (ie effectively same as using tags, but we have option to do updates if we like). 

Make sense? 

If you really want to just have a v1.8 I can live with that too. The problem for this case would be quite a lot of work cherry picking now to get things to match.

hamishwillee:
Nice. Link to moved master updated here: https://github.com/PX4/Devguide/pull/707

I'll do the other ones when the v1.8.x folders appear.

PR #708, opened 2019-01-04
Update config argument in in posix px4 executable calls.

karliss:
Some places were already using correct arguments, but some weren't updated.
See PX4/Firmware#11114

### DIFF
--- a/en/debug/gdb_debugging.md
+++ b/en/debug/gdb_debugging.md
@@ -45,7 +45,7 @@ Start jmavsim: `./Tools/jmavsim_run.sh`
 
 In another terminal, type:
 ```bash
-cd build/px4_sitl_default/tmp
+cd build/px4_sitl_default/tmp/rootfs
 export HEAPPROFILE=/tmp/heapprofile.hprof
 export HEAP_PROFILE_TIME_INTERVAL=30
 ```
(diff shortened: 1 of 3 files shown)


hamishwillee:
Thanks @karliss . @bkueng Can you please review - I don't know what this "should" look like. 

Note that the build failure looks like Jenkins OOM and not a "real" error.

en/debug/gdb_debugging.md
```diff
@@ -54,13 +54,13 @@ Enter this depending on your system:
 
 ##### Fedora:
 ```bash
-env LD_PRELOAD=/lib64/libtcmalloc.so ../src/firmware/posix/px4 ../../posix-configs/SITL/init/lpe/iris
+env LD_PRELOAD=/lib64/libtcmalloc.so ../src/firmware/posix/px4 -s ../../posix-configs/SITL/init/lpe/iris
```
bkueng: This should be:
```suggestion
env LD_PRELOAD=/lib64/libtcmalloc.so PX4_SIM_MODEL=iris ../../bin/px4 ../../../../ROMFS/px4fmu_common -s etc/init.d-posix/rcS
```

And further up: 
```
cd build/px4_sitl_default/tmp/rootfs
```

bkueng (commented):
Looks good except for SITL.

en/debug/gdb_debugging.md
karliss: I was a bit suspicious that commands were outdated in other ways beside what I was fixing. Implemented your fix in 3b7bb69 . Are you sure it isn't init.d-posix/rcS instead of etc/init.d-posix/rcS ? At least that's where it was for me.

bkueng: Yes, `etc` is required. In `rootfs` you'll have a symlink `etc` to ROMFS/px4fmu_common.

karliss: Ok, from the description didn't expect the rootfs_directory argument to work the way it works. I thought that rootfs_directory needs to point to directory containing etc directory and behaves more like cd to rootfs_directory or prepending it to all relative paths.

hamishwillee: @bkueng  Personally I don't understand any of the logic for any of what is described above - these docs/instructions are all just turnkey instructions with no context. That kind of sux, because (in a case like this) if something goes wrong there is nothing to be done except raise an error. 

But since you do, can you please check whether the description referred to in https://github.com/PX4/Devguide/pull/708#discussion_r246332358 is ambigous and update as appropriate?

Merge whenever you are happy.

bkueng: @karliss It's not quite clear to me how you would like to see it being updated, but please do the required changes as you see fit (https://github.com/PX4/Firmware/blob/master/platforms/posix/src/main.cpp#L557).

I updated the path and merged it.

PR #709, opened 2019-01-04
Jenkins cleanup when finished

dagar:


Jenkinsfile
```diff
@@ -49,14 +49,19 @@ pipeline {
           sh('cp -r _book/* dev.px4.io/${BRANCH_NAME}/')
           sh('cd dev.px4.io; git add ${BRANCH_NAME}; git commit -a -m "gitbook build update `date`"')
           sh('cd dev.px4.io; git push origin master')
+          
+        }
+        post { 
+          always { 
+            sh('rm -rf dev.px4.io')
```
hamishwillee, line 56: Doesn't this delete everything - ie all branches?

dagar: This is the local workspace after everything has been pushed. The error you shared earlier is the result of the build running on a build slave that's already built this before.

PR #710, opened 2019-01-08
Simulation: Update PX4 QGC listener port to 14570

### LINKED ISSUE #701: Simulation port numbers are incorrect
I want to tunnel QGC control ports to the simulation server, but I get stuck on port numbers selection. 

It is because of the [documentation](https://dev.px4.io/en/simulation/#default-px4-mavlink-udp-ports) claims the QGC listens on port 14550. It is correct because tunneling this port allows connection from PX4 to QGC.  But also the documentation mention the PX4 listens on port 14556, which is probably not correct, because of the tunneling this port to the PX4 does not allow QGC talk to PX4.
I have tried to scan listening ports on PX4 and there are no 14556 port open. Instead of this, there are few other documentation unmentioned ports opened (14570 and 14580). 

    kaklik@TF-R1:~$ ss -upnl4
    State Recv-Q Send-Q    Local Address:Port    Peer Address:Port                                       
    UNCONN0      0               0.0.0.0:14570        0.0.0.0:*     users:(("px4",pid=15118,fd=9))       
    UNCONN0      0               0.0.0.0:14580        0.0.0.0:*     users:(("px4",pid=15118,fd=10))      
    UNCONN0      0               0.0.0.0:11317        0.0.0.0:*     users:(("gzclient",pid=15116,fd=21)) 
    UNCONN0      0               0.0.0.0:11317        0.0.0.0:*     users:(("gzserver",pid=15089,fd=28)) 
    UNCONN0      0               0.0.0.0:11318        0.0.0.0:*     users:(("gzclient",pid=15116,fd=22)) 
    UNCONN0      0               0.0.0.0:11318        0.0.0.0:*     users:(("gzserver",pid=15089,fd=29)) 
    UNCONN0      0         127.0.0.53%lo:53           0.0.0.0:*                                          
    UNCONN0      0               0.0.0.0:68           0.0.0.0:*                                          
    UNCONN0      0               0.0.0.0:631          0.0.0.0:*                                          
    UNCONN0      0               0.0.0.0:33445        0.0.0.0:*                                          
    UNCONN0      0               0.0.0.0:5353         0.0.0.0:*                                          
    kaklik@TF-R1:~$ 

Above list corresponds to "gazebo_rover" simulation.

hamishwillee:
Fixes #701 

The SITL PX4 listener port for QGC/GCSs changed form 14566 to 14570. This updates the diagram appropriately.

Note, that this does not appear to be in any text, and before I added this to the diagram was undocumented. I believe it is considered "somewhat internal" but unfortunately it is known, and used by things like ROS.

@bkueng 
- What PX4 version did this go into? Ie was it in 1.8.2? or 1.8.1?
- Is there anything else we need to comment on? @kaklik refers to another open port on 14580?


bkueng:
> What PX4 version did this go into? Ie was it in 1.8.2? or 1.8.1?

Hmm I don't remember exactly. It was not that long ago.

> Is there anything else we need to comment on? @kaklik refers to another open port on 14580?

Yes the local offboard port changed as well from 14557 to 14580.

Another change: @julianoes recently changed the simulator to TCP on port 4560 (but we can leave UDP in, as it's still a viable option).

julianoes:
Hm, I don't actually think these new port numbers make sense. I would like to see some consistency like:
14550 -> 14551
14540- -> 14541
...

hamishwillee:
> Hm, I don't actually think these new port numbers make sense. I would like to see some consistency like:
> 14550 -> 14551
> 14540- -> 14541

That would be nice, particularly as by putting offboard to 14580 we've limited to 10 vehicle anyway. The only obvious reason to offset to somewhere else would be to allow many more vehicles, and I suspect the limit would then be simulation resource. @bkueng is this something you will do - ie should I wait for it?

> @julianoes recently changed the simulator to TCP on port 4560 (but we can leave UDP in, as it's still a viable option).

Would that look like below? ie with both UDP and TCP? Should I specify the protocol for UDP too? ie change text from *port 14560* to *UDP:14560*?

image

julianoes:
It's now either TCP 4560 or UDP 14560. I would write it as TCP in the diagram because that's currently the default.

bkueng:
>  @bkueng is this something you will do - ie should I wait for it?

No. If someone wants to do it, I'll review, but I'm not going to change it mysefl.

hamishwillee:
Thanks very much. I've updated and will merge now. If you could do a quick check after the fact that the updated image (below) is correct would be good:

julianoes:
I think we should be using random ports for the local port and not specify them at all. That way we stay with the spec which just says e.g. 14550 for QGC and make sure we always initiate the connection from the vehicle to the ground station (or whatever) and avoid conflicts as well.

hamishwillee:
> I think we should be using random ports for the local port and not specify them at all. That way we stay with the spec which just says e.g. 14550 for QGC and make sure we always initiate the connection from the vehicle to the ground station (or whatever) and avoid conflicts as well.

@julianoes @bkueng What about the fact that ROS explicitly uses them? http://dev.px4.io/en/simulation/ros_interface.html#launching-rossimulation

hamishwillee:
@julianoes @bkueng Ping ^^^^

What about the fact that ROS explicitly uses them? http://dev.px4.io/en/simulation/ros_interface.html#launching-rossimulation

bkueng:
> What about the fact that ROS explicitly uses them? http://dev.px4.io/en/simulation/ros_interface.html#launching-rossimulation

Well, I wasn't suggesting to change anything, exactly for such reasons. But I'm also not opposed if someone really wants to go through it.

hamishwillee:
>> What about the fact that ROS explicitly uses them? http://dev.px4.io/en/simulation/ros_interface.html#launching-rossimulation

> Well, I wasn't suggesting to change anything, exactly for such reasons. But I'm also not opposed if someone really wants to go through it.

Thanks @bkueng . So @julianoes I do think it is a good idea but you would have to liase with Nuno for ROS - and I wouldn't want to drive this. So end of conversation for now unless you take it on :-)

julianoes:
> What about the fact that ROS explicitly uses them? http://dev.px4.io/en/simulation/ros_interface.html#launching-rossimulation

So it looks like ROS explicitly uses it but we already changed it and broke it anyway with https://github.com/PX4/Firmware/pull/10503.

If we by default use random ports, ROS can still set them explicitly but wouldn't have to.

hamishwillee:
> If we by default use random ports, ROS can still set them explicitly but wouldn't have to.

I assume you are suggesting that ROS would have to change to listen on the standard port and from the UDP response could get the (random) response port. Because it does not make sense that they can use their current architecture and set the random port explicitly - since they won't know what it is :-)

So yes, that is what I mean - it is a good idea, but you would need to push with ROS because I'm not planning on doing so :-)

julianoes:
Exactly. Let's see when it comes up again :smile: .

PR #712, opened 2019-01-08
simulation: add note how to run SITL faster

julianoes:
These adds docs on how to run the SITL simulation at a faster than realtime rate.

### DIFF
--- a/en/simulation/README.md
+++ b/en/simulation/README.md
@@ -102,6 +102,26 @@ The simulation can be further configured via environment variables:
 The syntax described here is simplified, and there are many other options that you can configure via *make* - for example, to set that you wish to connect to an IDE or debugger. For more information see: 
 [Building the Code > PX4 Make Build Targets](../setup/building_px4.md#make_targets).
 
+### Run Simulation Faster than Realtime
+
+SITL can be run faster or slower than realtime. The speed factor is set using the environment variable `PX4_SIM_SPEED_FACTOR`.
+The speed factor is implemented for the jMAVSim and Gazebo SITL simulation.
+
+For instance to run the jMAVSim simulation at 2 times the real time speed, do:
+```
+PX4_SIM_SPEED_FACTOR=2 make px4_sitl jmavsim
+```
+
+Or to apply to all SITL runs in the current session:
+```
+export PX4_SIM_SPEED_FACTOR=2
+
+make px4_sitl jmavsim
+```
+
+> **Note** At some point IO or CPU will limit the speed that is possible on your machine and it will be slowed down "automatically".
+> Powerful desktop machines can usually run the simulation at around 6-10x, for notebooks the achieved rates can be around 3-4x.
+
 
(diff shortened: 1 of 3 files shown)


en/simulation/README.md
```diff
@@ -102,6 +102,26 @@ The simulation can be further configured via environment variables:
 The syntax described here is simplified, and there are many other options that you can configure via *make* - for example, to set that you wish to connect to an IDE or debugger. For more information see: 
 [Building the Code > PX4 Make Build Targets](../setup/building_px4.md#make_targets).
 
+### Run Simulation Faster than Realtime
+
+SITL can be run faster or slower than realtime. The speed factor is set using the environment variable `PX4_SIM_SPEED_FACTOR`.
+The speed factor is implemented for the jMAVSim and Gazebo SITL simulation.
+
+For instance to run the jMAVSim simulation at 2 times the real time speed, do:
```
hamishwillee: For instance, to run the jMAVSim simulation at 2 times the real time speed:

hamishwillee:
Looks great. Minor rendering issues I might play with.

Some points/questions
- "The speed factor is implemented for the jMAVSim and Gazebo SITL simulation." 
  - So this is something that requires both a PX4 sitl change and a simulator change?
  - Perhaps we should link to this section from http://dev.px4.io/en/simulation/jmavsim.html#usageconfiguration-options and http://dev.px4.io/en/simulation/gazebo.html#usageconfiguration-options 
- When would you want to do this? I presume it allows you to speed up automated tests? Anything else?
- Are there any limitations to do with using this - ie say you have a developer API implementing a protocol that has timeouts. If you run simulator at 10x speed might the timeouts expire too early. What if someone is using a joystick through QGC, will that result in jittery behaviour. These cases might not be realistic - I'm basically trying to get you to think about possible fail cases.

julianoes:
- The speed factor is obeyed by jMAVSim and Gazebo, so they are the ones reading the environment variable. You're right it should be in the configuration options, at least as well.
- I would always use it. When quickly testing something, it's pretty boring to watch the drone in real time.
- That's possible fail cases, yes.

hamishwillee:
@julianoes Thanks. I merged then made some minor editorial changes in #715, mostly to link back to the more detailed changes in this PR.

Further things that would be good to see:
- Add an example of slowing the simulation - ie is that done using a fractional value, or a negative value?
- Add some info about when it should be used and the risk/problem symptoms. ie I know you said use it all the time and that is the advice, then we should give it. Otherwise why not just set the factor to be 10 and be done with it :-)

julianoes:
- Slowing down to half would be 0.5.
- We'll find out :smile: .

hamishwillee:
Indeed we will. I've added info on half time in https://github.com/PX4/Devguide/commit/04b6d4f2f61142b7c67ef51dfc46bde9f29e4763

julianoes:
Great, thanks!

PR #717, opened 2019-01-13
Add coding standard info

hamishwillee:
@mcsauder proposed a set of coding standards in https://github.com/PX4/Firmware/issues/9931

This attempts to capture a much smaller useful subset that I believe reflects the status quo. Is this accurate, anything to add/remove?

### DIFF
--- a/en/contribute/README.md
+++ b/en/contribute/README.md
@@ -103,7 +103,81 @@ To contribute new functionality, [sign up for Github](https://help.github.com/ar
 
 All code contributions have to be under the permissive [BSD 3-clause license](https://opensource.org/licenses/BSD-3-Clause) and all code must not impose any further constraints on the use.
 
-### Code Style Formatting
+### Coding Standards {#code_standards}
+
+PX4 recommends the following coding standards:
+- In-source documentation [as below](#source_docs)
+- Formatting/layout using *astyle*, [as below](#code_formatting)
+
+Other conventions:
+- All files should have standard copyright header, with date range reflecting creation to last modification date
+  (note that the comment markup may change for different types of files).
+  ```
+  /****************************************************************************
+  *
+  *   Copyright (c) 2013-2019 PX4 Development Team. All rights reserved.
+  *
+  * Redistribution and use in source and binary forms, with or without
+  * modification, are permitted provided that the following conditions
+  * are met:
+  *
+  * 1. Redistributions of source code must retain the above copyright
+  *    notice, this list of conditions and the following disclaimer.
+  * 2. Redistributions in binary form must reproduce the above copyright
+  *    notice, this list of conditions and the following disclaimer in
+  *    the documentation and/or other materials provided with the
+  *    distribution.
+  * 3. Neither the name PX4 nor the names of its contributors may be
+  *    used to endorse or promote products derived from this software
+  *    without specific prior written permission.
+  *
+  * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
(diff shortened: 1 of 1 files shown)


en/contribute/README.md
```diff
@@ -103,7 +103,79 @@ To contribute new functionality, [sign up for Github](https://help.github.com/ar
...
+    - One class per file, with filename prefix matching classname.
+  - Methods/functions and variables
+    - Use snake case (words start with lower case letters and are separated by underscores).
+    - Files containing just functions may be named following pattern of main function.
+  - Member variables
+    - Start with an underscore and otherwise use snake case.
+- Prefer `#pragma once` over include guards
+
+
+<!--
```
hamishwillee, line 160: Note, I did not include the headers section below (from Mark's original post) but have left here for discussion. Do we follow these rules? If this is something astyle can and should enforce, can we push it into the tool?

dagar:
I'm actually a bit hesitant of the idea of having a coding standard, at least not without a more coordinated effort. It can help drive consistency and get new people up to speed, but they can also be misused as dogma. "Do this because the coding standards says to" instead of "do this because of x, y, z". It's also yet another thing we'll need to maintain.

One way to approach this that might be a bit more scalable is to work from an existing standard like the google c++ styleguide (https://google.github.io/styleguide/cppguide.html). However, instead of importing the entire document we could do it piece by piece as needed, making a few relevant changes. For example here's the section on include ordering. https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

I also like the Zubax C++ coding conventions, although they aren't that exhaustive. https://kb.zubax.com/pages/viewpage.action?pageId=2195699

Lastly, if we're going to start getting picky about naming I'd prefer to get clang-tidy involved to do the dirty work. 
See readability-identifier-naming
    - http://clang.llvm.org/extra/clang-tidy/checks/list.html
    - https://sarcasm.github.io/notes/dev/clang-tidy.html

mcsauder:
Thanks @hamishwillee and @dagar,

For my 2 cents, I like this PR, and I also do like the Zubax guide, (both share a lot of similarities to projects I have enjoyed working on in the past).

I can appreciate the sentiment about a hard standard, it's not fun having a guideline turned against you a beating stick... On the same note, however, I know I frustrated Beat Kung last year trying to guess (unsuccessfully) many days in a row what his expectations of documentation were.  It wasn't my intent to be a frustration at that point, but trying to guess without a basis of agreement to fall back against is difficult for all involved. I am in favor of this PR as a useful reference for newcomers to have something to refer to.

Nice work @hamishwillee.

bkueng:
I agree with @dagar's view, especially on using something like clang-tidy.
I want to avoid:
a) having a large document with a lot of (unrelated) details that newcomers have to go through
b) having to maintain such a document
Having a tool would pretty much solve this while also making the expectations clear.

hamishwillee:
I'm not obsessed by standards either. That is why I started with a minimum set of what we actually do rather than something like Zubax docs. If we were to go with something like this, we might start with a note that this is just general guidance/a few things commonly done by the dev team.

Automation is a great idea, particularly if the toolchain self-documents the standard it is enforcing. I guess it is a concern that tool-based enforcement is a more rigid stick that just pointing to standards. 

@dagar You didn't actually say you supported the idea of tool, just that they are easier if we're getting nitpicky. Do you want to/support trialing this?

hamishwillee:
@dagar @bkueng I'm happy to "support" a style guide but I don't have time for this if there is no pull from the dev team. I'm therefore closing this. 

Come back to me when/if this is something you want to progress.

PR #719, opened 2019-01-16
build script: add astyle, remove vim-common

MaEtUgR:
I think adding `astyle` to the deafult installation makes sense because then the `make format` command can be used. I tested any it works fine on Ubuntu 18.04.

On the other hand I don't see how `vim-common` is a necessary package. People using the editor will have/install it anyways and beginners will probably not decide because it's installed here.

@julianoes Any objection regarding vim?

FYI @Stifael

### DIFF
--- a/build_scripts/ubuntu_sim_common_deps.sh
+++ b/build_scripts/ubuntu_sim_common_deps.sh
@@ -20,7 +20,7 @@ sudo apt-get remove modemmanager -y
 # Common dependencies
 echo "Installing common dependencies"
 sudo apt-get update -y
-sudo apt-get install git zip qtcreator cmake build-essential genromfs ninja-build exiftool vim-common -y
+sudo apt-get install git zip qtcreator cmake build-essential genromfs ninja-build exiftool astyle -y
 # Required python packages
 sudo apt-get install python-argparse python-empy python-toml python-numpy python-dev python-pip -y
 sudo -H pip install --upgrade pip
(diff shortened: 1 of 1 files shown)


hamishwillee:
Adding astyle makes sense, but we we need vim-common as it is the easiest/only way to install xxd. See discussion in https://github.com/PX4/Devguide/issues/685

Can you also update the build scripts, at least for ubuntu?

MaEtUgR:
Seems that `xxd` only has its own aptitude package since Ubuntu 18.04. Before the binary was crammed into the `vim-common` package without documentation. This was likely never a big problem because `vim.tiny` (depending on `vim-common`) is part of all default distributions including the minimal setup.

On Ubuntu 18.04 or later `sudo apt install xxd -y` works fine. For older Ubuntu versions when we do `sudo apt-get install vim-common -y` aptitude will automatically install a huge pile of unnecessary "recommended" packages.

Based on these findings I suggest the line which does exactly what we want with highest compatibility:
`sudo apt install xxd -y || sudo apt-get install vim-common --no-install-recommends -y`

I'll verify this on different versions and then adapt the pr.

MaEtUgR:
The suggested command does exactly what expected:
Ubuntu 12.04:
1204 command full
Ubuntu 16.04:
1604 command full
Ubuntu 18.04:
1804 command full

Here's what I meant with pile of additional packages when just running `sudo apt install vim-common` e.g. on 18.04:
1804 package pile

MaEtUgR:
I added a commit to make sure `xxd` is present on the system and installs it if not. Read the commit message for more details.

julianoes (approved):
Nice!

MaEtUgR:
Good to go?

hamishwillee:
Awesome thanks. I updated the "manual instructions" in line with this in https://github.com/PX4/Devguide/commit/6147234f0ef70bd51e1b5e9eef72ea6c21ba1cb2

PR #721, opened 2019-01-22
Fix typo error on kr/ros/mavros_offboard.md

rapsealk:


hamishwillee:
Hi @rapsealk 

Thanks for this. Unfortunately while we are using github to store our translations we're not using them to update the content. Instead we use a much better system called crowdin - this allows us to track changes in the original much more easily. 

As soon as we have sufficient content we'll replace the existing documents. We're not keeping them updated since they will be deleted anyway. Taking these changes now might result in unnecessary merge conflicts later.

It would be great if you could join our translation team and work on the docs: https://crowdin.com/project/px4-developer-guide/invite?d=7555p455n6j5l4r4c58343k4a34333

Regards
Hamish

PR #728, opened 2019-01-31
simulation: add note how to avoid datalink timeout

julianoes:
When you speed up the simulation you can receive data link timeouts because px4 is "measuring time too fast". Therefore, you need to raise the timeout by setting a param.

### DIFF
--- a/en/simulation/README.md
+++ b/en/simulation/README.md
@@ -125,6 +125,8 @@ make px4_sitl jmavsim
 > **Note** At some point IO or CPU will limit the speed that is possible on your machine and it will be slowed down "automatically".
 > Powerful desktop machines can usually run the simulation at around 6-10x, for notebooks the achieved rates can be around 3-4x.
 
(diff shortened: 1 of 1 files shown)


en/simulation/README.md
```diff
@@ -125,6 +125,9 @@ make px4_sitl jmavsim
 > **Note** At some point IO or CPU will limit the speed that is possible on your machine and it will be slowed down "automatically".
 > Powerful desktop machines can usually run the simulation at around 6-10x, for notebooks the achieved rates can be around 3-4x.
 
+<span></span>
```
hamishwillee, line 128: FYI, glitch in gitbook means you need to prefix notes that occur after each other with span as shown.

hamishwillee:
@julianoes Looks good. I subedited - can you check you're happy with the changed text?

en/simulation/README.md
julianoes: Ok thx.

PR #729, opened 2019-01-31
Simulation: add page on how to test SITL failsafes

MaEtUgR:
Based on this comment: https://github.com/PX4/Firmware/pull/11357#issuecomment-459355990 the topic is not documented enough and that's why I created a page for it.

Note that the under link loss documented behavior is the one prepared in https://github.com/PX4/Firmware/pull/11357

@hamishwillee Feel free to change it based on your documentation experience.

PR #736, opened 2019-02-04
SITL failsafes: fix RC and DL loss

MaEtUgR:
Sorry @hamishwillee I must correct my mistakes from https://github.com/PX4/Devguide/pull/729:

- I wrote as if https://github.com/PX4/Firmware/pull/11357 was accepted, which it is not (yet)
- The parameters do not get reset on every SITL restart, I don't know why I thought that


hamishwillee:
Looks great. Merged. @MaEtUgR So does this mean you'll need to do another update when https://github.com/PX4/Firmware/pull/11357 is accepted?

MaEtUgR:
Thanks @hamishwillee !

I'll do another update when anything changes, the change as it is will not be accepted like we discussed in the Devcall (I summarized the conclusion there).

PR #739, opened 2019-02-06
hello_sky: fix link to px4_add_module docs

hamishwillee:
This fixes (I "think") broken link to px4_add_module() docs in hello_sky tutorial.

@bkueng

1. Can you also confirm that the code below basically identifies dependency library for this app (platforms__common) and that this library is the common platform components - as opposed to those for nuttx only or posix?
   ```
    DEPENDS
        platforms__common
  ```
2. That if removed the code will not work. What is the code that this provides access to? The bit that registers the app in the line `MAIN px4_simple_app`?

### DIFF
--- a/en/apps/hello_sky.md
+++ b/en/apps/hello_sky.md
@@ -133,7 +133,7 @@ In this section we create a *minimal application* that just prints out `Hello Sk
    The `px4_add_module()` method builds a static library from a module description. 
    The `MAIN` block lists the name of the module - this registers the command with NuttX so that it can be called from the PX4 shell or SITL console.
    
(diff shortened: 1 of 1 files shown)


julianoes:
It seems like this has changed is not necessary anymore looking at the updated source here:
https://github.com/PX4/Firmware/blob/master/src/examples/px4_simple_app/CMakeLists.txt

`px4_simple_app` is compiled as part of the SITL build, so I would assume that's all working.

hamishwillee:
Thanks very much @julianoes - you're right. @bkueng I updated the cmakelists file as well and will self merge. This is OK now

PR #741, opened 2019-02-08
modules doc: update to latest master and add distance sensor to TOC

bkueng:
Manual update because the TOC requires to be updated.


hamishwillee:
Nicely done. Thanks. 

Note/FYI, this is invisible if you navigate from https://dev.px4.io/ because that tree is not being updated :-( You have to use the version switcher to find the actual "master" and then can navigate to the new doc: 
dhttps://dev.px4.io/master/en/middleware/modules_driver_distance_sensor.html

@dagar Has a plan to fix this.

PR #743, opened 2019-02-18
propose to add PlotJuggler

facontidavide:
Hi, 

I added a plugin to my application to load data from ULog. It is somehow experimental but adding it to this list give me the opportunity to get more users and fix more bugs.

Cheers

Davide

### DIFF
--- a/en/log/ulog_file_format.md
+++ b/en/log/ulog_file_format.md
@@ -340,6 +340,8 @@ A valid ULog parser must fulfill the following requirements:
   via MAVLink.
 - [MAVGAnalysis](https://github.com/ecmnet/MAVGCL): Java, ULog streaming via
   MAVLink and parser for plotting and analysis.
+- [PlotJuggler](https://github.com/facontidavide/PlotJuggler): C++/Qt application to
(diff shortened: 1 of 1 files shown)


TSC21:
@facontidavide really cool! I usually use PJ with ROS and find it very powerful. Any docs on how to use it as well with uLog are welcomed.

facontidavide:
Simply open the file... then visualize data :)

image

TSC21:
Ahah sounds good!

bkueng:
Nice! We can also add it here: https://docs.px4.io/en/log/flight_log_analysis.html.

facontidavide:
That would be great. Do you need anything from me? A snapshot?

hamishwillee:
Thanks @facontidavide - merged.

> That would be great. Do you need anything from me? A snapshot?

A snapshot showing uLog plots and a short feature list would be great you can see the pattern in https://docs.px4.io/en/log/flight_log_analysis.html

MaEtUgR:
I opened a follow-up issue, see https://github.com/PX4/Devguide/issues/747#issuecomment-465918821

PR #746, opened 2019-02-19
ROS interface build and multi vehicle ports

julianoes:
A couple of small fixes. Check the commit messages for details.

### DIFF
--- a/en/simulation/ros_interface.md
+++ b/en/simulation/ros_interface.md
@@ -53,8 +53,7 @@ To run SITL wrapped in ROS the ROS environment needs to be updated, then launch 
 
 ```sh
 cd <Firmware_clone>
-make px4_sitl_default
-make px4_sitl_default sitl_gazebo
+DONT_RUN=1 make px4_sitl_default gazebo
 source ~/catkin_ws/devel/setup.bash    # (optional)
 source Tools/setup_gazebo.bash $(pwd) $(pwd)/build/px4_sitl_default
 export ROS_PACKAGE_PATH=$ROS_PACKAGE_PATH:$(pwd)
--- a/en/simulation/ros_interface.md
+++ b/en/simulation/ros_interface.md
@@ -53,7 +53,8 @@ To run SITL wrapped in ROS the ROS environment needs to be updated, then launch 
 
 ```sh
 cd <Firmware_clone>
-make px4_sitl_default gazebo
+make px4_sitl_default
+make px4_sitl_default sitl_gazebo
 source ~/catkin_ws/devel/setup.bash    # (optional)
 source Tools/setup_gazebo.bash $(pwd) $(pwd)/build/px4_sitl_default
 export ROS_PACKAGE_PATH=$ROS_PACKAGE_PATH:$(pwd)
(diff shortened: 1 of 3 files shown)


en/simulation/ros_interface.md
```diff
@@ -53,8 +53,9 @@ To run SITL wrapped in ROS the ROS environment needs to be updated, then launch
 
 ```sh
 cd <Firmware_clone>
-make px4_sitl_default gazebo
-source ~/catkin_ws/devel/setup.bash    // (optional)
+make px4_sitl_default
```
hamishwillee: Why is this step needed - ie isn't this done by the next step?

julianoes: I tried it and that's what I found

hamishwillee: @TSC21 Can you explain it? I'm happy to accept it, but I'd prefer to understand it first.

mhkabir: @dagar?

TSC21: This is not correct. This is only correct if Firmware is built under a catkin_ws, which is not the case. So this part itself needs to be reorganized.
There are 2 ways of handling this:
1. Firmware is built under a catkin_ws using catkin build tools, which then results on the binaries amd launch files to be put under the share folder of the same workspace. Then you just need to use the src cmd as bellow so to be able to use them;
2. As this guide implies, for using this with ROS but without the Firmware included on a catkin_ws, you need to manually export the env paths. By definition, this does not make much sense, since if you are using ROS, then you probably have a catkin_ws as a developer.

julianoes: So you're saying what I found by trial and error is not correct and I'm just confused? I guess I am now :smile:

TSC21: @juli

TSC21: Just to clarify: where's this catkin_ws being created in the first place? And why we are building with make inside a catkin_ws? There's no need for this at all. You just need to build using catkin build and then source the workspace. No need to export anything else.
There's no need to trial and error here. What I explained in the previous comment is the clear 2 ways of handling this.

julianoes: Trial and error is just me following the instructions of this guide and then realizing it's not working. If I'm doing the wrong thing I'm blaming this guide.

TSC21: Ok so then the guide needs to be updated.

julianoes: You tell me how. I don't know ROS things.

julianoes:
@hamishwillee @TSC21 I've updated this and now the instructions work. I would say we merge this because it's what should work properly with v1.9.

hamishwillee:
It looks good to me, but would prefer @TSC21 sanity checks!

hamishwillee:
FYI Nuno says he'll look at this tomorrow.

julianoes:
@TSC21 please review, otherwise I'll go ahead and merge it, thanks.

TSC21 (commented):
This is ok though if you are using a ROS envirnoment, you are usually expected to build the Firmware under catkin ws, under not mandatory. If you do it, none of these env variable setup is required, cause sitl_gazebo is built as a catkin_package and all the worlds and plugins are put under the dev dir. You just need to source the setup.bash of the env to use it.

julianoes:
Ok thanks. I'm not familiar with the way to use this in ROS, so it's up to you or others that use it that way to document it.

PR #760, opened 2019-03-29
BuildingPX4: Add info about how to format firmware version

hamishwillee:
From git tag, as per https://github.com/PX4/Firmware/pull/11730#issuecomment-477379841

@bkueng This is no "perfect" place for this, but IMO this is somewhere that people might reasonably find the information, and it should show up in search.


bkueng (approved):
Thanks!

PR #766, opened 2019-04-02
 fix some mistakes in the ulog documentation 

thomasetter:
The `hdr_size` is never in the `msg_size` as said at the top of this document:

> `msg_size` is the size of the message in bytes without the header (`hdr_size`= 3 bytes). 

This seems to also be how log data is read, e.g. in pyulog:
https://github.com/PX4/pyulog/blob/235ecaf26c67268d9b3d7b25ec94e50e34c60d90/pyulog/core.py#L479

The documentation is still a bit inconsistent with `ulog_message_flag_bits_s` not having a header as part of the struct, while e.g. `message_format_s` has one.

### DIFF
--- a/en/log/ulog_file_format.md
+++ b/en/log/ulog_file_format.md
@@ -106,7 +106,7 @@ struct message_header_s {
   ```
   struct message_format_s {
   	struct message_header_s header;
-  	char format[header.msg_size-hdr_size];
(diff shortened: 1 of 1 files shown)


bkueng (approved):
Thanks, that's correct.

PR #772, opened 2019-04-25
remove unnecessary symbols

IronClimber:
Remove unnecessary symbols in code in [Subscribing To Sensor Data](https://dev.px4.io/en/apps/hello_sky.html#subscribing-to-sensor-data) section.


bkueng (approved):
Thanks

PR #774, opened 2019-04-29
simulation: enable SDK port for HITL

julianoes:
This way jMAVSim forwards MAVLink traffic on port 14550 for QGC and 14540 for MAVROS or the Dronecode SDK.


bkueng (approved):
We should also change the diagram here: https://dev.px4.io/en/simulation/hitl.html#jmavsimgazebo-hitl-environment, where it says `Gazebo only`.

hamishwillee (commented):
And further to @bkueng message, there are a couple of other sentences in there that need to make it clear that offboard now works for jmavsim HITL

- "Gazebo can additionally share MAVLink data with an offboard API!" - suggest you just remove this note
- "(Optional - Gazebo only) Gazebo can also connect to an offboard API and bridge MAVLink messages to PX4." - move this above the other item (so optional stuff at bottom) and change to: "Gazebo and JMAVSim can also connect to an offboard API and bridge MAVLink messages to PX4."

hamishwillee:
Thanks! Made a minor change, but looks great to me. Merging.

PR #777, opened 2019-04-30
simulation: fix param name for RC lost action

julianoes:
NAV_DLL_ACT is for data link lost, not RC lost.


JonasVautherin (approved):
Good catch!

hamishwillee (approved):
Great. Thanks!

PR #778, opened 2019-04-30
simulation: improve HITL build instructions

julianoes:
This makes it clearer when we build what and why.


en/simulation/hitl.md
```diff
@@ -113,26 +113,24 @@ Follow the appropriate setup steps for your simulator in the following sections.
...
 
    > **Note** The serial device depends on what port is used to connect the vehicle to the computer (this is usually `/dev/ttyACM0`). An easy way to check on Ubuntu is to plug in the autopilot, open up a terminal, and type `dmesg | grep "tty"`. The correct device will be the last one shown.
 
-1. Close Gazebo, connect the flight controller to the computer and wait for it to boot.
-1. Run Gazebo in HITL mode 
+1. Set up the environment variables:
+   ```sh
+   source Tools/setup_gazebo.bash $(pwd) $(pwd)/build/px4_sitl_default
+   ```
+   and run Gazebo in HITL mode:
```
hamishwillee, line 131: I'd probably keep this as its own point, but up to you. 

I do find it a little confusing that running Gazebo in HITL mode is done by calling a SITL file? (`gazebo Tools/sitl_gazebo/worlds/iris.world`) - ie it isn't clear how this call starts Gazebo in HITL mode. Is it the environment variables that set the mode?

en/simulation/hitl.md
```diff
@@ -113,26 +113,24 @@ Follow the appropriate setup steps for your simulator in the following sections.
 
 > **Note** Make sure *QGroundControl* is not running!
 
-1. Update the environment variables: 
+1. Build PX4 with the Gazebo plugins although we only need the Gazebo part:
```
hamishwillee: What do you mean by "although we only need the Gazebo part"? Is this useful for people know?

The `DONT_RUN=1` change is good!

Perhaps 
```
1. Build (but do not run) PX4 and Gazebo:
```

en/simulation/hitl.md
julianoes: The settings whether HITL or SITL are set by the options in the iris.sdf file. It is quite confusing, I agree. And it's ugly because you need to manually edit the sdf file which is even auto-generated for the case of iris, so you need to edit it again after every build (or copy it from somewhere).

en/simulation/hitl.md
julianoes: We build SITL but we only need the build of the gazebo plugins.

hamishwillee: Not sure that answers my question - who cares? But assuming it is important, how about:

> 1. Build PX4 with Gazebo (in order to to build the Gazebo plugins).

en/simulation/hitl.md
```diff
@@ -113,26 +113,24 @@ Follow the appropriate setup steps for your simulator in the following sections.
...
+   DONT_RUN=1 make px4_sitl_default gazebo
    ```
-   In a new terminal, run:
-   ```sh
-   source Tools/setup_gazebo.bash $(pwd) $(pwd)/build/px4_sitl_default
-   ```
-  
 1. Open the vehicle model's sdf file (e.g. **Tools/sitl_gazebo/models/iris/iris.sdf**).
-1. Under the `mavlink_interface plugin` section, change the `serialEnabled` and `hil_mode` parameters to `true`. 
+1. Under the `mavlink_interface plugin` section, change the `serialEnabled` and `hil_mode` parameters to `true`.
```
hamishwillee, line 119: From your comment above https://github.com/PX4/Devguide/pull/778/files#r280022112 is it worth noting that the sdf file is autogenerated and that it needs to be re-edit or cached. 

I was thinking under the image below you might have
```
 ![HIL Parameters](../../assets/simulation/gazebo_sdf_model_hil_params.png)
> **Note** SDF files, including this one, may be autogenerated from XXXXX. If this is the case, then you will need to either keep a copy of your changed file or re-edit it for every build. 
```

en/simulation/hitl.md
julianoes: Yes that's better. I was just trying not to create confusion. So people might be wondering why SITL needs to be built if we're using HITL.

en/simulation/hitl.md
julianoes: Yes, I'll add that, thanks.

hamishwillee:
Thanks very much. Looks good. Merging!

PR #784, opened 2019-05-08
Add docs for vscode

hamishwillee:
VSCode now works for PX4. For more info see https://github.com/PX4/Firmware/pull/11978

@dagar This is very much draft. Questions inline


en/setup/vscode.md
```diff
@@ -0,0 +1,55 @@
+# Visual Studio Code IDE
+
+Visual Studio Code (VSCode) is a powerful cross-platform source code editor/IDE that can be used for PX4 development on Linux and macOS (Windows support coming soon).
+
+PX4 has been set up with all the definition files that you need to install needed VSCode extensions, and and start developing.
+This topic explains how to get started.
+
+<!--
```
hamishwillee: Some questions here. Normally it is good to explain why we're suggesting this IDE, in particular if there is a change - ie from Qt Creator
1. Are we _recommending_ this over Qt Creator, and if so why?

en/setup/vscode.md
```diff
@@ -0,0 +1,55 @@
...
+- Do we recommend VSCode?
+- what had to happen to make this setup work - ie presumably the directory contains definition files that vscode needs to know to set everything up?
+-->
+
+
+## Preconditions
+
+You must already have installed the command line [PX4 developer environment](../setup/dev_env.md) for your platform and downloaded the *Firmware* source code repo.
+
+<!-- xcode? -->
```
hamishwillee: Ignore this. If the macOS dev env says xcode, that will already be installed.

en/setup/vscode.md
```diff
@@ -0,0 +1,55 @@
...
+   - A file selection dialog will appear. 
+     Select the PX4 project directory (**Firmware**) from the file selection dialog to load the source.
+1. Press **Install All** on the *This workspace has extension recommendations* prompt (this will appear after a moment on the bottom right of the IDE).
+   ![Install extensions](../../assets/vscode/prompt_install_extensions.jpg)
+
+   VSCode will open the Extensions panel on the left hand side so you can watch the progress of installation.
+   ![PX4 loaded into VSCode Explorer](../../assets/vscode/installing_extensions.jpg)
+1. A number of notifications/prompts may appear in the bottom right corner (if these disappear, click the little "alarm" icon on the right of the blue bar at the bottom).
+   - If asked to install a new version of cmake you can say "no" on Ubuntu 1.18.04
+   - If prompted sign into github.com, do this if you wish to make pull requests directly from the IDE.
```
hamishwillee: Should we be recommending that they do this? 
Am I correct about what this does?

en/setup/vscode.md
```diff
@@ -0,0 +1,55 @@
...
+1. Select a kit. <!-- why/what does this do? -->
```
hamishwillee: Why do we have to select a kit? Presumably there is never a case where we'd use the other one? What does selecting this kit do for us?

en/setup/vscode.md
```diff
@@ -0,0 +1,55 @@
...
+   - You can launch the kit selection dialog by choosing *No Kit Selected* on the bottom bar (if a kit is already selected that will be displayed instead)
+     ![Select - No Kit Selected](../../assets/vscode/select_kit_start.jpg)
+   - Select *PX4 Detect* in the dialog:
+     ![Select PX4 Detect Kit](../../assets/vscode/select_kit.jpg)
+     
+## Build & Debugging
+
+To build or debug:
```
hamishwillee: Is the the right way to build and debug?

hamishwillee:
Some random notes:
1. I have this problem. Have install the SIM dependencies - gazebo and jmavsim but not nuttx
   image
1. In the bottom bar is px4, which has tooltip "select target to launch" and has options - test mixer multi_rotor or PX4". What are these for?
1. What do I have to do to build and run with gazebo - just press this?
   image
  I ask because nothing appears to happen. I wonder if it is just taking a very long time to do anything - this is in VM.
2. Can I run gazebo headless?

en/setup/vscode.md
dagar: It mostly comes down to richer extension ecosystem and simple in tree configuration.

We have good support for the main things like C/C++ (with solid cmake integration), but also python, jinja2, ROS messages, and even UAVCAN dsdl. Beyond that the important difference is that the configuration is simple and lives in the PX4/Firmware tree (https://github.com/PX4/Firmware/tree/master/.vscode). That's the somewhat mundane, yet critical difference that makes this a good solution for most people. It's possible to configure other editors and IDEs to do most of this, but this one is ready to go for a new user in minutes, not an afternoon of manual setup.

image

en/setup/vscode.md
dagar: Yes, although it can do everything, including reviewing pull requests with the diff shown in your local editor. I wouldn't necessarily recommend it one way or the other, it's just an optional thing that some people might like, and others probably won't.

en/setup/vscode.md
dagar: A kit is basically a cmake toolchain, which we currently handle entirely within PX4 cmake. I haven't yet been able to get rid of the prompt entirely. 

One example of a kit would be if you have multiple versions of a compiler installed locally. The kit would show you the different versions. Properly supporting cygwin might end up being a different kit. It's the same toolchain, but someone vscode needs to know that it's used within cygwin to build and run (TBD).

en/setup/vscode.md
dagar: Build yes, debug no. The Debug button will work, but it simply runs the executable. For almost any real debugging you need to kick it off from the Debug activity screen to also start the simulator or debug with hardware. 

image

These debug activities come from the launch.json file in PX4/Firmware. This will be expanded as needed. 
https://github.com/PX4/Firmware/blob/master/.vscode/launch.json

en/setup/vscode.md
dagar: I don't think you have to explicitly select a kit.

dagar:
> Some random notes:
> 
> 1. I have this problem. Have install the SIM dependencies - gazebo and jmavsim but not nuttx
>    image
> 2. In the bottom bar is px4, which has tooltip "select target to launch" and has options - test mixer multi_rotor or PX4". What are these for?
> 3. What do I have to do to build and run with gazebo - just press this?
>    image
>    I ask because nothing appears to happen. I wonder if it is just taking a very long time to do anything - this is in VM.
> 4. Can I run gazebo headless?

Let's followup on slack or skype.

en/setup/vscode.md
```diff
@@ -0,0 +1,55 @@
...
+## Preconditions
+
+You must already have installed the command line [PX4 developer environment](../setup/dev_env.md) for your platform and downloaded the *Firmware* source code repo.
+
+<!-- xcode? -->
+
+## Installation & Setup
+
+1. [Download and install VSCode](https://code.visualstudio.com/) (you will be offered the correct version for your OS).
+1. Add the PX4 source code into VSCode.
```
dagar: Let's keep it simple and only recommend opening the folder. The workspace allows you to have a collection of projects and I don't think we need to go into it for the vast majority.

en/setup/vscode.md
```diff
@@ -0,0 +1,55 @@
...
+     ![Menu - Open Folder/Add Folder to Workspace](../../assets/vscode/menu_file_open_folder.jpg)
+   - A file selection dialog will appear. 
+     Select the PX4 project directory (**Firmware**) from the file selection dialog to load the source.
+1. Press **Install All** on the *This workspace has extension recommendations* prompt (this will appear after a moment on the bottom right of the IDE).
+   ![Install extensions](../../assets/vscode/prompt_install_extensions.jpg)
+
+   VSCode will open the Extensions panel on the left hand side so you can watch the progress of installation.
+   ![PX4 loaded into VSCode Explorer](../../assets/vscode/installing_extensions.jpg)
+1. A number of notifications/prompts may appear in the bottom right corner (if these disappear, click the little "alarm" icon on the right of the blue bar at the bottom).
+   - If asked to install a new version of cmake you can say "no" on Ubuntu 1.18.04
```
dagar: Ubuntu version typo.

dagar: I think it's also fine to not bother mentioning this. In most setups it's not going to matter either way. It's the edge cases (that we don't really "support") where the answer is going to be "depends".

en/setup/vscode.md
```diff
@@ -0,0 +1,55 @@
...
+   - If prompted sign into github.com, do this if you wish to make pull requests directly from the IDE.
+1. Select a kit. <!-- why/what does this do? -->
+   - You can launch the kit selection dialog by choosing *No Kit Selected* on the bottom bar (if a kit is already selected that will be displayed instead)
+     ![Select - No Kit Selected](../../assets/vscode/select_kit_start.jpg)
+   - Select *PX4 Detect* in the dialog:
+     ![Select PX4 Detect Kit](../../assets/vscode/select_kit.jpg)
+     
+## Build & Debugging
```
dagar: I think we might want to mention the config selection (and possibly build earlier). In order for the intellisense magic to work (code completion, etc) you need an active config and to have at least generated things like the uORB messages.

dagar: Screenshot from 2019-05-09 13-45-22

en/setup/vscode.md
hamishwillee: Agreed. Fixed

hamishwillee:
Should I do instructions in this notif. ?

image

hamishwillee:
I presume I have to wait for this prompt to finish before doing anything else?

image

en/setup/vscode.md
hamishwillee: I'm pretty sure this appears as soon as you press the debug button. However you don't have to do it here, so I won't.

en/setup/vscode.md
hamishwillee: You say "For almost any real debugging you need to kick it off from the Debug activity screen to also start the simulator or debug with hardware."
Isn't that the screen you get to by pressing the debug sidebar:

image

I'm a bit confused, because, if so, I don't know how you can press the button without being on this screen.

Remember none of this actually works on windows so I am guessing in a VM that keeps crashing.

en/setup/vscode.md
hamishwillee: Personally I never know if they are important or can be ignored, so it is worth mentioning them. At least minimally with "All other notifications are optional".
I can do that if you prefer?

en/setup/vscode.md
hamishwillee: When you say active config, you mean that this has a value right?
image

Presumably people will just build the whole code, not just uorb to get the code completion?

Anyway, I've taken a guess at this and just added a section called code completion here.

hamishwillee:
@dagar Might be faster/better if you create a little walkthrough video showing setup and basic debugging features? Then at least I'd know I wasn't missing the obvious. Alternatively walk me through via screensharing?

hamishwillee:
GDB support:
- Is this just a matter of attaching your debug probe to the device and selecting the correct config in vscode. If not, what else? Ie does board need some sort of magic mode settings or whatever?
- What probes have been tried.
- Is there any difference than SITL debugging?

hamishwillee:
A quick status update:
1. I can get intellisense working. Essentially after you set the config you have to build once. The key is that configuration takes a long time. If you try build before it completes things may fall over, but in any case it feels like nothing is happening. Action is to add notes that you need to be patience.
2. I still can't get the debugging to work on SITL. Specifically, can't launch the simulator.

en/setup/vscode.md
```diff
@@ -0,0 +1,83 @@
...
+     ![Cmake config project](../../assets/vscode/cmake_configuring_project_done.jpg).
+1. You can then kick off a build from the config bar (select either **Build** or **Debug**).
+   ![Run debug or build](../../assets/vscode/run_debug_build.jpg)
+
+After building at least once you can now use [code completion](#code completion) and other *VSCode* features.
+
+<!-- px4 at the bottom - select target to launch - test mixer multi_rotor or PX4 -->
+
+## Debugging PX4 {#debugging_sitl}
+
```
hamishwillee, line 67: @dagar This is the bit that gives me trouble - I don't know how to select and executable and make it launch SITL. I would also like to be able to make the launch headless.

image

PR #785, opened 2019-05-10
Update port GCS connection

gamoreno:
The port on the PX4 side of the connection with the GCS has changed from 14556 to 14570

### DIFF
--- a/en/test_and_ci/docker.md
+++ b/en/test_and_ci/docker.md
@@ -93,7 +93,7 @@ docker run -it --privileged \
     -v <host_src>:<container_src>:rw \
     -v /tmp/.X11-unix:/tmp/.X11-unix:ro \
     -e DISPLAY=:0 \
-    -p 14556:14556/udp \
+    -p 14570:14570/udp \
     --name=<local_container_name> <container>:<tag> <build_command>
 ```
 Where,
@@ -114,7 +114,7 @@ sudo docker run -it --privileged \
 -v ~/src/Firmware:/src/firmware/:rw \
 -v /tmp/.X11-unix:/tmp/.X11-unix:ro \
 -e DISPLAY=:0 \
--p 14556:14556/udp \
+-p 14570:14570/udp \
 --name=mycontainer px4io/px4-dev-ros:2017-10-23 bash
 ```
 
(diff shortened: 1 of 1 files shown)


hamishwillee:
@gamoreno Thanks. Looking at the helper script [docker_run.sh](https://github.com/PX4/Firmware/blob/master/Tools/docker_run.sh) it still uses 14556 for the publish port - is there a reason for the mismatch? Also it takes far more recent date images - ie 2019-03-08 vs 2017-10-23

@dagar Thoughts?

gamoreno:
> @gamoreno Thanks. Looking at the helper script [docker_run.sh](https://github.com/PX4/Firmware/blob/master/Tools/docker_run.sh) it still uses 14556 for the publish port - is there a reason for the mismatch? Also it takes far more recent date images - ie 2019-03-08 vs 2017-10-23
> 
> @dagar Thoughts?

Good catch. The port in docker_run.sh is wrong too. The same problem is in the base Dockerfiles that should expose these ports:
[Dockerfile_base-archlinux](https://github.com/PX4/containers/blob/master/docker/Dockerfile_base-archlinux#L81)
[Dockerfile_base-bionic](https://github.com/PX4/containers/blob/master/docker/Dockerfile_base-bionic#L115)
[Dockerfile_base-xenial](https://github.com/PX4/containers/blob/master/docker/px4-dev/Dockerfile_base-xenial)

In these files, the port for the offboard is also wrong, it should be 14580 instead of 14557

hamishwillee:
@gamoreno Can you make PR for those files? If it gets accepted then I can just copy the pattern.

hamishwillee:
@julianoes Can you sanity check this? My understanding is that the ports like 14556 are "internal" - e.g. should not be relied upon: https://dev.px4.io/master/en/simulation/#sitl-simulation-environment

But then I don't know this script, so perhaps it is correct.

julianoes (approved):
Unfortunately this is correct.

Edit: I don't like the numbering scheme but it is what it is.

hamishwillee:
Thanks @gamoreno , @julianoes - merging then. Would be good to get the docker containers updated "hint hint"

PR #788, opened 2019-05-14
Update ulog_file_format.md

Marxlp:
`message_header_s` is also in the `ulog_message_info_multiple_header_s` data struct.


hamishwillee:
Thanks, looks consistent to me! @bkueng FYI!

bkueng:
Yup, thanks

PR #789, opened 2019-05-14
Update ulog_file_format.md

Marxlp:
`ulog_message_flag_bits_s` also needs `message_header_s` struct.

### DIFF
--- a/en/log/ulog_file_format.md
+++ b/en/log/ulog_file_format.md
@@ -72,6 +72,7 @@ struct message_header_s {
 - 'B': Flag bitset message.
   ```
   struct ulog_message_flag_bits_s {
+  	struct message_header_s;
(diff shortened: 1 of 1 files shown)


hamishwillee:
@bkueng 

The log format docs got a bit inconsistent. Up the top it says "The Definitions and Data sections consist of a stream of messages.
Each starts with this header:"

Then in some cases the header got added, and in others it did not. I think it might be worth adding the header everywhere. Otherwise we just close this, and remove it everywhere. Thoughts?

bkueng:
Yes we should add it everywhere.

hamishwillee:
Thanks @Marxlp merged - sorry for the delay. @bkueng I was wrong, this was the last one that hadn't been updated so no further work required.

PR #807, opened 2019-05-30
hellosky:clarify benefits of shared libraries

hamishwillee:
Follows on from #663. @m-ou-se  - can you please check?


m-ou-se:
Maybe `loaded` instead of `swapped` would be better. 'Swapped' could give the impression that they can be unloaded at runtime, but that's not the case (right now).

hamishwillee:
Thanks @m-ou-se - yes, that is a useful clarification. I have updated and merged.

PR #811, opened 2019-06-04
Update to use wget instead of browser

unipheas:
I spent about a week reinstalling operating systems and running this script to later find that by downloading from the browser some characters were being rewritten essentially corrupting the file, thus failing the install. After consulting @julianoes I found that if I wget the file then it wouldn't cause issues and everything would install perfectly, the first time.

### DIFF
--- a/en/setup/dev_env_linux.md
+++ b/en/setup/dev_env_linux.md
@@ -33,7 +33,7 @@ Then follow the instructions for your development target in the sections below.
 
 To install the development toolchain:
 
-1. Download <a href="https://raw.githubusercontent.com/PX4/Devguide/{{ book.px4_version }}/build_scripts/ubuntu_sim_nuttx.sh" target="_blank" download>ubuntu_sim_nuttx.sh</a>.
+1. Download <a href="https://raw.githubusercontent.com/PX4/Devguide/{{ book.px4_version }}/build_scripts/ubuntu_sim_nuttx.sh" target="_blank" download>ubuntu_sim_nuttx.sh</a>. It's best to use `wget`.
 1. Run the script in a bash shell:
    ```bash
    source ubuntu_sim_nuttx.sh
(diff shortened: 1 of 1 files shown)


en/setup/dev_env_linux.md
```diff
@@ -33,7 +33,7 @@ Then follow the instructions for your development target in the sections below.
 
 To install the development toolchain:
 
-1. Download <a href="https://raw.githubusercontent.com/PX4/Devguide/{{ book.px4_version }}/build_scripts/ubuntu_sim_nuttx.sh" target="_blank" download>ubuntu_sim_nuttx.sh</a>.
+1. Download <a href="https://raw.githubusercontent.com/PX4/Devguide/{{ book.px4_version }}/build_scripts/ubuntu_sim_nuttx.sh" target="_blank" download>ubuntu_sim_nuttx.sh</a>. It's best to use `wget`.
```
julianoes: I would add the actual command:
```
1. Download the script:
    ```bash
    wget https://raw.githubusercontent.com/PX4/Devguide/{{ book.px4_version }}/build_scripts/ubuntu_sim_nuttx.sh
    ```

Comment (unipheas) [reply en/setup/dev_env_linux.md]: By placing the command in a code block how are we able to use interpolation?
```bash
wget https://raw.githubusercontent.com/PX4/Devguide/{{ book.px4_version }}/build_scripts/ubuntu_sim_nuttx.sh
```

julianoes: I don't know if the gitbook magic can do it @hamishwillee ?

unipheas: I'm not seeing anything like that in their [documentation](https://docs.gitbook.com/content-editing/markdown)

The link points to master, could we just make that static if we can't use interpolation?

hamishwillee: There is no way to do the interpolation in a code block. The link points to master for this build, but if you look at book.json where it is defined, you will see that on 1.9.0 it points to the 1.9.0 branch. That is why my trees and Firmware match, and why I did it the way that I did rather than wget.

So, for this case, suggest just use "master" with no interpolation because in the very near future I hope to replace all this with scripts in the firmware tree. Then I could just say "sync Firmware via git, then run the file ....".

hamishwillee:
Note also that I have never had the problem you are describing from my browser - using Chrome on Ubuntu. I just click download link.

unipheas:
> Note also that I have never had the problem you are describing from my browser - using Chrome on Ubuntu. I just click download link.

Understood. I'm going through various scenarios for installation to ensure devs that want to get into this aren't running into problems.

hamishwillee:
THanks @unipheas - merged. Not completely happy that we can no longer fetch the "right" version, but as I said, should be less of a problem in future.

unipheas:
Thank you!!!

PR #812, opened 2019-06-05
debug: make C example C++ compatible

julianoes:
As raised in https://discuss.px4.io/t/debugging-in-sih/11243/7.


hamishwillee (commented):
Text is OK, but added @bkueng for technical review.

bkueng (approved):

PR #813, opened 2019-06-09
Add def of message_logging_tagged_s

devbharat:
Refer to https://github.com/PX4/pyulog/pull/43 and https://github.com/PX4/Firmware/pull/12144

### DIFF
--- a/en/log/ulog_file_format.md
+++ b/en/log/ulog_file_format.md
@@ -259,6 +259,29 @@ The following messages belong to this section:
 | INFO       |      '6'     | Informational                        |
 | DEBUG      |      '7'     | Debug-level messages                 |
 
+- 'C': Tagged Logged string message
+  ```
+  struct message_logging_tagged_s {
+    struct message_header_s header;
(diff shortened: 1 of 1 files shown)


bkueng (commented):
Can you add the `tag`s that you use, so it's well-defined?
I suggest to start with `0: not specified`, `1: autopilot general`, ...

devbharat:
> Can you add the tags that you use, so it's well-defined?

Might not be very relevant to others. They are names of different processes(handling different system components like different cameras, PPK, Pixracer, Disk etc) that generate those log strings. I could add suggestions as an example.

bkueng:
Yes please add them. It increases the chance that we get consistent tooling.

hamishwillee:
@bkueng FYI This one ready for your re-review.

bkueng (approved):
Thanks

bkueng:
Can you rebase? Looks like there are conflicts.

hamishwillee:
Thanks. Sorry for late merge!

PR #814, opened 2019-06-09
Add sync bytes

devbharat:
Refer to https://github.com/PX4/pyulog/pull/44 and https://github.com/PX4/Firmware/pull/12144


bkueng (approved):
Thanks!

PR #817, opened 2019-06-17
Fixed broken link.

dinomani:


hamishwillee:
Thanks Dino. FYI, this will appear in the master branch of the docs only in about 5 mins. That means other branches will be incorrect until we make a new release. For "important/big picture" stuff I will sometimes also move the fix into stable release, but probably not for this case.

PR #818, opened 2019-06-18
Adding VTOL Octoplane airframe type

irsdkv:
At this time some VTOLs could have 8 motors for MC mode.
I think that the ability to develop their support will be more convenient with the appropriate class.

Corresponding PR in Firmware: https://github.com/PX4/Firmware/pull/12303
Corresponding PR in User guide: https://github.com/PX4/px4_user_guide/pull/503
Corresponding PR in QGroundcontrol: https://github.com/mavlink/qgroundcontrol/pull/7537


en/airframes/airframe_reference.md
```diff
@@ -1052,3 +1052,20 @@ This page lists all supported airframes and types including
 </tr>
 </tbody></table>
 
+### VTOL Octoplane
```
hamishwillee: Please remove all the airframe_reference changes in all languages - this stuff is autgenerated from source so we just need the image.

More generally, and FYI only, we never edit the translations. That stuff is auto exported from our translation tool crowdin.

irsdkv: All done
Did not know this, thanks for explaining :)

hamishwillee: Thanks very much. It is documented, but very easily missed given how many other things you need to think about : https://dev.px4.io/master/en/airframes/adding_a_new_frame.html#adding-a-new-airframe-group

PR #819, opened 2019-06-21
included jMAVSim display-only documentation

romain-chiap:
I just updated the SIH documentation to include setting up jMAVSim to display the quad in SIH mode.

### DIFF
--- a/en/simulation/simulation-in-hardware.md
+++ b/en/simulation/simulation-in-hardware.md
@@ -39,7 +39,19 @@ When the SIH airframe is selected, the SIH module starts by itself, the vehicle 
 
 ## Setting up the Display
 
-The simulated quadrotor can be displayed in jMAVSim (coming soon).
+The simulated quadrotor can be displayed in [jMAVSim](https://github.com/PX4/jMAVSim). For that, start by [installing](https://github.com/PX4/jMAVSim#installation) the last jMAVSim version. 
+Then, 
+1. Plug the autopilot hardware to the computer. The airframe should be configured as **SIH Quadcopter X**.
+1. Close QGroundControl (or any ground control station).
+1. Execute jMAVSim with the `-disponly` flag. The full command line should look like 
+   - `java -cp lib/\*:out/production/jmavsim.jar me.drton.jmavsim.Simulator -disponly` 
+   - (on Windows replace `:` by `;`). 
+   - Use `-h` to display the help. 
+   - You may need to specify the serial port name and speed, such as `-serial /dev/ttyACM0 230400` for instance.
+
+> **Note** If you want to use QGroundControl alongside jMAVSim, this is possible, but jMAVSim needs to be started first. You may also need to add the `-qgc` flag when executing jMAVSim.
+
+jMAVSim should display the vehicle starting on the ground. You can then arm and fly normally.
 
 ## Credits
 
--- a/en/simulation/simulation-in-hardware.md
+++ b/en/simulation/simulation-in-hardware.md
@@ -39,7 +39,7 @@ When the SIH airframe is selected, the SIH module starts by itself, the vehicle 
 
 ## Setting up the Display
 
-The simulated quadrotor can be displayed in [jMAVSim](https://github.com/PX4/jMAVSim). For that, start by [installing](https://github.com/PX4/jMAVSim#installation) the last jMAVSim version. 
+The simulated quadrotor can be displayed in [jMAVSim](https://github.com/PX4/jMAVSim). For that, start by [installing](https://github.com/PX4/jMAVSim#installation) the latest jMAVSim version. 
 Then, 
 1. Plug the autopilot hardware to the computer. The airframe should be configured as **SIH Quadcopter X**.
 1. Close QGroundControl (or any ground control station).


en/simulation/simulation-in-hardware.md
```diff
@@ -39,7 +39,19 @@ When the SIH airframe is selected, the SIH module starts by itself, the vehicle
 
 ## Setting up the Display
 
-The simulated quadrotor can be displayed in jMAVSim (coming soon).
+The simulated quadrotor can be displayed in [jMAVSim](https://github.com/PX4/jMAVSim). For that, start by [installing](https://github.com/PX4/jMAVSim#installation) the last jMAVSim version.
```
bkueng:
```suggestion
The simulated quadrotor can be displayed in [jMAVSim](https://github.com/PX4/jMAVSim). For that, start by [installing](https://github.com/PX4/jMAVSim#installation) the latest jMAVSim version. 
```

en/simulation/simulation-in-hardware.md
```diff
@@ -39,7 +39,19 @@ When the SIH airframe is selected, the SIH module starts by itself, the vehicle
 
 ## Setting up the Display
 
-The simulated quadrotor can be displayed in jMAVSim (coming soon).
+The simulated quadrotor can be displayed in [jMAVSim](https://github.com/PX4/jMAVSim). For that, start by [installing](https://github.com/PX4/jMAVSim#installation) the last jMAVSim version. 
+Then, 
+1. Plug the autopilot hardware to the computer. The airframe should be configured as **SIH Quadcopter X**.
+1. Close QGroundControl (or any ground control station).
+1. Execute jMAVSim with the `-disponly` flag. The full command line should look like 
+   - `java -cp lib/\*:out/production/jmavsim.jar me.drton.jmavsim.Simulator -disponly`
```
bkueng, line 47: I prefer to use [this script](https://github.com/PX4/Firmware/blob/master/Tools/jmavsim_run.sh), as it's our canonical way to start jmavsim (and extend it as needed).

hamishwillee: @romain-chiap Did you want to fix in line with this suggestion? Then we can get this merged!

hamishwillee: @romain-chiap ^^^ Are you still working on this? Would be great to get this fix in

romain-chiap: I will look into that in the coming two weeks...

hamishwillee: Thanks very much!

romain-chiap: I added an option "o" for display Only, in the [script](https://github.com/PX4/Firmware/blob/master/Tools/jmavsim_run.sh), which adds the argument "-disponly" to call jmavsim. 

jmavsim does not seem to connect to the hardware... I'm still investigating that.

hamishwillee: @romain-chiap Thanks. So the launch line above would have to change to use the script then (or that is how I interpret Beat's note)

romain-chiap: in jmavsim_run.sh, I added an entry in the case:
```
o)
	extra_args="$extra_args -serial /dev/ttyACM0 230400 -disponly"
	;;
```
However, this code will only work on Linux, on MacOS it should be something like
`/dev/tty.usbserial-ABCBD`

@bkueng Is there a systematic way to call the correct serial path ?

bkueng: There's a `-d <device>` argument for that.
For Linux we could look for `/dev/serial/by-id/*_PX4_*`, but I don't think that works for MacOS either.

romain-chiap: Thanks! That seems to work on Linux, I just tried it.

I can't try on MacOS I don't have access to a computer with this OS.
Shall I do a pull request as it is?

bkueng: I checked and the Firmware is using different ports on MacOS, so it will not work there. Also the default is to use TCP, so I'm not sure it's worth auto-detecting it.

romain-chiap: Well according to the [documentation](https://dev.px4.io/v1.9.0/en/simulation/hitl.html#jmavsim-quadrotor-only), the HIL uses the serial port. 
1. I could do something similar where we leave it to the user to specify the serial port
```
./Tools/jmavsim_run.sh -q -d /dev/ttyACM0 -b 921600 -r 250 -o
```
in that case the flag `-o` only adds `-disponly`. (I tested it that way, this is a feasible solution).

2. Can we use tcp to communicate between the autopilot and jmavsim? (It didn't work right away with the `-tcp` flag when calling jmavsim).

bkueng: Exactly, 1. is what we can do.

> Can we use tcp to communicate between the autopilot and jmavsim?

You need someone to connect to the UART, except if you have a network-enabled autopilot (i.e. not a Pixhawk).

romain-chiap: I created two pull requests:
- [PX4 code, jmavsim_run.sh edited](https://github.com/PX4/Firmware/pull/13931)
- [Devguide, merged already](https://github.com/PX4/Devguide/pull/918)

So the current PR can be closed.

A big thanks for the support throughout this project !

Issue #823, opened 2019-06-26
Document multi-uav sim with sdf files


lamping7:
Need to add dependency: **xmlstarlet** and a brief description due to https://github.com/PX4/Firmware/pull/12306

hamishwillee:
That sounds cool - do you plan on doing this doc update?

Is this a new doc or an extension of the old one? https://dev.px4.io/master/en/simulation/multi-vehicle-simulation.html structure? 

Note that I don't fully understand the solution and how it intersects with the old solution (ie what is the same and what is is different - which is why I don't know if this is an addition to the old doc or a completely alternative option). 
(I THINK that the essential difference is that instead of hard coding the port for each model in the xacro you have a single definition and use this xmlstarlet thing to create multiple mavlink_tcp_port thingies. I think this is a solution if the vehicle is the same in each case, but won't work if you you need multiple types of vehicles - presumably because you're copying from a single model. I think that most of the other things to do the setup are the same - Ie you will still need  rcS startup files for each vehicle. 
But all this is a guess.

What I do know is that we will need to add the limitations from the linked firmware issue and perhaps expand a little on what they mean to end users. If this is a new doc, it should compare and contrast to the other solution. 

It would also be good to test on Ubuntu 18.04 and ROS Melodic as this is our new default target for CI.

PR #825, opened 2019-06-29
simulation: add multi uav with sdf

lamping7:
Added some words on https://github.com/PX4/Firmware/pull/12306 for https://github.com/PX4/Devguide/issues/823. @hamishwillee fell free to send me a note on slack. I'm sure more info here could be helpful, but It will be easier to know what else to include after you prompt me with questions.


hamishwillee:
@lamping7 Thank you for this. I'll add comments here. As a reader knowing very little about ROS and gazebo I found this confusing - all these questions occurred to me/I could not answer from the update:
- What is an SDF Model. 
- How does an SDF differ from a xacro and a urdf model?
- How do I use this?
- What are limitations?
- How do I extend this for more vehicles.

What I "guess":
- A xacro can  be used to define a urdf model. A URDF model and an SDF model are both ways of defining a gazebo model? how do they differ?
- The current doc uses a xacro to define the model and generate a URDF. It does this for each vehicle defined in the xacro and feeds in the sitl mavlink UDP port for each case. 
- The new approach described here uses the normal SDF model for a vehicle - e.g.   a plane. Instead of creating a new urdf model from the xacro definition the [multi_uav_mavros_sitl_sdf.launch](https://github.com/PX4/Firmware/blob/master/launch/multi_uav_mavros_sitl_sdf.launch) script launches the normal plane model by calling the `single_vehicle_spawn_sdf.launch` 2 times and feeds in the correct port in each case.
- The limitation is that the launch file only allows you to specify one SDF, so just one vehicle type. 
- A benefit of this limitation is that you wouldn't need multiple rCS startup files right? 

So to use this the steps would be exactly the same except you must install xmlstarlet  and step 3 would instead be:
```
multi_uav_mavros_sitl_sdf.launch
```
Right?

Might be helpful to duplicate the section that starts "The launch file multi_uav_mavros_sitl.launchdoes the following," with the new launch file? You could then explain exactly how it works and how you extend it to have more items. 

Sorry if I am being dim.

lamping7:
Those are good points/questions. Let me try to cover them briefly and we can continue to discuss what I'm missing here.
> What is an SDF Model.
How does an SDF differ from a xacro and a urdf model?

This is described by Gazebo. Basically, URDF is standard for ROS and SDF was created to overcome the limitations of URDF in being able to describe additional features for the simulation and robots. We already link to this page that describes this. http://gazebosim.org/tutorials/?tut=ros_urdf There's also this one if you think we should include it http://sdformat.org/ I'm not inclined to repeat this info. I'd rather let the sources explain it.

> How do I use this?

You got it. `roslaunch multi_uav_mavros_sitl_sdf.launch vehicle:=<model_file_name>` This works the same as the existing one for xacro models, but instead of sending the model through xacro to fill in the arguments, we're passing the model to xmlstarlet to change the value of the element. Everything else is the same in the launch files. The only difference between these guys is the `-sdf` argument to the gazebo node and this line: 
`<arg name="cmd" default="xmlstarlet ed -d '//plugin[@name=&quot;mavlink_interface&quot;]/mavlink_tcp_port' -s '//plugin[@name=&quot;mavlink_interface&quot;]' -t elem -n mavlink_tcp_port -v $(arg mavlink_tcp_port) $(find px4)/Tools/sitl_gazebo/models/$(arg vehicle)/$(arg vehicle).sdf"/>`

I can point out this little difference if you think we should. The result is 3 vehicles, just like the other launch file (multi_uav_mavros_sitl.launch). So, I'm not sure if a whole "does the following" is worth doing as it will be mostly a repeat. 

> What are limitations?

This doesn't really deal with this, but PX4 doesn't support different vehicles at once. It really isn't setup for multi-vehicle as is with the current structure how run-time data is setup. https://github.com/PX4/Firmware/issues/9046 I did this as a hack to fix this for my use https://github.com/lamping7/Firmware/commit/15325ffbceb2375d2ecb03212a80e290ad811310

> How do I extend this for more vehicles.

The same exact way it's done for the for the other one.

Additionally, TCP not UDP and the startup structure changed a bit not too long ago so you don't need make rCS copies as it can dynamically setup the ports there.

hamishwillee:
HI @lamping7 

Thanks for that. I think I'm getting my head around this a bit better.

A simpler set of questions:

> To launch multiple of the same SDF formatted models, rather than xacro ones, use multi_uav_mavros_sitl_sdf.launch. 

**Why** would you use this approach rather than the xacro approach? 

From what you've said you can't do multiple different vehicle types as a PX4 benefit so it isn't that.

Looking at this, the main difference would be ease of extending. IN the old way you needed to take copy of an existing iris rcS startup file but in the new version you don't need to.  Also now you don't need to specify port numbers in the model, which seems a cleaner model interface.

So to update this I would answer the above question, and show exactly how this is launched:
```
roslaunch multi_uav_mavros_sitl_sdf.launch vehicle:=<model_file_name> 
```

I'm not completely convinced that extending to another vehicle is exactly the same as for the other approach. After all, you've already said you don'tneed the step "create a startup file, and change the file as follows:"

So perhaps I'd explicitly state that.

lamping7:
I understand what you're saying. Perhaps a larger rewrite and structure changes would explain all of this better as the current version has some outdated info that confuse this more.

The simple answer to your why is: sitl_gazebo, and by extension px4, only offer one xacro model. People often what to use the sdf ones (which there are many) and this new launch file let's them do that the same way the existing one did.

I'll work on a larger change to this page.

hamishwillee:
> The simple answer to your why is: sitl_gazebo, and by extension px4, only offer one xacro model. People often what to use the sdf ones (which there are many) and this new launch file let's them do that the same way the existing one did.

We can go in circles here forever - what is the benefit of having multiple SDF files as opposed to a sincle xacro? I'm guessing that this allows you to do something that having just one Xacro file does not. Or it could be that it is just a better abstraction of a vehicle ... or whatever.

Note that it is perfectly possible that there is no real benefit - people just "prefer" to do it this way. If that is the case we could just say that. 

> I'll work on a larger change to this page.

That would be good. The structure of the current page works for me, but if it is not accurate then it does not form a good foundation for this addition.

hamishwillee:
PS, sorry to be so picky. It is just that I see my job as being that of a newbie coming to the doc. If I don't understand the point, then no one will.

lamping7:
> We can go in circles here forever - what is the benefit of having multiple SDF files?

I don't mind. And I don't know. All I know is that we have sdf models and people wanted to simulate multiple of them and this allows you to do that.

hamishwillee:
In that case, we just say that.  For this section we can do something like the text below (what do you think?). Note this still has a TBD for extending the model, and I still think it would be great if you could fix any errors in the current method text first).

----

## Multiple Vehicles using SDF Models

Developers can alternatively simulate multiple vehicles using vehicle models defined in Gazebp SDF files (instead of using models defined in the ROS Xacro file, as discussed in the rest of this topic).

The steps are:
1. Install *xmlstarlet* from your Linux terminal
   ```
   sudo apt install xmlstarlet
   ```
1. Use *roslaunch* with the **multi_uav_mavros_sitl_sdf.launch** launch file:
   ```bash
   roslaunch multi_uav_mavros_sitl_sdf.launch vehicle:=<model_file_name> 
   ```
   Note that the vehicle model file name parameter is optional; the [plane model](https://github.com/PX4/sitl_gazebo/tree/master/models/plane) will be used by default.

This method is similar to using the xacro except that the SITL/Gazebo port number is automatically inserted by *xmstarlet* for each spawned vehicle, and does not need to be specified in the SDF file. 

To add a new vehicle, simply ... (TBD!)

hamishwillee:
@lamping7 

Did suggestion above for docs make sense? https://github.com/PX4/Devguide/pull/825#issuecomment-508336772

The only bit that would then need to be added would be explanation of how to add a new vehicle.

I'd still like fixes to the rest of the doc, but I think with this structure perhaps this could go in and the other fixes could be done separately? Will be led by you!

lamping7:
Have at look at the most recent commit @hamishwillee

hamishwillee:
Thank you @lamping7 - much appreciated, as usual! I will do some tidy up of the markdown as a post process now I have merged.

My assumption is that you will, at your convenience, do further updates to tidy errors you know about in the rest of the doc :-)

tensarflow:
Thank you soooo much for this PR @lamping7 !! :)))
That saved me so much workarounds.. You guys are really profis. I really appreciate your work!

PR #828, opened 2019-07-01
Installs PYYAML 3.13 instead of 5.1

ghost:
In PYYAML 5.1 yaml.load is deprecated and add FullLoader and UnsafeLoader classes. The script ubuntu_sim_common_deps.sh throws an error on that.

### DIFF
--- a/build_scripts/ubuntu_sim_common_deps.sh
+++ b/build_scripts/ubuntu_sim_common_deps.sh
@@ -26,7 +26,7 @@ which xxd || sudo apt install xxd -y || sudo apt-get install vim-common --no-ins
 # Required python packages
 sudo apt-get install python-argparse python-empy python-toml python-numpy python-dev python-pip -y
 sudo -H pip install --upgrade pip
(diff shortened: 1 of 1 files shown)


hamishwillee:
@dagar This forces a specific lower version of Pyaml. We don't do this in the tested setup script: https://github.com/PX4/Firmware/blob/master/Tools/setup/requirements.txt#L11

1. Do we need to?
2. My understanding is that this script is tested for builds regularly. Is that the case? If this is an error, how did it get through the tests.

ghost:
I am getting this error when launching the SITL simulation.

yaml_error

Installing PYYAML==3.13 solves the issue.

mrivi:
@telegek @hamishwillee That's a MAVROS problem. It's fixed on master and it will be available in the next realese 0.32.0

ghost:
> @telegek @hamishwillee That's a MAVROS problem. It's fixed on master and it will be available in the next release 0.32.0

Great, thanks.

hamishwillee:
Thanks Martina!

PR #832, opened 2019-07-05
Added Arch Linux scripts.

Pedro-Roque:
Self explanatory title. Converted the Ubuntu scripts to Arch Linux. 

I had previously installed ros-melodic-desktop from AUR, that's the only modification to the system.

@hamishwillee should I add a page on Docs explaining this as well?

### DIFF
--- a/build_scripts/arch_sim.sh
+++ b/build_scripts/arch_sim.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+
+## Bash script for setting up a PX4 development environment on Arch Linux (as of July 5th, 2019)
+## It can be used for installing simulators (only) or for installing the preconditions for Snapdragon Flight or Raspberry Pi.
+##
+## Installs:
+## - Common dependencies libraries and tools as defined in `arch_sim_common_deps.sh`
+## - Gazebo10 simulator
+
+echo "Downloading dependent script 'arch_sim_common_deps.sh'"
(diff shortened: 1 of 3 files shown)


hamishwillee:
Hi @Pedro-Roque 

Thanks for this, but I don't think we want to take this change. The reason is that is v1.9 we had scripts like this, but in master we have just moved to using scripts that are maintained as part of our docker configuration - you can see these linked from the docs here: https://dev.px4.io/master/en/setup/dev_env_linux_arch.html#script-based-installation

This is better than having scripts in the devguide itself because they get tested with every PR submission etc.

If however you think that there is a problem with our docker scripts would be VERY happy to take fixes there :-)

Does that make sense?

Pedro-Roque:
I totally forgot to check this, I thought there was still no support for rolling distros!

In that case, my only complaint is the use of yaourt (its support is being dropped) and would suggest moving to yay: https://github.com/Jguer/yay

I can make the changes on the script, accordingly.

EDIT: I cant open the script, gives 404. Also, it is missing gazebo for SITL, should I update it?

hamishwillee:
There is partial support - we have a Ubuntu script too, but it isn't "official" yet.

Can you please raise a defect against the containers repo/script with the 404 and the changes you think should be made? I'll tag it for the right people - @ dagar!

PR #841, opened 2019-07-19
Add diagram showing main/io functions

### LINKED ISSUE #839: No documentation on how px4 main processor and IO processor works
An outline like this image, for example, seems to be an important information to have for the split up of tasks. 
image

This doesn't seem to be in the docs

hamishwillee:
Fixes #839

@bkueng @echoGee This look OK?

Note:
1. I added UART to the diagram
2. Rather than list what does and does not have the board, I noted the typical naming convention. While not fixed in stone, I actually think that is a little more future proof.

Questions:
1. @echoGee - do you have the source for that image? Could I have that so if things modify in future we can update
2. IF safety switch and LED go through the IO does this mean that these are not present if IO is removed?


en/hardware/reference_design.md
```diff
@@ -17,3 +17,16 @@ FMU generations 1-3 were designed as open hardware, while FMU generations 4 and
...
 * FMUv6: work in progress, final name TBD, variant 6s \(STM32H7, 400 MHz, 2 MB RAM,  2 MB flash\) and variant 6i \(i.MX RT1050, 600 MHz, 512 KB RAM, external flash\)
+
+
+## Main/IO Function Breakdown
+
+The diagram below shows the division of bus and functional responsibilities between the FMU and I/O boards in a Pixhawk-series flight controller. 
+
+![PX4 Main/IO Function Breakdown main and io functions](../../assets/diagrams/px4_main_io_functions.png)
+
+Note that the I/0 board is optional; some Pixhawk-series controllers are built without the I/O board in order to save space.
```
bkueng: It's not only a question of space, but also:
- complexity
- cost
- use-case of the board

en/hardware/reference_design.md
```diff
@@ -17,3 +17,16 @@ FMU generations 1-3 were designed as open hardware, while FMU generations 4 and
...
+
+> **Note** Manufacturers often create flight controller variants with and without the I/O board. 
+  The version without an I/O bard is usually a size-reduced version (e.g. for use in racing vehicles) that is named as a "diminutive" of the version that has both boards.
```
bkueng: I wouldn't write racing vehicles: for multicopters in general the IO is not useful. The same holds for fully autonomous vehicles (w/o safety pilot).

bkueng (commented):
Thanks @hamishwillee 

> @echoGee - do you have the source for that image? Could I have that so if things modify in future we can update

And to clarify whether we can actually use the image.

> IF safety switch and LED go through the IO does this mean that these are not present if IO is removed?

No, the can also be attached to the FMU.

echoGee:
> Fixes #839
> 
> @bkueng @echoGee This look OK?
> 
> Note:
> 
> 1. I added UART to the diagram
> 2. Rather than list what does and does not have the board, I noted the typical naming convention. While not fixed in stone, I actually think that is a little more future proof.
> 
> Questions:
> 
> 1. @echoGee - do you have the source for that image? Could I have that so if things modify in future we can update

Unsure of this: I got it from https://discuss.px4.io/t/system-architecture-px4io-or-px4fmu/3794 and http://www.hex.aero/wp-content/uploads/2016/07/DRS_Pixhawk-2-17th-march-2016.pdf links
> 2. IF safety switch and LED go through the IO does this mean that these are not present if IO is removed?

hamishwillee:
Thanks very much. 
- I have regenerated the diagram and moved the buzzer/LED to be either/or. Also removed the FMU/IO part number as that could change. 
- Updated text in line with your feedback.

@bkueng ready for re-review

bkueng (approved):
Thanks @hamishwillee, looks nice.

PR #844, opened 2019-07-23
Add new frame - fix mixer note rendering

### LINKED ISSUE #842: document render problem in https://dev.px4.io/
in https://dev.px4.io/v1.9.0/en/airframes/adding_a_new_frame.html

the note render result is like this
```
In short, the output of this mixer would be SERVO = ( (roll input -0.6 + 0) + (pitch input 0.65 + 0) ) * 1 + 0
```

which is confusing

---

the github https://github.com/PX4/Devguide/blob/master/en/airframes/adding_a_new_frame.md

markdown render result is 
```
In short, the output of this mixer would be SERVO = ( (roll input * -0.6 + 0) + (pitch input * 0.65 + 0) ) * 1 + 0
```

correct

hamishwillee:
Fixes #842

- [x] Add to v1.9 docs when this is merged.

@bkueng This was rendering badly because the `*` were being interpretted as bold markup. First change was to escape all of those.
image

What I also did though is change the first section from `(roll input \* -0.6 + 0) ` to `(roll input \* -0.6 + 0)  \* 1` (ie add `\* 1` at the end) to reflect the scaling - this matches the second part of the markup.

At the very end there is a `+0` (see below). Is this to reflect the "0" line, or something else?
```
> **Note** In short, the output of this mixer would be SERVO = ( (roll input \* -0.6 + 0)  \* 1 + (pitch input \* 0.65 + 0) ) \* 1 + 0
```

### DIFF
--- a/en/airframes/adding_a_new_frame.md
+++ b/en/airframes/adding_a_new_frame.md
@@ -130,7 +130,7 @@ Where each number from left to right means:
 * S: Indicates the first input scaler: It takes input from control group #0 (Flight Control) and the first input (roll). It scales the roll control input * 0.6 and reverts the sign (-0.6 becomes -6000 in scaled units). It applies no offset (0) and outputs to the full range (-1..+1)
 * S: Indicates the second input scaler: It takes input from control group #0 (Flight Control) and the second input (pitch). It scales the pitch control input * 0.65. It applies no offset (0) and outputs to the full range (-1..+1)
 
-> **Note** In short, the output of this mixer would be SERVO = ( (roll input \* -0.6 + 0)  \* 1 + (pitch input \* 0.65 + 0) ) \* 1 + 0
+> **Note** In short, the output of this mixer would be SERVO = ( (roll input \* -0.6 + 0)  \* 1 + (pitch input \* 0.65 + 0)  \* 1 ) \* 1 + 0
 
 Behind the scenes, both scalers are added, which for a flying wing means the control surface takes maximum 60% deflection from roll and 65% deflection from pitch.
 
--- a/en/airframes/adding_a_new_frame.md
+++ b/en/airframes/adding_a_new_frame.md
@@ -118,8 +118,11 @@ S: 0 1   6500   6500      0 -10000  10000
 Where each number from left to right means:
 
 * M: Indicates two scalers for two control inputs. It indicates the number of control inputs the mixer will receive.
-* O: Indicates the output scaling (*1 in negative, *1 in positive), offset (zero here), and output range (-1..+1 here).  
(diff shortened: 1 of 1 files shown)


en/airframes/adding_a_new_frame.md
```diff
@@ -127,7 +127,7 @@ Where each number from left to right means:
 * S: Indicates the first input scaler: It takes input from control group #0 (Flight Control) and the first input (roll). It scales the roll control input * 0.6 and reverts the sign (-0.6 becomes -6000 in scaled units). It applies no offset (0) and outputs to the full range (-1..+1)
 * S: Indicates the second input scaler: It takes input from control group #0 (Flight Control) and the second input (pitch). It scales the pitch control input * 0.65. It applies no offset (0) and outputs to the full range (-1..+1)
 
-> **Note** In short, the output of this mixer would be SERVO = ( (roll input * -0.6 + 0) + (pitch input * 0.65 + 0) ) * 1 + 0
+> **Note** In short, the output of this mixer would be SERVO = ( (roll input \* -0.6 + 0)  \* 1 + (pitch input \* 0.65 + 0) ) \* 1 + 0
```
bkueng: If you add the scaling for roll you need to add it for pitch as well:
```suggestion
> **Note** In short, the output of this mixer would be SERVO = ( (roll input \* -0.6 + 0)  \* 1 + (pitch input \* 0.65 + 0)  \* 1 ) \* 1 + 0
```

bkueng (commented):
Looks like we also need to fix line 121:  https://github.com/PX4/Devguide/pull/844/files#diff-5872d7aa311843218a3260485c0d94b4R121

> At the very end there is a +0 (see below). Is this to reflect the "0" line, or something else?

It's the zero offset from the O: line.

en/airframes/adding_a_new_frame.md
hamishwillee: > If you add the scaling for roll you need to add it for pitch as well:

@bkueng My point was that pitch already had the scaling but roll did not - so I added it to roll. So doesn't this suggestion add the `x1` scaling for pitch a second time?

bkueng: No, the existing `\* 1` scaling applies to both roll and pitch (there is another bracket). It refers to the output scaling.
The added one refers to the second (output) scaling of the input scaler.

hamishwillee: Thanks @bkueng . I accepted this. It did take me a while to work out what you are talking about. Might have been easier to parse if some of the scalers here were not 1 - ie so that we could easily visually map between roll, pitch input scalers, and  0: output scaler.

PR #845, opened 2019-07-29
Update mavlink.md

brodieclark:
Omission of static return type keyword in get_id_static() causes compiler error.


hamishwillee:
Thanks @brodieclark ! Merged.

PR #853, opened 2019-08-14
mixing: explain rc scaling if FMU/IO stop communicating

hamishwillee:
As discussed in slack: https://px4.slack.com/archives/C581ST17V/p1565163436054700

### DIFF
--- a/en/concept/mixing.md
+++ b/en/concept/mixing.md
@@ -70,7 +70,7 @@ For a simple plane control 0 (roll) is connected straight to output 0 (aileron).
 * 7: RC aux3
 
 > **Note** This group is only used to define mapping of RC inputs to specific outputs during *normal operation* (see [quad_x.main.mix](https://github.com/PX4/Firmware/blob/master/ROMFS/px4fmu_common/mixers/quad_x.main.mix#L7) for an example of AUX2 being scaled in a mixer).
-  In the event of manual IO failsafe override (if the PX4FMU stops communicating with the PX4IO board) the mapping/mixing defined by control group 0 inputs for roll, pitch, yaw and throttle are used instead (other mappings are ignored).
+  In the event of manual IO failsafe override (if the PX4FMU stops communicating with the PX4IO board) only the mapping/mixing defined by control group 0 inputs for roll, pitch, yaw and throttle are used (other mappings are ignored).
 
 
 ### Control Group #6 (First Payload)
--- a/en/concept/mixing.md
+++ b/en/concept/mixing.md
@@ -70,7 +70,7 @@ For a simple plane control 0 (roll) is connected straight to output 0 (aileron).
 * 7: RC aux3
 
 > **Note** This group is only used to define mapping of RC inputs to specific outputs during *normal operation* (see [quad_x.main.mix](https://github.com/PX4/Firmware/blob/master/ROMFS/px4fmu_common/mixers/quad_x.main.mix#L7) for an example of AUX2 being scaled in a mixer).
-  In the event of manual IO failsafe override (if the PX4FMU stops communicating with the PX4IO board) inputs are passed to matching outputs by item number, unscaled.
+  In the event of manual IO failsafe override (if the PX4FMU stops communicating with the PX4IO board) the mapping/mixing defined by control group 0 inputs for roll, pitch, yaw and throttle are used instead (other mappings are ignored).
 
 
 ### Control Group #6 (First Payload)
--- a/en/concept/mixing.md
+++ b/en/concept/mixing.md
@@ -69,6 +69,10 @@ For a simple plane control 0 (roll) is connected straight to output 0 (aileron).
 * 6: RC aux2
 * 7: RC aux3
 
+> **Note** This group is only used to define mapping of RC inputs to specific outputs during *normal operation* (see [quad_x.main.mix](https://github.com/PX4/Firmware/blob/master/ROMFS/px4fmu_common/mixers/quad_x.main.mix#L7) for an example of AUX2 being scaled in a mixer).
+  In the event of manual IO failsafe override (if the PX4FMU stops communicating with the PX4IO board) inputs are passed to matching outputs by item number, unscaled.
+
+
 ### Control Group #6 (First Payload)
 
 * 0: function 0 (default: parachute)


en/concept/mixing.md
```diff
@@ -69,6 +69,10 @@ For a simple plane control 0 (roll) is connected straight to output 0 (aileron).
 * 6: RC aux2
 * 7: RC aux3
 
+> **Note** This group is only used to define mapping of RC inputs to specific outputs during *normal operation* (see [quad_x.main.mix](https://github.com/PX4/Firmware/blob/master/ROMFS/px4fmu_common/mixers/quad_x.main.mix#L7) for an example of AUX2 being scaled in a mixer).
+  In the event of manual IO failsafe override (if the PX4FMU stops communicating with the PX4IO board) inputs are passed to matching outputs by item number, unscaled.
```
bkueng: It's not clear to me what you mean with matching outputs and unscaled.

hamishwillee: @bkueng This was the guess I put into slack - you didn't comment so I assumed it was correct. Lets fix.

So this is how I was guessing it worked. Looking at it now I'm pretty sure it is all wrong:
-  IO board stops communicating with FMU
   - I am assuming the FMU controls everything - so IO is just dead/doing nothing? Or does it also have some sort of timeout/failsafe? 
  - Anyway, assuming it is dead, the MAIN ports output nothing. Seems unlikely because presumably motors/control surfaces are connected to MAIN. 
- FMU still working. manual IO failsafe happens. 
- Outputs from FMU (ie AUX) will get the RC input values. These will be the exact PWM values of the input from RC (ie no scaling). If the RC channels are 1-x, these will map to AUX 1 - 8. What happens to any additional channels and channels that were going to MAIN?

bkueng: > I am assuming the FMU controls everything - so IO is just dead/doing nothing? Or does it also have some sort of timeout/failsafe?

That is the whole point of the IO: if the communcation to the FMU times out, the IO will take over completely: it handles RC input, does the mixing and drives the outputs. It does not have sensors, so it cannot stabilize, you're flying fully manual (fixed-wing only).

> Outputs from FMU (ie AUX) will get the RC input values. These will be the exact PWM values of the input from RC (ie no scaling). If the RC channels are 1-x, these will map to AUX 1 - 8. What happens to any additional channels and channels that were going to MAIN?

Do you mean in a failsafe case? If the FMU crashes, you don't have any output on these pins.

hamishwillee: @bkueng You learn something every day. I thought the FMU was expected to live and the IO die. But what you're saying is that the IO lives. That explains why it has the MAIN ports too. - so that the main controls are on the bit that lives after FMU crashes.

So basically with that assumption, in the failsafe case, can you confirm how the mapping or RC input to outputs happens. My assumption above is that RC channels 1-8 map to MAIN1-8 (on I/O) and that any other channels are just lost.

Correct?

bkueng: :slightly_smiling_face: 

> My assumption above is that RC channels 1-8 map to MAIN1-8 (on I/O) and that any other channels are just lost.

It's defined by the mixer - i.e. your roll stick stays roll, goes through the mixing and is added to the appropriate outputs. As if you were flying in Manual. And I think you only have roll, pitch, yaw and throttle.

hamishwillee: OK, so now I'm confused again. HOw then do the failsafe and non failsafe modes differ, if both are going through the mixer. Is there a different control group for defining the mapping in the failsafe case?

bkueng: The difference is that in failsafe mode the IO has only limited information available, and therefore you are limited to Manual. Remember, all complex high-level operations (such as fly to a waypoint) are all broken down to simple mixer commands (e.g. change in roll), which is then mapped to the outputs through the mixer.
The IO is limited to doing the very last steps of that pipeline.

> Is there a different control group for defining the mapping in the failsafe case?

No

hamishwillee: @bkueng Arrrggg. Still not clear to me. Perhaps because I don't now how "same as flying in manual" works in terms of the mixer ...

You could mean that the same control group 3 mixing is used in normal and fail case. Just during fail you only use these inputs:  RC roll, RC pitch, RC yaw, RC throttle and DROP the others - RC mode switch, RC aux1, RC aux2, RC aux3

Or you could mean that in the fail case you don't use group 3 at all (as docs say), but instead use mixer definition for [control group 0](https://dev.px4.io/master/en/concept/mixing.html#control-group-0-flight-control) or control group 1 [albeit ignoring everything except the roll, pitch, yaw, throttle defined in those control groups]

Which is true? If neither, perhaps you could propose some text?

bkueng: > Or you could mean that in the fail case you don't use group 3 at all (as docs say), but instead use mixer definition for control group 0 or control group 1 [albeit ignoring everything except the roll, pitch, yaw, throttle defined in those control groups]

This one (although the mixer could already be configured to use group 3, but I think for failsafe only group 0 is available).
And yes it is a confusing topics.

hamishwillee:
Thanks @bkueng , I think I have this now - please do (hopefully) final check. One day i hope to chat through the mixing things again in person and make sure I have it down solid.

en/concept/mixing.md
```diff
@@ -69,6 +69,10 @@ For a simple plane control 0 (roll) is connected straight to output 0 (aileron).
 * 6: RC aux2
 * 7: RC aux3
 
+> **Note** This group is only used to define mapping of RC inputs to specific outputs during *normal operation* (see [quad_x.main.mix](https://github.com/PX4/Firmware/blob/master/ROMFS/px4fmu_common/mixers/quad_x.main.mix#L7) for an example of AUX2 being scaled in a mixer).
+  In the event of manual IO failsafe override (if the PX4FMU stops communicating with the PX4IO board) the mapping/mixing defined by control group 0 inputs for roll, pitch, yaw and throttle are used instead (other mappings are ignored).
```
bkueng: I'd say it like this:
```suggestion
  In the event of manual IO failsafe override (if the PX4FMU stops communicating with the PX4IO board) only the mapping/mixing defined by control group 0 inputs for roll, pitch, yaw and throttle are used (other mappings are ignored).
```

hamishwillee: Thanks. Works for me. All other inputs no longer enabled. Merging now!

PR #856, opened 2019-08-22
mixing: expand on mixing and output group

hamishwillee:
This better explains how output groups are mapped to physical bus, and hence how the magic of "MAIN" and "AUX" going to different physical but on different Pixhawk boards works. IT follows on from discussion here: https://github.com/PX4/Devguide/pull/852#issuecomment-522808899

### DIFF
--- a/en/concept/mixing.md
+++ b/en/concept/mixing.md
@@ -125,7 +125,9 @@ The mixer file does not explicitly define the actual *output group* (physical bu
 Instead, the purpose of the mixer (e.g. to control MAIN or AUX outputs) is inferred from the mixer [filename](#mixer_file_names), and mapped to the appropriate physical bus in the system [startup scripts](../concept/system_startup.md) (and in particular in [rc.interface](https://github.com/PX4/Firmware/blob/master/ROMFS/px4fmu_common/init.d/rc.interface)).
 
 > **Note** This approach is needed because the physical bus used for MAIN outputs is not always the same; it depends on whether or not the flight controller has an IO Board (see [PX4 Reference Flight Controller Design > Main/IO Function Breakdown](../hardware/reference_design.md#mainio-function-breakdown)) or uses UAVCAN for motor control.
-  The startup scripts load the mixer into the appropriate driver device for the board: `/dev/pwm_output0` (IO driver) or `/dev/pwm_output1` (FMU driver), `/dev/uavcan/esc` (uavcan driver).
+  The startup scripts load the mixer files into the appropriate device driver for the board, using the abstraction of a "device".
+  The main mixer is loaded into device `/dev/uavcan/esc` (uavcan) if UAVCAN is enabled, and otherwise `/dev/pwm_output0` (this device is mapped to the IO driver on controllers with an I/O board, and the FMU driver on boards that don't).
+  The aux mixer file is loaded into device `/dev/pwm_output1`, which maps to the FMU driver on Pixhawk controllers that have an I/O board.
 
 Since there are multiple control groups (like flight controls, payload, etc.) and multiple output groups (busses), one control group can send commands to multiple output groups.
 
@@ -135,6 +137,10 @@ graph TD;
   actuator_group_0-->output_group_6
   actuator_group_1-->output_group_0
 {% endmermaid %}
+
+> **Note** In practice, the startup scripts only load mixers into a single device (output group).
+  This is a configuration rather than technical limitation; you could load the main mixer into multiple drivers and have, for example, the same signal on both UAVCAN and the main pins.
+
 
 ## PX4 Mixer Definitions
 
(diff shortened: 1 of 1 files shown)


en/concept/mixing.md
```diff
@@ -117,9 +117,17 @@ These groups are NOT mixer inputs, but serve as meta-channels to feed fixed wing
...
+
+The mixer file does not explicitly define the actual *output group* (physical bus) where the outputs are applied.
+Instead, the purpose of the mixer (e.g. to control MAIN or AUX outputs) is inferred from the mixer [filename](#mixer_file_names), and mapped to the appropriate physical bus in the system [startup scripts](../concept/system_startup.md) (and in particular in [rc.interface](https://github.com/PX4/Firmware/blob/master/ROMFS/px4fmu_common/init.d/rc.interface)).
+
+> **Note** This approach is needed because the physical bus used for MAIN outputs is not always the same; it depends on whether or not the flight controller has an IO Board (see [PX4 Reference Flight Controller Design > Main/IO Function Breakdown](../hardware/reference_design.md#mainio-function-breakdown)) or uses UAVCAN for motor control.
+  The startup scripts load the mixer into the appropirate driver device for the board: `/dev/pwm_output0` (IO driver) or `/dev/pwm_output1` (FMU driver), `/dev/uavcan/esc` (uavcan driver).
+
+Since there are multiple control groups (like flight controls, payload, etc.) and multiple output groups (busses), one control group can send commands to multiple output groups.
```
hamishwillee, line 132: @bkueng So my inference in this paragraph and the preceding one, is that a mixer can be loaded into multiple drivers.

YOu can choose to use UAVCAN or not for motor control. My assumption is that the scripts load the mixer into both drivers. If you don't use UAVCAN the mixer code will still happily be trying to mix, but without any ESCs attached there will be no outputs for it to control - is that correct?

Also, UAVCAN have way more than 8 outputs but a mixer only defines 8, and from what I can see in the startup script, only MAIN mixer gets loaded into UAVCAN driver.  So assuming a QUAD with 4 motors gets loaded, does this mean that you can only use UAVCAN for 4 more outputs? Does my question make sense?

en/concept/mixing.md
```diff
@@ -117,9 +117,17 @@ These groups are NOT mixer inputs, but serve as meta-channels to feed fixed wing
 * 6: reserved / aux2
 * 7: reserved / aux3
 
-## Mapping
+## Output Groups/Mapping
 
-Since there are multiple control groups (like flight controls, payload, etc.) and multiple output groups (first 8 PWM outputs, UAVCAN, etc.), one control group can send command to multiple output groups.
+An output group is one physical bus (e.g. FMU PWM outputs, IO PWM outputs, UAVCAN etc.) that has 8 normalized (-1..+1) command ports that can be mapped and scaled through the mixer.
```
bkueng: There can be more or less than 8:
```suggestion
An output group is one physical bus (e.g. FMU PWM outputs, IO PWM outputs, UAVCAN etc.) that has N (usually 8) normalized (-1..+1) command ports that can be mapped and scaled through the mixer.
```

en/concept/mixing.md
```diff
@@ -117,9 +117,17 @@ These groups are NOT mixer inputs, but serve as meta-channels to feed fixed wing
...
+
+The mixer file does not explicitly define the actual *output group* (physical bus) where the outputs are applied.
+Instead, the purpose of the mixer (e.g. to control MAIN or AUX outputs) is inferred from the mixer [filename](#mixer_file_names), and mapped to the appropriate physical bus in the system [startup scripts](../concept/system_startup.md) (and in particular in [rc.interface](https://github.com/PX4/Firmware/blob/master/ROMFS/px4fmu_common/init.d/rc.interface)).
+
+> **Note** This approach is needed because the physical bus used for MAIN outputs is not always the same; it depends on whether or not the flight controller has an IO Board (see [PX4 Reference Flight Controller Design > Main/IO Function Breakdown](../hardware/reference_design.md#mainio-function-breakdown)) or uses UAVCAN for motor control.
+  The startup scripts load the mixer into the appropirate driver device for the board: `/dev/pwm_output0` (IO driver) or `/dev/pwm_output1` (FMU driver), `/dev/uavcan/esc` (uavcan driver).
```
bkueng:
```suggestion
  The startup scripts load the mixer into the appropriate driver device for the board: `/dev/pwm_output0` (IO driver) or `/dev/pwm_output1` (FMU driver), `/dev/uavcan/esc` (uavcan driver).
```

bkueng: Might be worth to clarify that `/dev/pwm_output0` is the fmu driver if no IO is there.

en/concept/mixing.md
bkueng: > @bkueng So my inference in this paragraph and the preceding one, is that a mixer can be loaded into multiple drivers.

Yes. But this sentence also applies to the case where you define the same control group in aux and main mixer files (for example).

> If you don't use UAVCAN the mixer code will still happily be trying to mix, but without any ESCs attached there will be no outputs for it to control - is that correct?

If you have UAVCAN enabled, the main mixer will not be loaded to the IO driver, so you don't have the main pins. It's probably done for simplicity, but technically it would easily be possible. Then you would have the same signal on UAVCAN as on the main pins.

> Also, UAVCAN have way more than 8 outputs but a mixer only defines 8, and from what I can see in the startup script, only MAIN mixer gets loaded into UAVCAN driver

I think a mixer file can go up to 16 outputs, not quite sure. But essentially yes, either 8 or 16 is the limit.

bkueng (commented):
Thanks @hamishwillee. Another mystery solved.

en/concept/mixing.md
hamishwillee: Thanks. Would be good to know which. But not sure how to capture this here, so lets see if the question ever comes up again "for real"

hamishwillee:
>  Another mystery solved.

A lot of effort, for possibly a small set of interested users. However it does satisfy me to know how this works :-)

I reworded slightly to capture the mapping of device abstraction through device driver. Can you give it a quick recheck ... feel free to merge if it seems OK.

bkueng (approved):

Issue #798, opened 2019-05-28
Arch Linux - which script should users run?


hamishwillee:
[Arch Linux > Script-based Installation](https://dev.px4.io/master/en/setup/dev_env_linux_arch.html#script-based-installation) suggests running https://github.com/PX4/containers/blob/master/docker/px4-dev/scripts/archlinux_install_script.sh - which no longer exists

@dagar I don't think there is a setup script for archlinux any more right - just https://github.com/PX4/containers/blob/master/docker/px4-dev/Dockerfile_base-archlinux

So is right move just to remove this whole section?

PR #859, opened 2019-08-26
setup: remove 404 link, reference to Dockerfile

julianoes:
This link was no longer valid, the Dockerfile is more helpful.

### DIFF
--- a/en/setup/dev_env_linux_arch.md
+++ b/en/setup/dev_env_linux_arch.md
@@ -19,21 +19,7 @@ Then log out and log in for changes to take effect.
 
 > **Note** This script installs the (unsupported) latest GCCE from the package manager. MicroRTPS is not built.
 
-Once ArchLinux is installed you can use the docker script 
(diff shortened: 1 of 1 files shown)


julianoes:
Also see: https://github.com/PX4/containers/issues/195 and #798.

hamishwillee (approved):

hamishwillee:
Thanks very much @julianoes  - merging.

Should we remove all the "manual instructions" too? (except the group permissions, which you can't infer from docker)

julianoes:
> Should we remove all the "manual instructions" too? 

I want to follow up with Arch Linux, I'll do it then.

hamishwillee:
Thanks. Sounds excellent. 

What I'd quite like is install scripts that are self documenting, and run in CI - ie run to setup dev environment AND that builds are done for those environments. Then we could reduce the manual setup instructions to just what can't be automated.

julianoes:
Yes, exactly!

Issue #861, opened 2019-08-28
Ubuntu install script breaks apt if lsb_release is missing


julianoes:
I ended up with apt not working anymore because the Ubuntu install script added a  malformed entry in `/etc/apt/sources.list.d/gazebo-stable.list` becuase `lsb_release` was not available.

This is on a clean docker `ubuntu:xenial`.

```
sh: 1: lsb_release: not found
--2019-08-28 12:37:41--  http://packages.osrfoundation.org/gazebo.key
Resolving packages.osrfoundation.org (packages.osrfoundation.org)... 54.176.9.38
Connecting to packages.osrfoundation.org (packages.osrfoundation.org)|54.176.9.38|:80... connected.
HTTP request sent, awaiting response... 200 OK
Length: 1772 (1.7K) [application/pgp-keys]
Saving to: 'STDOUT'

-                                           100%[========================================================================================>]   1.73K  --.-KB/s    in 0s      

2019-08-28 12:37:42 (76.4 MB/s) - written to stdout [1772/1772]

OK
E: Malformed entry 1 in list file /etc/apt/sources.list.d/gazebo-stable.list (Component)
E: The list of sources could not be read.
E: Malformed entry 1 in list file /etc/apt/sources.list.d/gazebo-stable.list (Component)
E: The list of sources could not be read.
E: Malformed entry 1 in list file /etc/apt/sources.list.d/gazebo-stable.list (Component)
E: The list of sources could not be read.
E: Malformed entry 1 in list file /etc/apt/sources.list.d/gazebo-stable.list (Component)
E: The list of sources could not be read.
E: Malformed entry 1 in list file /etc/apt/sources.list.d/gazebo-stable.list (Component)
E: The list of sources could not be read.
E: Malformed entry 1 in list file /etc/apt/sources.list.d/gazebo-stable.list (Component)
E: The list of sources could not be read.
E: Malformed entry 1 in list file /etc/apt/sources.list.d/gazebo-stable.list (Component)
E: The list of sources could not be read.
E: Malformed entry 1 in list file /etc/apt/sources.list.d/gazebo-stable.list (Component)
E: The list of sources could not be read.
E: Malformed entry 1 in list file /etc/apt/sources.list.d/gazebo-stable.list (Component)
E: The list of sources could not be read.
sudo: add-apt-repository: command not found
```

julianoes:
Will be fixed in the `ubuntu.sh` script in firmware.

PR #863, opened 2019-08-28
Fix two build script issues on minimal Ubuntu docker images

### LINKED ISSUE #862: Ubuntu install script assumes add-apt-repository
On minimal docker Ubuntu images I got the error:
```
add-apt-repository: command not found
```

It looks like this command should be installed by the `software-properties-common` package.

julianoes:
Fixes #862 and #861.

### DIFF
--- a/build_scripts/ubuntu_sim_nuttx.sh
+++ b/build_scripts/ubuntu_sim_nuttx.sh
@@ -22,7 +22,7 @@ sudo apt-get install python-serial openocd \
     libftdi-dev libtool zlib1g-dev -y
 
 # Clean up old GCC
-sudo apt-get remove gcc-arm-none-eabi gdb-arm-none-eabi binutils-arm-none-eabi gcc-arm-embedded -y
+sudo apt-get remove gcc-arm-none-eabi gdb-arm-none-eabi binutils-arm-none-eabi gcc-arm-embedded software-properties-common -y
 sudo add-apt-repository --remove ppa:team-gcc-arm-embedded/ppa -y
 
 
--- a/build_scripts/windows_bash_nuttx.sh
+++ b/build_scripts/windows_bash_nuttx.sh
@@ -35,9 +35,10 @@ fi
 
 # Common Dependencies
 echo "Installing common dependencies"
-sudo add-apt-repository ppa:george-edison55/cmake-3.x -y
 sudo apt-get update
-sudo apt-get install python-argparse git git-core wget zip python-empy python-toml python-numpy cmake build-essential genromfs vim-common -y
+sudo apt-get install python-argparse git git-core wget zip python-empy python-toml python-numpy build-essential genromfs vim-common software-properties-common -y
+sudo add-apt-repository ppa:george-edison55/cmake-3.x -y
+sudo apt-get install cmake -y
 # required python packages
 sudo apt-get install python-dev -y
 sudo apt-get install python-pip -y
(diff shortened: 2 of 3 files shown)


hamishwillee (approved):
I assume you did basic testing on these? Did you go all the way through to building?

julianoes:
I could not completely test it because the scripts download other scripts from master, so it's tricky.

hamishwillee:
That's exactly why I asked - same problem. Feel free to merge then test. If not, let me know, and I will do that on monday.

build_scripts/ubuntu_sim_nuttx.sh
```diff
@@ -22,7 +22,7 @@ sudo apt-get install python-serial openocd \
     libftdi-dev libtool zlib1g-dev -y
 
 # Clean up old GCC
-sudo apt-get remove gcc-arm-none-eabi gdb-arm-none-eabi binutils-arm-none-eabi gcc-arm-embedded -y
+sudo apt-get remove gcc-arm-none-eabi gdb-arm-none-eabi binutils-arm-none-eabi gcc-arm-embedded software-properties-common -y
```
hamishwillee, line 25: How come this script need this but ubuntu.py does not?

build_scripts/windows_bash_nuttx.sh
```diff
@@ -35,9 +35,10 @@ fi
 
 # Common Dependencies
 echo "Installing common dependencies"
-sudo add-apt-repository ppa:george-edison55/cmake-3.x -y
 sudo apt-get update
-sudo apt-get install python-argparse git git-core wget zip python-empy python-toml python-numpy cmake build-essential genromfs vim-common -y
+sudo apt-get install python-argparse git git-core wget zip python-empy python-toml python-numpy build-essential genromfs vim-common software-properties-common -y
```
hamishwillee, line 39: So just FMI, why can't we use the cmake from the normal repo? This is why we need software-properties-common here right?

hamishwillee: Actually I see that it is because we need a different cmake. But again - "why"? The ubuntu.sh script seems to install cmake fine.

I'd quite like to dump these scripts and move to ubuntu.sh, which is why I'm asking if perhaps we can avoid this change.

build_scripts/ubuntu_sim_nuttx.sh
julianoes: It's required for `add-apt-repository` which is not used in ubuntu.sh.

hamishwillee: Yeah, but they are both setting up the same environment right - so the question is really "why does one need this and not the other". 

I'm hoping to dump this script as soon as we're happy ubuntu.sh does the job adequately.

build_scripts/windows_bash_nuttx.sh
julianoes: Agreed. We need to check what cmake version we require.

According to https://github.com/PX4/Firmware/blob/60e5e0821a0ee64c42cd77e85e0c0c39526dca88/CMakeLists.txt#L102 we need 3.2.
Ubuntu 16.04 comes with 3.5.1, so that should be fine.

hamishwillee: So I'm a little lost, but does that mean most of the changes except LSB-core addition should be reverted?

julianoes: It means that we should probably remove the ppa for cmake because it's not required anymore, and that means we also would not need the `software-properties-common` anymore.

hamishwillee: Great. I'll leave this fix to you when you do the thingy for https://github.com/PX4/Firmware/pull/12827#issuecomment-527953817

I'll add that to ubuntu.py

PR #865, opened 2019-09-16
Fix up simulator port label

hamishwillee:
This fixes up label of simulation port, as highlighted by @Stifael 

@Stifael Look OK to you? Note, ignore the svg file changes - the PNG is rendered from it so you only really need to look at the final doc..
Question inline.


en/simulation/README.md
```diff
@@ -57,13 +57,14 @@ Message | Direction | Description
...
-By default, PX4 uses commonly established UDP ports for MAVLink communication with ground control stations (e.g. *QGroundControl*), Offboard APIs (e.g. MAVSDK, MAVROS) and simulator APIs (e.g. Gazebo). These ports are:
+By default, PX4 uses commonly established UDP ports for MAVLink communication with ground control stations (e.g. *QGroundControl*), Offboard APIs (e.g. MAVSDK, MAVROS) and simulator APIs (e.g. Gazebo).
+These ports are:
 
 * UDP Port **14540** is used for communication with offboard APIs.
   Offboard APIs are expected to listen for connections on this port.
 * UDP Port **14550** is used for communication with ground control stations.
   GCS are expected to listen for connections on this port. *QGroundControl* listens to this port by default.
-* TCP Port **4560** is used for communication with simulators.
+* The simulator's local TCP Port **4560** is used for communication with PX4.
```
hamishwillee, line 67: @Stifael Is the line below correct? 

I understand how the UDP stuff works - PX4 broadcasts to local UDP ports, and simulators listening on that port and connect when they detect . Not so sure on TCP - I think the simulator defines the connection by broadcasting to this port and PX4 is expected to listen for the simulator ... but it could be the other way (simulator listens on its port and PX4 broadcasts).

Stifael: The simulator is expected to have a local TCP port of 4560 and PX4 will connect to it. This means that the simulator is just the server that you can reach via 4560 port (similar to http with port 80)

Stifael (approved):

Stifael:
I think that looks ok given that it is not the px4 devguide's responsibility to explain tcp

en/simulation/README.md
hamishwillee: Thanks. Stuff I should know, but I don't want to screw up.

hamishwillee:
Thanks! I agree. Merged.

jeprince:
How does this work when running multiple SITL PX4s and loading them in gazebo? By default, iris.sdf file has the tcp port set to 4560. Is that supposed to be the random port the SITL PX4 uses to connect to the simulator's port on 4560?

PR #867, opened 2019-09-17
Devguide update to use ubuntu.sh

hamishwillee:
At high level this replaces the existing devguide scripts with ubuntu.sh. Note:
- New structure removes manual steps except where not covered in scripts from ubuntu docs.
- Old scripts replaced in doc, but not removed from tree (yet). Exception is ROS melodic script, which is not covered by ubuntu.sh. I did some tidy up of the old scripts to in order to remove installing firmware and also remove qtcreator.

Rendered version here: https://hamishwillee.gitbooks.io/havdevguide/content/v/pr_ubuntu_sh/en/setup/dev_env_linux_ubuntu.html

Also hoping to close some other things around the place - e.g. https://github.com/PX4/Firmware/issues/11024

### DIFF
--- a/en/setup/dev_env_linux.md
+++ b/en/setup/dev_env_linux.md
@@ -1,123 +1,18 @@
 # Development Environment on Linux
 
-Linux allows you to build for [all PX4 targets](../setup/dev_env.md#supported-targets) (NuttX based hardware, Qualcomm Snapdragon Flight hardware, Linux-based hardware, Simulation, ROS).
+Linux allows you to build for [all PX4 targets](../setup/dev_env.md#supported-targets) (NuttX based hardware, Qualcomm Snapdragon Flight hardware, Bebop, Linux-based hardware, Simulation, ROS).
 
-> **Tip** [Ubuntu Linux LTS](https://wiki.ubuntu.com/LTS) 16.04 is the tested/supported Linux distribution for most development.
-  Ubuntu 18.04 LTS with ROS Melodic is used for [ROS development](#ros). 
-  Instructions are also provided for [CentOS](../setup/dev_env_linux_centos.md) and [Arch Linux](../setup/dev_env_linux_arch.md).
+> **Tip** [Ubuntu Linux LTS](../setup/dev_env_linux_ubuntu.md) 16.04 is recommended.
+  Other environments can be made to work, but are less well tested and documented.
 
-The following instructions explain how to set up a development environment on Ubuntu LTS using convenience bash scripts. 
-Instructions for *manually installing* these and additional targets can be found in [Ubuntu/Debian Linux](../setup/dev_env_linux_ubuntu.md).
+The following instructions explain how to set up a development environment on various Linux platforms.
 
-
-## Development Toolchain
-
-The instructions below show how you can use our [convenience bash scripts](../setup/dev_env_linux_ubuntu.md#convenience-bash-scripts) to setup the developer toolchain on Ubuntu LTS. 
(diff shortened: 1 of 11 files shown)


en/setup/dev_env_linux.md
```diff
@@ -1,123 +1,18 @@
 # Development Environment on Linux
```
hamishwillee, line 1: FYI Top level script has just become index to other scripts rather than being script instructions that linked to the "manual" version in the lower level ubuntu page.

en/setup/dev_env_linux_ubuntu.md
```diff
@@ -1,304 +1,80 @@
 # Development Environment on Ubuntu LTS / Debian Linux
```
hamishwillee, line 1: FYI This now is the sole "ubuntu linux" setup page. It removes most of the explanation of how scripts work, which was a maintenance problem. The only instructions are for using scripts and things that aren't in the scripts.

en/setup/dev_env_linux_ubuntu.md
```diff
@@ -1,304 +1,80 @@
...
-> **Note** Packages with specified versions should be installed with the specified package version.
-
-```sh
-sudo apt-get install python-serial openocd \
-    flex bison libncurses5-dev autoconf texinfo \
-    libftdi-dev libtool zlib1g-dev -y
-```
-
 
+<!-- Do we need to add to our scripts or can we assume correct version installs over?
```
hamishwillee, line 53: @dagar original instructions suggested removing old version of arm-none-eabi compiler before installing new one. We don't do that in script. Can you confirm it is not needed?

hamishwillee:
@julianoes Any thoughts on this?

julianoes (commented):
Why don't you want to release the files yet? Can't we do a clean cut?

hamishwillee:
> Why don't you want to release the files yet? Can't we do a clean cut?

I don't know what you mean by the above. I want to merge this change, which uses the ubuntu.sh script. I think having just one script is a "good thing".

julianoes:
Sorry, I don't know why I wrote that this way. What I wanted to task why you don't want to delete the files already?

hamishwillee:
@julianoes 

> What I wanted to task why you don't want to delete the files already?

Because the last change in 1.9.0 instructions hard coded the path to those scripts in master. Either way, that shouldn't really affect whether this change  makes sense (and I plan to fix it separately).

This is a pretty big change because it removes the step by step manual instructions; if you're broadly happy then I'm happy to merge.

If you happen to have an opinion on this question https://github.com/PX4/Devguide/pull/867#discussion_r325033346 it would be useful (but would not block me merging)

julianoes:
Ok, I understand. I guess we would have to change this in v1.9:

> 1. Download the script in a bash shell:
> 
> wget https://raw.githubusercontent.com/PX4/Devguide/master/build_scripts/ubuntu_sim.sh

and change `master` to `v1.9.0` but agreed we can do that later.

en/setup/dev_env_linux_ubuntu.md
julianoes: Generally I'm against removing stuff on someones computer. I think it's one thing to install dependencies but I disagree with removing them.

That being said by adding the custom arm-none-eabi compiler and then adding the path **to the beginning** we prioritize our custom compiler anyway and it should work just fine even if you have it already installed.

hamishwillee: Makes sense! Thanks. Another good reason to remove our old script and instructions :-)

hamishwillee:
@julianoes Thank you. So is it good to merge otherwise?

julianoes (approved):
I think that's good to go in.

hamishwillee:
Thanks @julianoes . Merging. FYI @dagar - if ubuntu.sh is not in CI, now would be a good time :-)

hamishwillee:
@julianoes FYI, deleted the scripts I can. Everything still depends on most of them because ROS scripts depends on the sub scripts. We need to think if we can/should recreate that to build on the new script.

PR #874, opened 2019-10-07
Update offboard_control.md

fabrizioschiano:
The link to the Odroid XU4 was broken


TSC21 (changes requested):
I really doubt this link will be permanent. What abou https://magazine.odroid.com/odroid-xu4?

fabrizioschiano:
I also think that the link I proposed will not be permanent. However, it is the closest to the previous link. The one proposed by @TSC21 could be ok too.

TSC21:
> I also think that the link I proposed will not be permanent. However, it is the closest to the previous link. The one proposed by @TSC21 could be ok too.

Can you please update the commit with that link then? Thanks

TSC21:
Thanks!

TSC21 (approved):

PR #878, opened 2019-10-12
Arch setup: Update how to use new setup script

MaEtUgR:
Since I made a new setup script in the Firmware repository for Arch including the right version arm-gcc compiler and gazebo
https://github.com/PX4/Firmware/pull/13111
https://github.com/PX4/Firmware/pull/13160
and use it as a daily driver in a VM now I wanted to update instructions on how to reproduce.

FYI @hamishwillee @julianoes @jlecoeur


en/setup/dev_env_linux_arch.md
```diff
@@ -1,27 +1,48 @@
...
 ```
 
-Then log out and log in for changes to take effect.
+#### Option 2 only install Toolchain
+```sh
+wget https://raw.githubusercontent.com/PX4/Firmware/master/Tools/setup/arch.sh
+source arch.sh # optionally append --gazebo
+```
+
+## Manual Installation
```
jlecoeur: I think this section is not needed.

Actually it can be confusing. I read this document quickly and wondered why these steps were necessary after running the install script.

I expect Arch Linux users to be able to read the install script and run the commands manually if they prefer to. This is also more text to maintain. I suggest to move any useful comments into the install script and to remove this section.

en/setup/dev_env_linux_arch.md
```diff
@@ -1,27 +1,48 @@
...
 
+### Run the setup script
+The script by default installs all tools to build PX4 (without RTPS) for NuttX targets and run simulation with jMAVsim.
+
+Optionally you can specify the parameter `--gazebo` to also install gazebo simulation from the [AUR](https://aur.archlinux.org/packages/gazebo/). Note that gazebo gets compiled from source and therefore takes some time to install and requires entering the sudo password multiple times for dependencies.
+
+#### Option 1 Clone PX4, install Toolchain
 ```sh
-sudo usermod -a -G uucp $USER
+git clone --recurse-submodules -j8 https://github.com/PX4/Firmware.git
```
jlecoeur: Submodules are cloned on demand during compilation, aren't they? 
`--recursive-submodules` is useful but takes time and disk space for someone who just wants to install the toolchain.

jlecoeur (commented):
I made two comments, looks good otherwise.

en/setup/dev_env_linux_arch.md
MaEtUgR: I agree. I just didn't want to wrecklessly remove all your instructions without having checked if they contain anything additional e.g. I don't remember if I would need `genromfs` for a specific case or it's anyways already contained somewhere else. I can go though.

en/setup/dev_env_linux_arch.md
MaEtUgR: Are they reliably? I always clone all of them. 😇 TIL, I'll remove it 👍 
Btw if someone want to just install the toolchain he can `wget` the raw script but I see your point.

en/setup/dev_env_linux_arch.md
```diff
@@ -1,27 +1,48 @@
...
+# Development Environment on Arch Linux
 
-> **Note** These instructions allow you to build PX4 (without RTPS) for NuttX targets, using an unsupported version of GCCE from the package manager.
-The instructions have been tested on Antergos (an Arch Linux based distribution) as it is easier to set up than Arch Linux.
-We hope to provide fully tested instructions with the supported toolchain in the near future.
+> **Note** These instructions allow you to build PX4 (without RTPS) for NuttX targets and run simulation with jMAVsim or gazebo.
+The instructions have been tested on Manjaro (an Arch Linux based distribution) as it is much easier to set up than Arch Linux.
 
-## Permissions
+## Setup script
```
hamishwillee: @MaEtUgR Next time you do a docs push could you please push to the PX4 repo rather than a personal branch. That way I can apply technical editing in place.

en/setup/dev_env_linux_arch.md
```diff
@@ -1,27 +1,48 @@
...
 
-The user needs to be added to the group "uucp":
+Using the the provided setup script it's very convenient to set your Arch installation up for PX4 development. Here are the steps to get you started from scratch.
 
+### Run the setup script
+The script by default installs all tools to build PX4 (without RTPS) for NuttX targets and run simulation with jMAVsim.
+
+Optionally you can specify the parameter `--gazebo` to also install gazebo simulation from the [AUR](https://aur.archlinux.org/packages/gazebo/). Note that gazebo gets compiled from source and therefore takes some time to install and requires entering the sudo password multiple times for dependencies.
```
hamishwillee: Would be good next time we update the script to match the ubuntu.sh approach - that has `--no-nuttx` and `--no-sim-tools` arguments.

en/setup/dev_env_linux_arch.md
hamishwillee: Agree too. I  would like the script to be sufficient and remove this altogether. If we need to add more docs to the script to explain them then go for it.

en/setup/dev_env_linux_arch.md
dagar: Yes it works reliably. My typical workflow (often jumping between various old branches and back to master) is to `make distclean` and let the next build reinit the subset of submodules actually used by that particular config.

MaEtUgR: Nice! Then we could add gtest as a submodule as well. I was only doing the configure time hackery because I didn't want everyone to clone it even though they're not testing.

en/setup/dev_env_linux_arch.md
MaEtUgR: @hamishwillee I would love to, can you give me access? I have the same rights as any stranger here :smirk:

dagar: @MaEtUgR welcome to the docs team.

en/setup/dev_env_linux_arch.md
MaEtUgR: Yeah, it follows exactly that approach and has these same parameters, docs don't tell you, I have to add 😬 
Gazebo not being enabled by defaults with "sim-tools" and having an additional parameter is done only because I consider it a bit less stable until I've seen how this AUR way of installing things holds up on the longer run. I'm kind of new to it. I can change it as soon as I have more insight into reproducibility.

en/setup/dev_env_linux_arch.md
MaEtUgR: Thanks!

hamishwillee:
@MaEtUgR Let me know when you think ready to merge. There are style fixes I'll make as a post process.

en/setup/dev_env_linux_arch.md
MaEtUgR: I checked, it's really not needed. Even genromfs is not mandatory (anymore). I removed the section one by one while checking.

en/setup/dev_env_linux_arch.md
MaEtUgR: Fixed that.

jlecoeur (approved):

en/setup/dev_env_linux_arch.md
MaEtUgR: Added

MaEtUgR:
@hamishwillee I let you know I consider it ready.
Sorry @jlecoeur I had to push again because I forgot some stuff, can you check the diff? https://github.com/PX4/Devguide/compare/ebe3d977025751fedf0bec856a3cd13ee2064afa..a6598a6584e8a53257a0d32dcd4949c8a82479a9

jlecoeur:
looks good !

hamishwillee:
Thanks very much! I merged and then did updates: https://github.com/PX4/Devguide/pull/879

The only one of interest was to add download of requirements.txt too.

PR #881, opened 2019-10-28
docker: use docker without sudo

julianoes:
We should not net to use it with sudeo because we have the magic in docker_run.sh to make sure the privileges are all correct.


LorenzMeier (approved):

hamishwillee:
@julianoes Awesome, provided it actually works.  I tested this originally with `sudo` when I first subedited it. But not without.

PR #882, opened 2019-10-31
Ulog_file_format: topic name

kamilritz:
Not sure if this is correct. @bkueng can you comment on that?


bkueng (commented):
I'd rather keep it as is. The ULog file format does not make a distinction between topics and messages.
We can add a note about this though (comparing it to uORB).

PR #888, opened 2019-11-14
Update building_px4.md

unipheas:
Gave instruction to beginners on how to install git on their computer.

### DIFF
--- a/en/setup/building_px4.md
+++ b/en/setup/building_px4.md
@@ -26,6 +26,8 @@ The steps to fork and clone the project source code are:
    The easiest way to do this is to click the **Clone or download** button and then copy the URL:
 
    ![Github Clone or download button](../../assets/toolchain/github_clone_or_download.png)
+1. Install git on your computer. Instructions can be found [here](https://git-scm.com/downloads)
+   * On macOS you can use `Homebrew` by typing `brew install git` in your terminal
 1. Open a command prompt/terminal on your computer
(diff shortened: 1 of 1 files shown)


en/setup/building_px4.md
```diff
@@ -26,6 +26,8 @@ The steps to fork and clone the project source code are:
    The easiest way to do this is to click the **Clone or download** button and then copy the URL:
 
    ![Github Clone or download button](../../assets/toolchain/github_clone_or_download.png)
+1. Install git on your computer. Instructions can be found [here](https://git-scm.com/downloads)
```
hamishwillee: Many users  users should already have git on their computer from having installed the dev toolchain, so perhaps we should do this like (?):
```
1. Install *git* (if you haven't already done so as part of setting up the development environment):
  * On macOS use the terminal command: `brew install git`
  * On Ubuntu use the terminal command: `sudo apt install git`
  * For other platforms see the [git documentation](https://git-scm.com/downloads).
```

Reasonable?

unipheas: I like this. I was initially trying to make it very easy for beginners to get into, as clear as possible.

unipheas:
@hamishwillee I've updated based on your recommendations.

en/setup/building_px4.md
hamishwillee: Thanks. Merged. It's always a good goal to go for the a beginner base level. I draw the line at Git though :-)

PR #889, opened 2019-11-14
Update building_px4.md

unipheas:
Added a fix for macOS Catalina when trying to run jmavsim

### DIFF
--- a/en/setup/building_px4.md
+++ b/en/setup/building_px4.md
@@ -76,6 +76,8 @@ Navigate into the **Firmware** directory and start [jMAVSim](../simulation/jmavs
 make px4_sitl jmavsim
 ```
 
+> If you are on macOS Catalina then you might have a problem with cmake. Run `xcode-select --install` then `sudo ln -s /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/* /usr/local/include/` in your terminal and it should all work.
+
 This will bring up the PX4 console below:
 
 ![PX4 Console (jMAVSim)](../../assets/console_jmavsim.png)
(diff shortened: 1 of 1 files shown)


en/setup/building_px4.md
```diff
@@ -76,6 +76,8 @@ Navigate into the **Firmware** directory and start [jMAVSim](../simulation/jmavs
 make px4_sitl jmavsim
 ```
 
+> If you are on macOS Catalina then you might have a problem with cmake. Run `xcode-select --install` then `sudo ln -s /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/* /usr/local/include/` in your terminal and it should all work.
```
hamishwillee: @bkueng Can you please sanity check?

@unipheas Would the problem have obvious symptoms?

unipheas: I haven't had any problems. This was a solution proposed on StackOverflow that worked for me. I'm not sure if it would though. That said, I noticed that this may be better placed at the bottom of the document under Troubleshoot section.

hamishwillee: Ah, so you proactively did the fix - never actually saw a problem?
The thing is we want people to be able to find this by search too.

Yes, I think putting in troubleshooting section would be better. 

FYI only if this is done as a note or tip, then it is better to to like this `> **Note** Your text` than `Your text` - as the format of the former creates a nice rendering in gitbook, and does not break github.

unipheas: Sorry. With the most recent macOS update I did have a problem with CMAKE. I did have to use this `hack` to fix the issue so I could build the project.

Thanks for the tip. So, then we should add this to the troubleshooting section then?

hamishwillee:
> So, then we should add this to the troubleshooting section then?

Yes please!

en/setup/building_px4.md
bkueng: Looks reasonable, although I'm not using macOS.

hamishwillee:
Thank you very much!

PR #898, opened 2019-11-21
params: multi-instance and yaml

hamishwillee:
This follows on from discussion that multi-instance params are not documented: https://github.com/PX4/Firmware/pull/12551#issuecomment-555281171

@bkueng This is a starting point covering just multi-instance. However reading over this, the very existance of yaml parameter instances is not covered - currently everything is about documenting in the .c file. We need full docs.

My understanding is that the YAML format is
- an alternative to using a .c file.
- It must follow the format defined in https://github.com/PX4/Firmware/blob/master/validation/module_schema.yaml
- It allows extra stuff like multi-instance. Anything else?
- Any limitations? ie can you do anything with .c format you can't with this. Timmy mentioned something about problems with parameters that are defined in libraries "since some parameters are defined outside of a module (in the battery library)."

Do you want to write that/extend this? Then I subedit?

### DIFF
--- a/en/advanced/parameters_and_configurations.md
+++ b/en/advanced/parameters_and_configurations.md
@@ -274,3 +274,21 @@ The purpose of each line is given below (for more detail see [module_schema.yaml
  * @group <a title for parameters that form a group>
  */
 ```
+
+### Multi-Instance (Templated) Meta Data {#multi_instance_metadata}
+
+Templated parameter definitions are supported in [YAML parameter definitions](https://github.com/PX4/Firmware/blob/master/validation/module_schema.yaml) (templated parameter code is not supported).
+
+The YAML allows you to define instance numbers in parameter names, descriptions, etc. using `${i}`.
+For example, below will generate MY_PARAM_1_RATE, MY_PARAM_2_RATE etc.
+```
+MY_PARAM_${i}_RATE:
+            description:
+                short: Maximum rate for instance ${i}
+```
+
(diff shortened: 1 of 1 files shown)


AmeliaEScott:
Regarding "some parameters are defined outside of a module (in the battery library)": I found that I can just add the following in the CMakeLists.txt for the library and it works fine:

```
set_property(GLOBAL APPEND PROPERTY PX4_MODULE_CONFIG_FILES ${CMAKE_CURRENT_SOURCE_DIR}/module.yaml)
```

The file can also have another name, like `parameters.yaml`, since it is not in a module. (Side note: It might be worth it to add some functionality to our custom CMake functions to better support this, but this is perhaps not the right place to discuss it)

Another limitation with multi-instance parameters is that there are issues using them with the C++ API. Beat can explain that better than me.

bkueng:
The yaml config is meant as a full replacement for the .c files, and supports everything in terms of metadata.

> The file can also have another name, like parameters.yaml, since it is not in a module. (Side note: It might be worth it to add some functionality to our custom CMake functions to better support this, but this is perhaps not the right place to discuss it)

Exactly. That's the missing part: enabling libraries to add yaml config files.

hamishwillee:
@bkueng OK, I have updated to separate out .c and .yaml, and explain why both exist. Can you please review.

bkueng (approved):
Thanks

PR #905, opened 2019-11-27
uorb: add paragraph for message/field deprecation

bkueng:
Add some guidelines for uORB message changes so that external (log analysis) tools can detect changes without ambiguities.

@kamilritz fyi


hamishwillee (approved):
Looks good

PR #911, opened 2019-12-17
Update vscode.md

oneWayOut:
Change Linux 18.04 to Ubuntu 18.04


hamishwillee:
Thanks very much! Merged. We're having some build issues, so this might not appear live for a little while.

oneWayOut:
Great!
Just one more thing: I wonder why Ubuntu 16.04 allows to build but not debug targets with vscode?
 I have exactly the same problem.

hamishwillee:
> Just one more thing: I wonder why Ubuntu 16.04 allows to build but not debug targets with vscode?
I have exactly the same problem.

No idea. But could not get it to work. More important to get the documentation out that resolve that issue.

oneWayOut:
OK, Thank you!

PR #912, opened 2019-12-29
Update dev_mac ulimit

aokholm:
The maximum allowed number of open files on macOS needs to be increased even further to 2048. See thread https://discuss.px4.io/t/errors-about-too-many-open-files/9150


hamishwillee:
Thank you very much! Merged.

PR #913, opened 2019-12-30
Add instructions to simulate multi vehicles without ROS

Jaeyoung-Lim:
This describes the instructions on launching multi vehicle simulation without ROS from https://github.com/PX4/Firmware/pull/13806 https://github.com/PX4/Firmware/pull/13729 [Note, don't merge until these in]

@hamishwillee It would be nice if I can add videos examples as the following. How do you suggest that we can do this?
- [multiple iris gazebo sitl](https://youtu.be/Mskx_WxzeCk)
- [multiple plane gazebo sitl](https://youtu.be/aEzFKPMEfjc)

### DIFF
--- a/en/simulation/multi-vehicle-simulation.md
+++ b/en/simulation/multi-vehicle-simulation.md
@@ -6,12 +6,13 @@ A different approach is used for simulation with and without ROS.
 
 ## Multiple Vehicle with Gazebo (No ROS) {#no_ros}
 
-To launch multiple vehicles in Gazebo (without ROS) use the following commands in the terminal (from the root of the *Firmware* tree):
+To simulate (up to 10) iris or plane vehicles in Gazebo use the following commands in the terminal (from the root of the *Firmware* tree):
 ```
-Tools/gazebo_sitl_multiple_run.sh [-m <model>] -n <number_of_vehicles>
+Tools/gazebo_sitl_multiple_run.sh [-m <model>] [-n <number_of_vehicles>]
 ```
 
-The `<model>` flag specifies which model to spawn: `iris` (default), `plane`.
+- `<model>`: The vehicle type/model to spawn: `iris` (default), `plane`.
+- `number_of_vehicles`: The number of vehicles to spawn. Default is 3. Maximum is 10.
 
 
 ### Video: Multiple Multicopter (Iris)
--- a/en/simulation/multi-vehicle-simulation.md
+++ b/en/simulation/multi-vehicle-simulation.md
@@ -6,7 +6,7 @@ A different approach is used for simulation with and without ROS.
 
 ## Multiple Vehicle with Gazebo (No ROS) {#no_ros}
 
-To simulate (up to 10) iris or plane vehicles in Gazebo use the following commands in the terminal (from the root of the *Firmware* tree):
+To simulate multiple iris or plane vehicles in Gazebo use the following commands in the terminal (from the root of the *Firmware* tree):
 ```
 Tools/gazebo_sitl_multiple_run.sh [-m <model>] [-n <number_of_vehicles>]
 ```
@@ -14,7 +14,10 @@ Tools/gazebo_sitl_multiple_run.sh [-m <model>] [-n <number_of_vehicles>]
 - `<model>`: The vehicle type/model to spawn: `iris` (default), `plane`.
 - `number_of_vehicles`: The number of vehicles to spawn. Default is 3. Maximum is 10.
 
-Each vehicle instance is allocated a unique MAVLink system id (1, 2, 3, etc) and can be accessed from a unique UDP port (14540, 14541, 14542, etc.).
+Each vehicle instance is allocated a unique MAVLink system id (1, 2, 3, etc.) and can be accessed from a unique remote offboard UDP port (14540, 14541, 14542, etc.).
+
+> **Note** The 10-vehicle limitation occurs because the 11th vehicle would be allocated a remote offboard UDP port 14550, which is already used for QGC.
+  The `MAV_SYS_ID` and various UDP ports are allocated in the SITL rcS: [init.d-posix/rcS](https://github.com/PX4/Firmware/blob/master/ROMFS/px4fmu_common/init.d-posix/rcS#L108-L112)
 
 ### Video: Multiple Multicopter (Iris)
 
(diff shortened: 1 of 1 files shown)


hamishwillee:
Thanks @Jaeyoung-Lim . I updated these with video embedded and tidied a bit. If you would prefer we could just link to the videos.

Some questions:
1. Can I still use MAVSDK to control the vehicles as usual - ie by talking to the normal port? How do I address the individual vehicles to do so?
2. Is there any limit on the number of vehicles?
3. Is there a default number of vehicles?
4. Is there any special setup or is it just standard Gazebo environment?
5. What if I want to have a different background than just the boring default world - can I do that?
6. Can I run the command HEADLESS?

Jaeyoung-Lim:
@hamishwillee Thanks for the update! These are the best answers I can provide

1. This I believe is true, but I have never tried running it. Maybe @JonasVautherin Can help us out here.
2. The number of vehicles are in theory not limited, but the script shows wierd behaviors after 10 vehicles since the ports start to overlap. This is something I would need to look into and possibly fix, but currently the limit is 10
3. The default number of vehicles are 3
4. This is just a standard gazebo environment
5. For that you need to modify the `<model>.world` file as usual
6. There is currently no option for that. May be that is a good idea and should be added, but this would be a separate PR

hamishwillee:
@Jaeyoung-Lim Thanks very much!

>> Can I still use MAVSDK to control the vehicles as usual - ie by talking to the normal port? How do I address the individual vehicles to do so?

> This I believe is true, but I have never tried running it. Maybe @JonasVautherin Can help us out here.

I think it is probably more of an @julianoes question. Specifically, consider https://dev.px4.io/master/en/simulation/multi_vehicle_jmavsim.html states *Developer APIs such as MAVSDK connect on the UDP port 14540 (first instance), UDP port 14541 (second instance), and so on.*

So the assumption is that each instance needs its own UDP port for talking to developer APIs (is that necessary? they should differentiate by system id).

Jaeyoung-Lim:
@hamishwilee It also has different systemid.

But for udp ports I think it is necessary given how the simulation is done. Not sure how this relates to mavsdk. Maybe it is a good opportunity to get my hands on it :)

hamishwillee:
@Jaeyoung-Lim So currently each instance has a system id from 1 (e.g. 1, 2, 3,..) and uses a UDP port (14540, 14541, 14542, ...)?

Jaeyoung-Lim:
Yes

hamishwillee:
Thanks! I added a line to clarify this. I think it is good to go. Can you please check it again, and let me know when the corresponding Firmware changes have merged so this can go in?

JonasVautherin:
About MAVSDK, I believe the answer is [here in the SITL rcS](https://github.com/PX4/Firmware/blob/master/ROMFS/px4fmu_common/init.d-posix/rcS#L108-L112):

```
param set MAV_SYS_ID $((px4_instance+1))
simulator_tcp_port=$((4560+px4_instance))
udp_offboard_port_local=$((14580+px4_instance))
udp_offboard_port_remote=$((14540+px4_instance))
udp_gcs_port_local=$((14570+px4_instance))
```

My understanding is the following (disclaimer: I'm not super experienced with that, so take it with a grain of salt):

* Each instance gets a new sysid (incremented by 1 from the previous one)
* The simulator ports are TCP 4560, 4561, ...
* The "API ports" (i.e. MAVROS/MAVSDK) are UDP 14540, 14541, ...
* All the px4 instances run one mavlink interface on 14550 (for QGC)

I would expect to find 14550 somewhere (for QGC), but I guess that's the default when running `mavlink start` without specifying the remote port (see [here](https://github.com/PX4/Firmware/blob/master/ROMFS/px4fmu_common/init.d-posix/rcS#L265)).

The limit of 10 instances is explained by the fact that the "API mavlink interface" can go from 14540 up to 14549, but then 14550 is already taken for QGC. That's a limitation coming from the port selection, one could probably chose ports in such a way that it works with more than 10 :smile:.

Finally, all instances run on 14550 for QGC (i.e. all vehicles will appear in QGC, each with a different sysid), and I believe they could all run on 14540. My guess would be that MAVROS may not support more than one vehicle on the same port, and that's originally the reason for having ports 14540, 14541, etc (@TSC21 is that correct?). On the MAVSDK side, having multiple vehicles on the same port is only supported for MAVSDK-C++. The way we currently support multiple drones in the language bindings is by running one instance of `mavsdk_server` for each drone, on a different port, e.g.:

```
./mavsdk_server -p 50040 udp://:14540
./mavsdk_server -p 50041 udp://:14541
...
```

I hope it helps!

Jaeyoung-Lim:
@JonasVautherin Thanks for the detailed explanation! Yes the port limitation is indeed an allocation problem. I saw in the community someone running 60 at once with a smarter port numbering :)

hamishwillee:
@JonasVautherin Thanks very very much!

> On the MAVSDK side, having multiple vehicles on the same port is only supported for MAVSDK-C++. The way we currently support multiple drones in the language bindings is by running one instance of mavsdk_server for each drone, on a different port, e.g.:
```
./mavsdk_server -p 50040 udp://:14540
./mavsdk_server -p 50041 udp://:14541
```
We have this documented in MAVSDK for both C++ and Python/whatever?

I updated the docs to link to explain the 10-vehicle limit. It is superficial, but I wanted to make sure the link to this info is easy to find next time I want it :-)

hamishwillee:
Thanks @Jaeyoung-Lim et al. Merged.

karanveerdewan:
Hi folks,

I am facing a problem while using multi-vehicle SITL with Gazebo. The "iris" model works flawlessly even with 10 instances following a mission via QGCS (real time factor close to 1). But while running "plane" or "standard_vtol" model, even when they are grounded, the simulation speed real time factor drops down to ~0.1 (even at 2 instances), even lesser when adding more instances.

How do I make sure the "plane" model (multiple instances) also works at a real time factor close to 1?

TIA

Jaeyoung-Lim:
@karanveerdewan Please post a issue or for questions post on discuss.px4.io and not hijack the topic of this thread

hamishwillee:
Further, If you create a new issue you can link to this if it is relevant. However the Devguide is closed, so you shouldn't really be creating issues here.

PR #914, opened 2020-01-06
Purge bebop from docs (build guides)

hamishwillee:
Follows from https://github.com/PX4/Firmware/pull/13855


mhkabir (approved):

mhkabir:
Good go to, only wondering if we have its toolchain installation stuff still in scripts somewhere...

hamishwillee:
Thanks. Yes, docs fix for that is https://github.com/PX4/Devguide/pull/914. The env setup scripts don't do anything specific for this.

PR #915, opened 2020-01-06
setup: Fix build command for RPi

julianoes:
It seems like the `_cross` variant is not available anymore and instead `_default` refers to the cross compilation.

### DIFF
--- a/en/setup/building_px4.md
+++ b/en/setup/building_px4.md
@@ -195,17 +195,17 @@ Build the executable file:
 
 ```sh
 cd Firmware
-make emlid_navio2_cross # for cross-compiler build
+make emlid_navio2 # for cross-compiler build
 ```
 
-The "px4" executable file is in the directory **build/emlid_navio2_cross/**.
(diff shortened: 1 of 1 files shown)


hamishwillee:
Thanks @julianoes. Changes look fine to me. Merged.

hamishwillee:
PS. Would this be back in v1.10 too? I don't generally backport, but if you think important I could do that.

dagar:
> PS. Would this be back in v1.10 too? I don't generally backport, but if you think important I could do that.

No it just changed in master.

dagar:
The same _default target can be built on a desktop or directly on the target.

hamishwillee:
> The same _default target can be built on a desktop or directly on the target.

@dagar Then why is there a `emlid_navio2_native` target for native? Does this mean emlid_navio2, emlid_navio2_default, emlid_navio2_native are all the same?

What change do you want to reflect this? A note at the end of the "cross compile" section that emlid_navio2/emlid_navio2_default can also be used to compile on the native platform.?

PR #916, opened 2020-01-09
Devenv - use get-pip to get pip.

hamishwillee:
easy_intstall usually used to get pip on systems that don't have it. But this is now deprecated. Update to use the recommended method.

### DIFF
--- a/en/setup/dev_env_mac.md
+++ b/en/setup/dev_env_mac.md
@@ -44,7 +44,11 @@ brew cask install xquartz java
 Install pip if you don't already have it and use it to install the required packages:
 
 ```sh
-sudo easy_install pip
+# get pip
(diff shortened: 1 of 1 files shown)


bkueng (approved):
Looks good, but can't test it myself.

hamishwillee:
Me either. I'm going to merge the change anyway :-). Thanks!

julianoes:
I vote to use python from brew. I don't see why you would use easy_install for this (and only this) dependency.

hamishwillee:
@julianoes We use pip for all other dependencies, but this is a chicken an egg problem - there is no pip by default on macOS. If we can get pip from brew (?) then great. How?

julianoes:
Aha. I thought it comes with `brew install python`. I'll verify that.

hamishwillee:
Thanks. I see it worked for you :-).

PR #919, opened 2020-01-13
setup: use bash instead of source

julianoes:
As far as I know the command source is usually to include functions and variables defined in a script but not to run script.

I find it more intuitive to use bash to call a script.

### DIFF
--- a/en/setup/dev_env_linux_arch.md
+++ b/en/setup/dev_env_linux_arch.md
@@ -13,13 +13,13 @@ To get and run the scripts, do either of:
 * [Download PX4 Source Code](../setup/building_px4.md) and run the scripts in place:
   ```
   git clone https://github.com/PX4/Firmware.git
-  source Firmware/Tools/setup/arch.sh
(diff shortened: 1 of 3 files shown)


hamishwillee (commented):
`source` is used to run commands as though they were hand typed in the current command window. This makes a difference if you're using multiple related scripts as the context is kept. Not sure how this differs with bash.

I don't know how this differs with using `bash` (@dagar ?) . I have no problem with it if it is synonymous. Have you tested the scripts with this and know they work?

julianoes:
My only data is that I have never seen `source` anywhere else than in the PX4 Devguide world.

And I always run them like that and not with source. And I only ever run these scripts in docker anyway because I use Fedora, and because I find the scripts way too intrusive. (I'm planning to improve that.)

hamishwillee:
I run the scripts in a VM. If we can show bash works there then fine. But right now the script is broken (as you know) so can we wait on that merging before I test this?

LorenzMeier (approved):

PR #920, opened 2020-01-13
Update gazebo.md

LorenzMeier:
Fix GStreamer install instructions.

### DIFF
--- a/en/simulation/gazebo.md
+++ b/en/simulation/gazebo.md
@@ -242,11 +242,16 @@ PX4 SITL for Gazebo supports UDP video streaming from a Gazebo camera sensor att
 
 ### Prerequisites
 
-Install *Gstreamer 1.0* and its dependencies:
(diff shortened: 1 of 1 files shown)


en/simulation/gazebo.md
```diff
@@ -242,11 +242,16 @@ PX4 SITL for Gazebo supports UDP video streaming from a Gazebo camera sensor att
 
 ### Prerequisites
 
-Install *Gstreamer 1.0* and its dependencies:
+Ubuntu: Install *Gstreamer 1.0* and its dependencies:
```
hamishwillee, line 245: Thanks. FWIW the setup scripts for Linux do this set automatically except for the `gstreamer1.0-hybris` dependency: https://github.com/PX4/Firmware/blob/master/Tools/setup/ubuntu.sh#L200

I think worth leaving this in though

PR #921, opened 2020-01-14
Update C++ ModuleParams header file name

tzai:
Fixes references from the old px4_module_params.h to the new px4_platform_common/module_params.h

PR #923, opened 2020-01-15
Fix up helloworld to use correct location for log file

hamishwillee:
@bkueng Following up from https://github.com/PX4/Devguide/pull/921#issuecomment-574102667 the only other instance of the following that I found was px4_log.h (by search).
```
#include <px4_log.h> -> #include <px4_platform_common/log.h>
#include <px4_config.h> -> #include <px4_platform_common/px4_config.h>
#include <px4_tasks.h> -> #include <px4_platform_common/tasks.h>
#include <px4_posix.h> -> #include <px4_platform_common/posix.h>
```
I've updated that in the example and merged. If you find others, let me know!


bkueng:
Thanks Hamish

PR #927, opened 2020-01-21
Updates GPS failsafe description

RJHughes:
With the removal of DriverFramework there is now a new way to simulate GPS failure through the SIM_X_BLOCK parameter, so this just updates the documentation to reflect that.


bkueng (approved):
Thanks

hamishwillee:
Yes, thanks very much.

PR #935, opened 2020-01-29
Update README.md

stou-sandalski:
Added Pomegranate Systems Power Module to list of UAVCAN compatible devices.


hamishwillee:
@stou-sandalski Looks great. Has it been tested with PX4? 

You might also want to consider making it an entry page for it as a peer of https://docs.px4.io/master/en/power_module/cuav_can_pmu.html (?)

stou-sandalski:
It has been tested with the PX4 master from the 6th of January (reports version 1.11.0). Thanks for the suggestion, I was looking at the peripheral list earlier but somehow completely missed the Power Modules page.

hamishwillee:
Excellent. I will merge then. Thank you!

PR #936, opened 2020-01-30
Fix typo in system_startup.md

RicardoM17:


hamishwillee:
Thanks @RicardoM17 . The text was actually intentional - "canned" meaning "ready to use". I modified this to *predefined* (and did some other minor layout changes"

RicardoM17:
Hey @hamishwillee thanks for the heads up. To be honest I wasn't completely sure that it was a typo as like you said it did have intentional meaning but I believe that with your changes it is at least more user friendly for the non-native english users which are probably a majority :)

hamishwillee:
@RicardoM17 Yes, "canned" is slang English. More generally if you see something confusing that highlights a potential issue whether or not the writing is technically correct. 

Or to put it another way. Thanks, you're right!

PR #947, opened 2020-02-07
Remove x-plane HITL simulation info

hamishwillee:
The HIL UI for X-Plane was removed from QGC in update to use QtQuick. As there is now no way to use this simulation with PX4 this removes the docs.

@julianoes Should we also remove HILSTar airframe?


en/simulation/hitl.md
```diff
@@ -1,36 +1,28 @@
...
-PX4 supports HITL for multicopters (using jMAVSim or Gazebo) and fixed wing (using Gazebo or X-Plane demo/full version). 
+PX4 supports HITL for multicopters (using jMAVSim or Gazebo) and fixed wing (using Gazebo).
 
 
 ## HITL-Compatible Airframes {#compatible_airframe}
 
 The current set of compatible airframes vs Simulators is:
 
-Airframe | `SYS_AUTOSTART` | X-Plane | Gazebo | jMAVSim
+Airframe | `SYS_AUTOSTART` | Gazebo | jMAVSim
```
hamishwillee, line 13: @julianoes - what frames can be used for Gazebo Fixed-Wing simulation. There were none listed here previously (only HILStar for XPlane).

julianoes: I don't think one is compatible. Last time I tried it was messy and the mixers were not correct.

julianoes (approved):

en/simulation/hitl.md
hamishwillee: @julianoes OK, so that means we no longer have a Fixed wing HIL simulator right?

julianoes: I guess so. At least none that is maintained.

hamishwillee: Thanks. Not ideal.

PR #949, opened 2020-02-12
System Console: Improve mavlink shell + usage doc

hamishwillee:
@bkueng This improves shell docs in these ways:
1. Suggests QGC first for mavlink shell, which is to my mind the easiest way.
2. Moves the "how to use" after shell and makes it clear that this applies to both mavlink and system console. [with difference of boot info].
3. Adds link to list of available commands/modules.

### DIFF
--- a/en/debug/system_console.md
+++ b/en/debug/system_console.md
@@ -2,7 +2,7 @@
 
 The system console allows low-level access to the system, debug output and analysis of the system boot process. The most convenient way to connect it is by using a [Dronecode probe](https://kb.zubax.com/display/MAINKB/Dronecode+Probe+documentation), but a plain FTDI cable can be used as well.
 
-## System Console vs. Shells
+## System Console vs. Shells {#console_vs_shell}
 
 There is just one *System Console*, which runs on one specific UART (the debug port, as configured in NuttX), and is commonly attached via FTDI cable.
 - Used for *low-level debugging/development*: bootup, NuttX, startup scripts, board bringup, development on central parts of PX4 (e.g. uORB).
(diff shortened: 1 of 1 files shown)


en/debug/system_console.md
```diff
@@ -96,26 +96,51 @@ Then select 'serial connection' and set the port parameters to:
...
+   sudo pip3 install pymavlink pyserial
+   ```
+1. Open terminal (in Firmware directory) and start the shell:
+   ```sh
+   # For serial port
+   ./Tools/mavlink_shell.py /dev/ttyACM0
+   ```
+    ```sh
+   # For Wifi connection
+   ./Tools/mavlink_shell.py <IP address>
```
bkueng:
```suggestion
   ./Tools/mavlink_shell.py 0.0.0.0:14550
```
That will be the most common usage.

en/debug/system_console.md
```diff
@@ -96,26 +96,51 @@ Then select 'serial connection' and set the port parameters to:
...
+   ```
+
+Use `mavlink_shell.py -h` to get a description of all available arguments.
+
+
+
+## Getting Started on the Console/Shell {#getting-started-on-the-console}
+
+The System Console will display the system boot log when power-cycling the board.
+The MAVLink shell/console and the System Console are otherwise used in much the same way.
```
bkueng: Maybe clarify this a bit more: the system console is more low-level and can help if the system does not boot anymore (though there is an overlap with information already provided in `## System Console vs. Shells`, and it can be added there).

And with the mavlink shell, `dmesg` can be used to look at the boot output.

bkueng (commented):
Thanks @hamishwillee. Maybe we should make the distinction even clearer in the 2. sentence of the page: `The system console allows low-level access to the system, debug output and analysis of the system boot process. The most convenient way to connect it is by using a [Dronecode probe](https://kb.zubax.com/display/MAINKB/Dronecode+Probe+documentation), but a plain FTDI cable can be used as well.`
This is already specific for the system console.

en/debug/system_console.md
hamishwillee: Excellent clarification - did not know about `dmesg`. 

I have added a tip about system console being useful if system doesn't boot to the `## System Console vs. Shells`. I removed the specific information about the differences in this section and linked up to the `## System Console vs. Shells`. I also added `dmesg` as an example here.

hamishwillee: Merging, because I think this works. But if you think needs something different, please edit yourself :-)

PR #950, opened 2020-02-13
Gazebo Sim: Docs updates

hamishwillee:
This attempts to capture the results of my recent work with Gazebo. Major/intersting changes:
1. The Gazebo UI no longer needs to be updated to get the video streaming button, so that section removed.
2. Made it clear that you don't need to do anything to stream except use a target that has the camera. Also that there is only one target that supports the camera - typhoon.
3. UPdated what you actually see in the PX4 shell, which is not what was shown previously

There are still a couple of known issues, and I'd love to be able to explain how to change the world to warehouse.


en/simulation/gazebo.md
```diff
@@ -63,14 +62,23 @@ make px4_sitl gazebo
...
 
 ![3DR Solo in Gazebo](../../assets/gazebo/solo.png)
 
+
+## Typhoon H480 (Hexrotor) {#typhoon_h480}
+
+```
+make px4_sitl gazebo_typhoon_h480
+```
+
```
hamishwillee, line 79: FYI this still needs a screenshot.

en/simulation/gazebo.md
```diff
@@ -140,15 +138,26 @@ ______  __   __    ___
...
+Released under the Apache 2 License.
+http://gazebosim.org
+...
+INFO  [ecl/EKF] 5188000: commencing GPS fusion
 ```
+The console will print out status as PX4 loads the airframe-specific initialisation and parameter files, waits for (and connects to) the simulator.
+Once there is an INFO print that [ecl/EKF] is `commencing GPS fusion` the vehicle is ready to arm.
+
 
 > **Note** Right-clicking the quadrotor model allows to enable follow mode from the context menu, which is handy to keep it in view.
```
hamishwillee, line 159: FYI, this still needs to be tested - image not that great.

en/simulation/gazebo.md
```diff
@@ -197,19 +206,29 @@ For more information see: [Simulation > Run Simulation Faster than Realtime](../
 Joystick and thumb-joystick support are supported through *QGroundControl* ([setup instructions here](../simulation/README.md#joystickgamepad-integration)).
 
 
-### Simulating GPS Noise
+### Improving Distance Sensor Performance.
```
hamishwillee: Note to me. Remove fullstop.

@julianoes FYI, this is the old "Change World" section. Name was confusing.

en/simulation/gazebo.md
```diff
@@ -197,19 +206,29 @@ For more information see: [Simulation > Run Simulation Faster than Realtime](../
...
+
+The current default world is [PX4/sitl_gazebo/worlds/**iris.world**](https://github.com/PX4/sitl_gazebo/tree/master/worlds)), which uses a heightmap as ground.
```
hamishwillee, line 213: FMI, what is a heightmap and why is it a problem for distance sensors?

en/simulation/gazebo.md
```diff
@@ -236,90 +255,72 @@ This thin wrapper simply embeds app arguments like current working directories a
 To start Gazebo and PX4 separately:
 
 * Run gazebo (or any other sim) server and client viewers via the terminal:
-  ```
+  ```sh
```
hamishwillee, line 264: @julianoes The command on next line no longer works.

What is the right way to start SITL for use with an IDE? I know on VSCode that it is not needed, but presumably still for Eclipse? Who would know?

en/simulation/gazebo.md
hamishwillee: Added

en/simulation/gazebo.md
julianoes: I have no idea, I never really understood this. I would remove it for now. If people that require it want it again they can contribute that again.

What you can do though is `make px4_sitl none` which might serve the same purpose.

en/simulation/gazebo.md
```diff
@@ -236,90 +257,72 @@ This thin wrapper simply embeds app arguments like current working directories a
...
+## Video Streaming {#video}
 
 PX4 SITL for Gazebo supports UDP video streaming from a Gazebo camera sensor attached to a vehicle model.
-You can connect to this stream from *QGroundControl* (on UDP port 5600) and view video of the Gazebo environment from the simulated vehicle - just as you would from a real camera.
-The video is streamed using a *gstreamer* pipeline.
+When streaming is enabled, you can connect to this stream from *QGroundControl* (on UDP port 5600) and view video of the Gazebo environment from the simulated vehicle - just as you would from a real camera.
+The video is streamed using a *gstreamer* pipeline and can be enabled/disabled using a button in the Gazebo UI.
+
+The Gazebo camera sensor is supported/enabled on the following frames:
+* [Typhoon H480](#typhoon_h480)
```
julianoes, line 280: I believe there is now also the `iris_fpv_cam` which has the same camera.

en/simulation/gazebo.md
```diff
@@ -236,90 +257,72 @@ This thin wrapper simply embeds app arguments like current working directories a
...
+The required dependencies should already have been [installed when you set up Gazebo](#installation) (they are included in the standard PX4 installation scripts/instructions).
+
+For information only:
+- **Ubuntu:** Install *Gstreamer 1.0* and its dependencies:
+  ```
+  sudo apt-get install $(apt-cache --names-only search ^gstreamer1.0-* | awk '{ print $1 }' | grep -v gstreamer1.0-hybris) -y
+  ```
+- **Mac OS:**
+  ```
+  brew install gstreamer gst-libav gst-plugins-base gst-plugins-good gst-plugins-bad gst-plugins-ugly
```
julianoes: This is included in `brew install px4-sim-gazebo` we should just refer to that.

en/simulation/gazebo.md
```diff
@@ -236,90 +257,72 @@ This thin wrapper simply embeds app arguments like current working directories a
...
-```
-sudo apt-get install $(apt-cache --names-only search ^gstreamer1.0-* | awk '{ print $1 }' | grep -v gstreamer1.0-hybris) -y
-```
+*Gstreamer 1.0* is required for video streaming.
+The required dependencies should already have been [installed when you set up Gazebo](#installation) (they are included in the standard PX4 installation scripts/instructions).
+
+For information only:
+- **Ubuntu:** Install *Gstreamer 1.0* and its dependencies:
+  ```
+  sudo apt-get install $(apt-cache --names-only search ^gstreamer1.0-* | awk '{ print $1 }' | grep -v gstreamer1.0-hybris) -y
```
julianoes: These dependencies are included in `ubuntu.sh`. Why not just link to that?

hamishwillee: @julianoes Essentially I do above all this by saying "this is all installed previously using standard installation - this is for info only". I figured it might be useful for people to see what it is that is needed. I could delete both this and the mac stuff if you think a good ideas?

julianoes (approved):

en/simulation/gazebo.md
julianoes: It's a very complicated command but yes :smile: it can be useful. (I would not even try it out).

en/simulation/gazebo.md
hamishwillee: @julianoes I'm not sure. What does `_ide` actually *do*? Note there are a bunch of ide options still present - e.g. `gazebo_iris_ide` so not sure.

Do you know anyone who debugs using these or who might know?

en/simulation/gazebo.md
hamishwillee: OK, in that case I have done as you suggested and removed the command. However I've provided an FYI only list of the dependencies so they can check at least for those.

en/simulation/gazebo.md
hamishwillee: @julianoes I can't find `iris_fpv_cam` (or `gazebo_iris_fpv_cam`) or similar searching source tree, or issues or checking all make targets. I tried one ofthe `_vision` targets but that didn't work. 

Thoughts?

en/simulation/gazebo.md
julianoes: I just checked, this actually still works. You just need to add these extra underscores.

en/simulation/gazebo.md
julianoes: I think @Jaeyoung-Lim is about to clean this up.

Jaeyoung-Lim: Those targets were used by ROS, which doesn't require the corresponding world definitions. I will fix this in the world spawning PR :)

en/simulation/gazebo.md
hamishwillee: @julianoes Can you be explicit? ie ` make px4_sitl gazebo_none_ide` changes to what?

en/simulation/gazebo.md
hamishwillee: Ah, so for now this is correct. Closing. Thanks.

en/simulation/gazebo.md
julianoes: These two variants work:
```
make px4_sitl gazebo___ide
make px4_sitl gazebo_iris_ide
```

hamishwillee: Thanks very much. Integrated and will now merge.

PR #953, opened 2020-02-17
Split out MAVLink Shell

hamishwillee:
The MAVLink Shell is inside a doc named System Console (so not obvious to find). The idea here is to split it out for visibility. I've done this by creating a parent doc that covers the differences and includes the common information on how to use them to enter commands.

The system console and mavlink shell docs are then simpler. 

As part of this I realised that the system console doc should be even simpler. It suffers from the same problem as SWD - every device has different pins. I have pushed the Pixhawk 1 docs into the autopilot do here: https://github.com/PX4/px4_user_guide/pull/657
So next step would be to remove that from here and cross link. 
What I will keep in this doc is links to autopilots, pixhawk standard setup info, and overview.

Hope this makes sense.


en/debug/mavlink_shell.md
```diff
@@ -0,0 +1,39 @@
...
+
+> **Tip** The [QGroundControl MAVLink Console](#qgroundcontrol) is the easiest way to access the console.
+  If the system does not start properly you should instead use the [System Console](../debug/system_console.md).
+
+## Opening the Shell
+
+### QGroundControl MAVLink Console {#qgroundcontrol}
+
+The easiest way to access shell is to use the [QGroundControl MAVLink Console](https://docs.qgroundcontrol.com/en/analyze_view/mavlink_console.html) (see **Analyze View > Mavlink Console**).
+
```
hamishwillee, line 16: Probably will extend this with an image. Looks a bit silly as just a section with a sentence.

en/debug/system_console.md
```diff
@@ -39,17 +50,18 @@ Connect the 6-pos JST SH 1:1 cable to the Dronecode probe or connect the individ
 |5         | SWCLK      |         | N/C   |
 |6         | GND     | 1       | FTDI GND (black)   |
 
-## Pixhawk v1: Wiring the Console
+### Pixhawk v1
```
hamishwillee: FYI section I plan to remove, with info moved into https://github.com/PX4/px4_user_guide/pull/657

en/debug/system_console.md
```diff
@@ -70,6 +82,13 @@ The complete wiring is shown below.
 
 ![Console Debug](../../assets/console/console_debug.jpg)
 
+
+### Snapdragon Flight
```
hamishwillee: @bkueng Not sure about this section, and how it works with the other snapdragon flight section. Again though, I  proposed to move this into the Snapdragon docs.

What I THINK is that essentially this is essentially a debug adaptor board to which you connect the FTDI cable?

en/debug/system_console.md
```diff
@@ -102,53 +121,9 @@ Then select 'serial connection' and set the port parameters to:
...
-    ```sh
-   # For Wifi connection
-   ./Tools/mavlink_shell.py 0.0.0.0:14550
-   ```
-
-Use `mavlink_shell.py -h` to get a description of all available arguments.
-
-
-## Getting Started on the Console/Shell {#getting-started-on-the-console}
+### Snapdragon DSP Console
```
hamishwillee: So this is I think how you access the PX4 console. Ie you can't just use the terminal app out of the box but have to run some other stuff first. Is that correct?

Because this is different, I proposed to also split this out into snapdragon setup docs too. OK?

en/debug/consoles.md
```diff
@@ -0,0 +1,43 @@
...
+
+The PX4 *System Console* provides low-level access to the system, debug output and analysis of the system boot process.
+
+There is just one *System Console*, which runs on one specific UART (the debug port, as configured in NuttX), and is commonly attached to a computer via an FTDI cable (or some other debug adapter like  a [Dronecode probe](https://kb.zubax.com/display/MAINKB/Dronecode+Probe+documentation)).
+- Used for *low-level debugging/development*: bootup, NuttX, startup scripts, board bringup, development on central parts of PX4 (e.g. uORB).
+- In particular, is the only place where all boot output (including information about applications auto-started on boot) is printed.
+
+Shells provide higher-level access to the system:
+- Used for basic module testing/running commands.
+- Only display the output of modules you start (and therefore cannot debug the boot process).
```
bkueng: Partially: as long as the system boots, you can still use `dmesg` for debugging. In general `dmesg` is very powerful and brings the MAVLink shell much closer to the system shell (with `dmesg -f &` for example, you also see the output of background tasks).

en/debug/mavlink_shell.md
```diff
@@ -0,0 +1,39 @@
+# MAVLink Shell
+
+The MAVLink Shell is an *NSH console* that can be accessed via MAVLink over serial (USB/Telemetry) or WiFi (UDP/TCP) links (in particular, on NuttX-based systems like: Pixhawk, Pixracer, etc.).
+
+The shell can be used for running commands and modules (and displaying the output of modules that it starts). 
+It cannot display the output of tasks running on the work queue (or modules that it does not start).
```
bkueng: You can do that now with `dmesg -f &`

en/debug/system_console.md
bkueng: I think so, but I'm not entirely familiar either

en/debug/system_console.md
bkueng: Yes this is a lot different on Snapdragon and very specific to that platform, so splitting it out makes sense.

bkueng (approved):
Looks good.
I would specifically mention `dmesg`, as it is quite new and very useful, specifically to inspect bootup and background tasks (though it is not enabled on all boards - only the ones with enough free RAM).

en/debug/consoles.md
hamishwillee: Thanks @bkueng. I re-wrote this - can you please check. Tried to capture that you can only "directly" see stuff you start, but dmesg allows indirect access to that.

en/debug/mavlink_shell.md
hamishwillee: Fixed (as per https://github.com/PX4/Devguide/pull/953/files#r380466900). Can you sanity check the change.

en/debug/mavlink_shell.md
hamishwillee: But not now!

en/debug/system_console.md
hamishwillee: Now moved and linked

en/debug/system_console.md
hamishwillee: I've moved them. There were stub docs that the content just slipped into.

hamishwillee:
FYI last bit I'd like to do before merging is remove the pixracer/pixhawk3 section - see https://github.com/PX4/px4_user_guide/pull/665

en/debug/consoles.md
```diff
@@ -0,0 +1,48 @@
...
+
+```bash
+nsh> ls
+nsh> free
+nsh> dmesg
+```
+
+Many other system commands and modules are listed in the [Modules and Command Reference](../middleware/modules_main.md) (e.g. `top`, `listener`, etc.).
+
+> **Tip** Some commands may be disabled on some boards (i.e. the some modules are not included in firmware for boards with RAM constraints).
```
bkueng:
```suggestion
> **Tip** Some commands may be disabled on some boards (i.e. the some modules are not included in firmware for boards with RAM or FLASH constraints).
```

bkueng (approved):

hamishwillee:
Merging as is. Will further update when https://github.com/PX4/px4_user_guide/pull/665 merges

PR #956, opened 2020-02-20
micrortps: update to fit new API and ROS2 Dashing support

TSC21:
This is a general update to fit the changes done on the `px4_ros_com`, which bring ROS2 Dashing support and new API for the nodes and scripts.

### DIFF
--- a/en/middleware/micrortps.md
+++ b/en/middleware/micrortps.md
@@ -70,7 +70,7 @@ This is needed because the first version of ROS does not support RTPS.
 
 ## Code generation
 
-> **Note** [Fast RTPS must be installed](../setup/fast-rtps-installation.md) in order to generate the required code!
+> **Note** [Fast RTPS 1.8.2 or later must be installed](../setup/fast-rtps-installation.md) in order to generate the required code!
 > *Fast RTPS* is installed *by default* if you use the normal installers/scripts for [macOS](../setup/dev_env_mac.md), [Windows Cygwin](../setup/dev_env_windows_cygwin.md) or [Ubuntu](../setup/dev_env_linux_ubuntu.md).
 
 ### ROS-independent applications
@@ -132,10 +132,10 @@ rtps:
     send: true
 ```
 
-> **Note** The `px4_msgs` build process (only) runs the CMake macro `rosidl_generate_interfaces()` to generate ROS2 message header files, while `px4_ros_com` build process generates the IDL files through the `rosidl_generate_dds_interfaces()` CMake macro.
+> **Note** Currently, the `px4_msgs` build process (only) runs the CMake macro `rosidl_generate_interfaces()` to generate ROS2 message header and source files and also the required IDL files, but previously it was using the `rosidl_generate_dds_interfaces()` CMake macro for the same purpose. A change on the API between the ROS2 Crystal and Dashing versions required that the build process was also adjusted to generate different headers a source files according to the versions.
 > The PX4 Firmware includes a template for the IDL file generation, which is only used during the PX4 build process.
 >
-> The `px4_ros_com` build generates *slightly different* IDL files for use with ROS2/ROS (than are built for PX4 firmware).
+> The `px4_msgs` build generates *slightly different* IDL files for use with ROS2/ROS (than are built for PX4 firmware).
 > The **uorb_rtps_message_ids.yaml** is transformed in a way that the message names become *PascalCased*
 > (the name change is irrelevant to the client-agent communication, but is critical for ROS2, since the message naming must follow the PascalCase convention).
 > The new IDL files also reverse the messages that are sent and received
@@ -161,16 +161,16 @@ The *Client* application can be launched from [NuttShell/System Console](../debu
 The command syntax is shown below (you can specify a variable number of arguments):
 
 ```sh
-> micrortps_client start|stop [options]
+> micrortps_client start|stop|status [options]
   -t <transport>          [UART|UDP] Default UART
   -d <device>             UART device. Default /dev/ttyACM0
-  -u <update_time_ms>     Time in ms for uORB subscribed topics update. Default 0
   -l <loops>              How many iterations will this program have. -1 for infinite. Default -1.
   -w <sleep_time_ms>      Time in ms for which each iteration sleep. Default 1ms
   -b <baudrate>           UART device baudrate. Default 460800
   -p <poll_ms>            Time in ms to poll over UART. Default 1ms
   -r <reception port>     UDP port for receiving. Default 2019
   -s <sending port>       UDP port for sending. Default 2020
+  -i <ip_address>         Select IP address (remote) values: <x.x.x.x>. Default: 127.0.0.1
 ```
 
 > **Note** By default the *Client* runs as a daemon, but you will need to start it manually.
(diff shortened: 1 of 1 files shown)


en/middleware/micrortps.md
```diff
@@ -236,23 +236,29 @@ and separately clone the `px4_ros_com` and `px4_msgs` repo for both the `master`
...
+   ```
+
 1. *setuptools* must also be installed (using *pip* or *apt*):
 
    ```sh
    sudo pip3 install -U setuptools
    ```
 
-   > **Note** This install and build guide is not applicable to ROS2 Ardent anymore, since this has reached EOL December 2018.
+   > **Note** This install and build guide is not applicable to ROS2 Ardent, Bouncy or Crystal anymore, since they all reached EOL.
```
hamishwillee: This should move to the top of the install section. It should say something like "This install and build guide explains how to install X, Y, Z. It does cover ROS3 Ardent, ...."

en/middleware/micrortps.md
```diff
@@ -70,7 +70,7 @@ This is needed because the first version of ROS does not support RTPS.
 
 ## Code generation
 
-> **Note** [Fast RTPS must be installed](../setup/fast-rtps-installation.md) in order to generate the required code!
+> **Note** [Fast RTPS 1.8.2 or later must be installed](../setup/fast-rtps-installation.md) in order to generate the required code!
 > *Fast RTPS* is installed *by default* if you use the normal installers/scripts for [macOS](../setup/dev_env_mac.md), [Windows Cygwin](../setup/dev_env_windows_cygwin.md) or [Ubuntu](../setup/dev_env_linux_ubuntu.md).
```
hamishwillee: @TSC21 we should remove Ubuntu from this: "Fast RTPS is installed by default if you use the normal installers/scripts for macOS, Windows Cygwin or Ubuntu."

hamishwillee:
It might be nice to do something with these two paragraphs.

> The px4_ros_com package, when built, generates everything needed to access PX4 uORB messages from a ROS2 node (for ROS you also need ros1_bridge). This includes all the required components of the PX4 RTPS bridge, including the micrortps_agent and the IDL files (required by the micrortps_agent).
>
>The ROS and ROS2 message definition headers and interfaces are generated from the px4_msgs package, which match the uORB messages counterparts under PX4 Firmware. These are required by px4_ros_com when generating the IDL files to be used by the micrortps_agent.

The problem with this is it refers to `px4_ros_com` and `px4_msgs` without context. Why do we need this/why do we do it this way? 

I personally am not sure how this works. I THINK that what happens is that:
- there are a set of uorb topics that we support for interworking with ROS2 (and via that to ROS1).
- these are defined in PX4 Firmware, which builds and includes the client side.
- As part of that build process the px4_msgs is created.
- From px4_ros_com uses px4_msgs.

But it could be that px4_msgs is used by both Firmware AND ROS to define the mapped topics (or that is what it looks like)????

Either way px4_msgs and px4_ros_com essentially statically define the current PX4 "ROS2/uorb interface". Doing it this way is a separation of concerns, and means that you can target particular systems without having to have all their source.

Presumably there is a way to map the interface as it changes? I.e. if I'm using PX4 v1.8, presumably I can fetch that set of msgs and px4_ros_com? Do we need to mention that?

en/middleware/micrortps.md
```diff
@@ -132,10 +132,10 @@ rtps:
     send: true
 ```
 
-> **Note** The `px4_msgs` build process (only) runs the CMake macro `rosidl_generate_interfaces()` to generate ROS2 message header files, while `px4_ros_com` build process generates the IDL files through the `rosidl_generate_dds_interfaces()` CMake macro.
+> **Note** Currently, the `px4_msgs` build process (only) runs the CMake macro `rosidl_generate_interfaces()` to generate ROS2 message header and source files and also the required IDL files, but previously it was using the `rosidl_generate_dds_interfaces()` CMake macro for the same purpose. A change on the API between the ROS2 Crystal and Dashing versions required that the build process was also adjusted to generate different headers a source files according to the versions.
```
hamishwillee: There are some minor problems with text above. Is this still accurate:

> **Note** The `px4_msgs` build process (only) runs the CMake macro `rosidl_generate_interfaces()` to generate ROS2 message header and source files, and also the required IDL files (previously it used the `rosidl_generate_dds_interfaces()` CMake macro for the same purpose). This change was required because  different headers and source files are now needed for ROS2 Crystal and Dashing versions.

en/middleware/micrortps.md
```diff
@@ -236,23 +236,29 @@ and separately clone the `px4_ros_com` and `px4_msgs` repo for both the `master`
 
 ### Installing ROS and ROS2 and respective dependencies
 
-In order to install ROS Melodic and ROS2 Crystal (officially supported) or Bouncy on a Ubuntu 18.04 machine, follow the links below, respectively:
+In order to install ROS Melodic and ROS2 Dashing (officially supported) or Bouncy on a Ubuntu 18.04 machine, follow the links below, respectively:
```
hamishwillee: Below you say that Bouncy is not supported.

en/middleware/micrortps.md
TSC21: Yes we should. I will create a build script for the ROS2 development environment after and we can add it here.

en/middleware/micrortps.md
TSC21: Sure.

TSC21:
> As part of that build process the px4_msgs is created.

`px4_msgs` is a standard repo where the ROS message definitions are kept. Those correspond to the uORB counterparts. But the repo by itself is not *created* in the build process - the repo exists and what happens is that in Jenkins master build, the message files are deployed to this repo. But this is not relevant to this guide. What's relevant is that for the bridge to work with ROS 2, you need the `px4_msgs` package so to get the IDL and header files it provides to use them in the microRTPS agent and on the ROS2 nodes.

> But it could be that px4_msgs is used by both Firmware AND ROS to define the mapped topics (or that is what it looks like)????

No. `px4_msgs` is used on the ROS side. What we do on the Firmware side is that we generated our own IDL files and used them to generated the header and source files for the client (and agent code if we build the microRTPS agent without using `px4_ros_com`). While if you build `px4_msgs` on a colcon/ament workspace, you used ROS generators to generate the IDL files you require.

> Either way px4_msgs and px4_ros_com essentially statically define the current PX4 "ROS2/uorb interface". Doing it this way is a separation of concerns, and means that you can target particular systems without having to have all their source.

That is true. It also means that I wouldn't have to always build both under a workspace if they were released as deb packages that you can install with `apt`. I did it already with `px4_msgs`, but that uORB message definitions change would make the release versioning hard - I still have to figure out a better way of doing it, if I just align the `px4_msgs` release with the upstream releases of PX4 Firmware or if I do multiple releases and have a document aligning them with the Firmware commit hash.

> Presumably there is a way to map the interface as it changes? I.e. if I'm using PX4 v1.8, presumably I can fetch that set of msgs and px4_ros_com? Do we need to mention that?

As I say above, yes I need to find a better way of aligning that, since currently we consider that the user needs to align always both masters to actually use the bridge. So I need to think of a better versioning system at this point.

en/middleware/micrortps.md
TSC21: > This change was required because different headers and source files are now needed for ROS2 Crystal and Dashing versions.

Well not exactly. The problem was that before Dashing (meaning Crystal and previous versions), the `rosidl_generate_interfaces()`, or generated IDL files which weren't compatible with the ones that `fastrtpsgen` required to generated the code for the agent (in Crystal), or they didn't generate any IDL files (which was the case of Ardent and Bouncy). So you required to use `rosidl_generate_dds_interfaces()` to generate the IDL files you required.

What happened is that an API change between ROS2 Crystal and ROS2 Dashing releases on the `rosidl_generate_dds_interfaces()` CMake module led to `rosidl_generate_dds_interfaces()` not generating IDL files from the injected msg files (it now rather expects that you feed a tuple composed by the path of the file and an IDL file that will be adjusted to the IDL file spec that `ros_idl` expects). On the other way, `rosidl_generate_interfaces()` now generates the IDL files you need to use on `fastrtpsgen` to generate the agent code.

TL;DR, what is required to say is that an API change led us to rely only in the `rosidl_generate_interfaces()` CMake module (used in `px4_msgs`) to generate the required IDL files we use on the microRTPS agent generation in `px4_ros_com`.

en/middleware/micrortps.md
TSC21: yep that's a typo.

en/middleware/micrortps.md
hamishwillee: OK. I will update this one.

en/middleware/micrortps.md
hamishwillee: OK I will removed. FYI, Feel free to directly edit your own doc to fix :-)

en/middleware/micrortps.md
hamishwillee: I have fixed.

en/middleware/micrortps.md
hamishwillee: Yeah, but you made me read it :-)

I've updated pretty much like you said in https://github.com/PX4/Devguide/pull/956/commits/7b4ab468e9f4318c355af6ac6e549ae0e4df8ba9 - OK?

TSC21: Yes that looks good!

hamishwillee:
@TSC21 This is merged. Feel free to patch the note if you didn't like the change. I decided not to change anything in response to https://github.com/PX4/Devguide/pull/956#issuecomment-592470661  [I get it now, and it is fairly clear and accurate] - we should address that again though once you have a versioning strategy in place.

PR #962, opened 2020-02-26
UAVCAN: note dependecy on sdcard

hamishwillee:
Comes from https://github.com/PX4/Firmware/issues/14206

### DIFF
--- a/en/uavcan/notes.md
+++ b/en/uavcan/notes.md
@@ -12,13 +12,10 @@ A great tool to debug the transmission on the UAVCAN bus is the [Zubax Babel](ht
 They can also be used independently from Pixhawk hardware in order to test a node or manually control UAVCAN enabled ESCs.
 
 
+### UAVCAN devices dont get node ID/FW doesn't update
 
-
-
-
-
-
-
+PX4 requires an SD card for UAVCAN node allocation and firmware upgrade (both of which happen during boot).
+Check that there is a (working) SD card present and reboot.
 
 
 
--- a/en/uavcan/notes.md
+++ b/en/uavcan/notes.md
@@ -8,7 +8,8 @@ If the PX4 Firmware arms but the motors do not start to rotate, check the parame
 
 ### Debugging with Zubax Babel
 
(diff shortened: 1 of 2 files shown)


en/uavcan/notes.md
```diff
@@ -8,7 +8,8 @@ If the PX4 Firmware arms but the motors do not start to rotate, check the parame
 
 ### Debugging with Zubax Babel
 
-A great tool to debug the transmission on the UAVCAN bus is the [Zubax Babel](https://docs.zubax.com/zubax_babel) in combination with the [GUI tool](http://uavcan.org/GUI_Tool/Overview/). They can also be used independently from Pixhawk hardware in order to test a node or manually control UAVCAN enabled ESCs.
+A great tool to debug the transmission on the UAVCAN bus is the [Zubax Babel](https://docs.zubax.com/zubax_babel) in combination with the [GUI tool](http://uavcan.org/GUI_Tool/Overview/).
+They can also be used independently from Pixhawk hardware in order to test a node or manually control UAVCAN enabled ESCs.
```
hamishwillee, line 13: @AlexKlimaj This is kind of a troubleshooting page on working with UAVCAN. Is there a good way of describing the symptoms of not having an SD Card? Ie in very first event would it simply be that motors don't turn, power modules don't provide information ..... or is there some other symptom?

If not, how about something like: ....

### UAVCAN devices dont get node ID/FW doesn't update

PX4 requires an SD card for UAVCAN node allocation and firmware upgrade (both of which happen during boot).
Check that there is a (working) SD card present and reboot.

AlexKlimaj: All I have to test with are my smart batteries. The symptom is that they don't get a Node ID and the battery information isn't transmitted. So I think what you have is correct.

hamishwillee:
Thanks @AlexKlimaj ! Are you satisfied that, had this note existed, you would have had a significant chance of working out the error?

AlexKlimaj:
@hamishwillee Yes, that would have helped and I wouldn't have had to ask.

PR #964, opened 2020-03-02
test_and_ci: install mavsdk on Arch Linux

julianoes:


hamishwillee:
@julianoes As per my last comment to your MAVSDK PR for "getting started" I think that should be named to "Installation". Then all this stuff would just link to that document.  [a quick start would then be a simplified example for just one platform]

I'll merge this now, but let's discuss.

PR #967, opened 2020-03-04
change documentation on setup windows bash toolchain 

bestaps:
[https://github.com/PX4/Devguide/blob/master/en/setup/dev_env_windows_bash_on_win.md](url)

this documentation provided a command for adding a line to the environment (bash profile to make the change permanent)
it goes like 
`export PATH=$HOME/gcc-arm-none-eabi-5_4-2017q2/bin:\$PATH
`which causes error and removes access to /etc/bin files of the root.... 

the correction is `export PATH=$HOME/gcc-arm-none-eabi-5_4-2017q2/bin:$PATH`

adding a pull request for the same....

### DIFF
--- a/en/setup/dev_env_windows_bash_on_win.md
+++ b/en/setup/dev_env_windows_bash_on_win.md
@@ -88,5 +88,5 @@ To add this compiler to your environment manually:
    ```
 1. Add the to the environment (add the line to your bash profile to make the change permanent)
    ```
-   export PATH=$HOME/gcc-arm-none-eabi-5_4-2017q2/bin:\$PATH
+   export PATH=$HOME/gcc-arm-none-eabi-5_4-2017q2/bin:$PATH
    ```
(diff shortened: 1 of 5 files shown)


hamishwillee:
@prem48science Thanks! Can you please delete the changes except to the file in /en tree (e.g. zh, ko variants). The reason is that the other files get created by a translation service and will be "round tripped" through our translation service. If you leave this in then I would have to manage a merge.

@MaEtUgR Can you please confirm this change is suitable?

bestaps:
@hamishwillee sure. I deleted changes except in /en tree and commited the changes.

MaEtUgR (approved):
Good catch, thanks @prem48science !

MaEtUgR:
Why do these work nicely in our ubuntu scripts? https://github.com/PX4/Firmware/blob/master/Tools/setup/arch.sh#L103
Let me check.

MaEtUgR:
**All good** 💡, that is where the error came from, the original documenter copied the line from a setup script that first adds the command into a string without resolving `$PATH` and hence escaping the `$` before executing the command in the string with environment variable resolution afterwards.

bestaps:
@MaEtUgR, yeah that might be cause of error. anyhow its resolved. thank you.

bestaps:
oops! a mistake in hurray...

hamishwillee:
Thanks very much. Merging now!

PR #972, opened 2020-03-12
PX4 SD Card layout info

hamishwillee:
@bkueng As discussed in slack.

### DIFF
--- a/en/concept/sd_card_layout.md
+++ b/en/concept/sd_card_layout.md
@@ -1,0 +1,20 @@
+# PX4 SD Card Layout
+
+The PX4 SD Card is used for storing configuration files, flight logs, mission information etc. 
+
+> **Note** The file system is FAT32.
+
+The directory structure/layout is shown below.
+
+Directory/File(s) | Description
+--- | ---
+/etc/ | Extra config (+ mixers)
+/etc/mixers/ | [Mixers](../concept/mixing.md)
+/log/ | Full [flight logs](../log/logging.md)
+/mission_log | Reduced flight logs
+/fw/ | [UAVCAN](../uavcan/README.md) firmware
+/uavcan.db/ | UAVCAN DB + logs
+/params | Parameters (if not in FRAM/FLASH)
+/dataman | Mission storage file
+/fault_<datetime>.txt | Hardfault files
+/bootlog.txt | Boot log file
--- a/en/concept/sd_card_layout.md
+++ b/en/concept/sd_card_layout.md
@@ -2,16 +2,17 @@
 
 The PX4 SD Card is used for storing configuration files, flight logs, mission information etc. 
 
-> **Note** The file system is FAT32.
+> **Note** The default file system is FAT32.
+  SD card formatting is not required before use.
 
 The directory structure/layout is shown below.
 
 Directory/File(s) | Description
 --- | ---
-/etc/ | Extra config (+ mixers)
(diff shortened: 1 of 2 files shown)


en/concept/sd_card_layout.md
```diff
@@ -0,0 +1,20 @@
...
+
+The PX4 SD Card is used for storing configuration files, flight logs, mission information etc. 
+
+> **Note** The file system is FAT32.
+
+The directory structure/layout is shown below.
+
+Directory/File(s) | Description
+--- | ---
+/etc/ | Extra config (+ mixers)
```
bkueng: Link to https://dev.px4.io/master/en/concept/system_startup.html#replacing-the-system-startup?

en/concept/sd_card_layout.md
```diff
@@ -0,0 +1,20 @@
...
+/etc/mixers/ | [Mixers](../concept/mixing.md)
+/log/ | Full [flight logs](../log/logging.md)
+/mission_log | Reduced flight logs
```
bkueng:
```suggestion
/mission_log/ | Reduced flight logs
```

en/concept/sd_card_layout.md
```diff
@@ -0,0 +1,20 @@
+# PX4 SD Card Layout
+
+The PX4 SD Card is used for storing configuration files, flight logs, mission information etc. 
+
+> **Note** The file system is FAT32.
```
bkueng: Maybe note that this is the default, and you don't have to format before using it?

bkueng (approved):
Thanks @hamishwillee

en/concept/sd_card_layout.md
hamishwillee: Yes. That implies you can use other formats if you like. I've updated, but if we can change formats, we should probably link to how/where?

en/concept/sd_card_layout.md
hamishwillee: Yes!

hamishwillee:
Thanks for the review @bkueng. There is a comment above, but merging this as is OK for now.

en/concept/sd_card_layout.md
bkueng: Sorry, what do you mean by that?

hamishwillee: > Sorry, what do you mean by that?

So you said "Maybe note that this is the default". If something has a default of FAT32, that implies that you might have another file system. Right?

If so, what are the other options for a file system, and how do we set PX4 to use them?

bkueng: I see. You can format an SD card to any file system you want. But that's your conscious choice then and you have to make sure you switch back to FAT32 again if you want to use it with PX4 (we don't really support others).

hamishwillee: Thanks. So I had assumed that when you said "formatting is not required before use" the card will be formatted with FAT32 automatically by PX4 if it wasn't already. Is that the case.

If not, we should just say use a formatted FAT32 SD card, and ignore any nuances.

bkueng: There's logic to reformat the SD card if mouting fails, but I rather suggest users to format themselves (I'm not sure how reliable NuttX does it).

hamishwillee: I agree. Done in https://github.com/PX4/Devguide/pull/980.

Thanks.

PR #973, opened 2020-03-13
Custom payload mixers

hamishwillee:
Beat, @LorenzMeier created this doc for improving mixer doc in general and payload doc in particular
https://docs.google.com/document/d/1xCEQh48uDWyo7TjqedW6gYxBxMtNyuYZ2Xkt2MBb2-w/edit

I have played around a bit but I still have questions. So I've decided to do a minimal update for payload and think about this some more in future. The doc is linked from source.

Thoughts welcome - and answers to specific questions. This is still draft.

### DIFF
--- a/en/concept/custom_mixer_payload.md
+++ b/en/concept/custom_mixer_payload.md
@@ -61,10 +61,25 @@ Remove the first section with a payload control group function input:
 Because this output is in the first position in the file it will map to the first AUX PWM output (unless UAVCAN is enabled).
 This output will now respect updates to the payload control group (6) output 1.
 
+In the code, control group 6 will need to be defined as well:
+- Add `actuator_controls_6` to the TOPICS definition in https://github.com/PX4/Firmware/blob/master/msg/actuator_controls.msg#L17
+- Increase `NUM_ACTUATOR_CONTROL_GROUPS` to 7 in the same file.
+- Subscribe to the additional control group in the output library under
+  https://github.com/PX4/Firmware/blob/master/src/lib/mixer_module/mixer_module.cpp#L52. It should look like this:
+  ```
+	{&interface, ORB_ID(actuator_controls_0)},
+	{&interface, ORB_ID(actuator_controls_1)},
+	{&interface, ORB_ID(actuator_controls_2)},
+	{&interface, ORB_ID(actuator_controls_3)},
+	{&interface, nullptr},
+	{&interface, nullptr},
+	{&interface, ORB_ID(actuator_controls_6)},
+  ```
+
 Putting an output on group 6 works by publishing actuator control group 6. First you have to create the publication.
 This should happen once when the PX4 module is initialized (look for places where this pattern is already being used):
 ```
-uORB::Publication<actuator_controls_s> _actuator_controls_pub{ORB_ID(actuator_controls_0)};
+uORB::Publication<actuator_controls_s> _actuator_controls_pub{ORB_ID(actuator_controls_6)};
 ```
 
 Then you need to publish the first message:
--- a/en/concept/custom_mixer_payload.md
+++ b/en/concept/custom_mixer_payload.md
@@ -38,8 +38,8 @@ S: 3 4  10000  10000      0 -10000  10000
 > **Note** The file defines a [simple mixer](../concept/mixing.md#simple-mixer) with four outputs.
 > - `M: 1` indicates an output that is defined by one control input (the following `S` line).
 > - `S: 3 `_`n`_ indicates that the input is the n<>th<> input of [Control Group 3 (Manual Passthrough)](../concept/mixing.md#control-group-3-manual-passthrough).
->  So for `S: 3 5` the input is called "RC aux1", and will output the value of the channel mapped by parameter `RC_MAP_AUX1`.
-> - The order of output declaration of 
+>  So for `S: 3 5` the input is called "RC aux1" (this maps to the RC channel set in parameter `RC_MAP_AUX1`).
+> - The section declaration order defines the order of the outputs when assigned to a physical bus (e.g. the third section might be assigned to AUX3).
 
 
 Start by copying the mixer file and putting it onto the SD Card at: **/etc/mixers/pass.aux.mix** (see [Mixing and Actuators > Loading a Custom Mixer](../concept/mixing.md#loading_custom_mixer).
--- a/en/concept/custom_mixer_payload.md
+++ b/en/concept/custom_mixer_payload.md
@@ -61,22 +61,28 @@ Remove the first section with a payload control group function input:
 Because this output is in the first position in the file it will map to the first AUX PWM output (unless UAVCAN is enabled).
 This output will now respect updates to the payload control group (6) output 1.
 
-In the code, control group 6 will need to be defined as well:
-- Add `actuator_controls_6` to the TOPICS definition in https://github.com/PX4/Firmware/blob/master/msg/actuator_controls.msg#L17
+Control group 6 will need to be defined in the code as well (it is missing!):
+- Add `actuator_controls_6` to the TOPICS definition in [/msg/actuator_controls.msg](https://github.com/PX4/Firmware/blob/master/msg/actuator_controls.msg#L17):
+  ```
+  # TOPICS actuator_controls actuator_controls_0 actuator_controls_1 actuator_controls_2 actuator_controls_3 actuator_controls_6
+  ```
 - Increase `NUM_ACTUATOR_CONTROL_GROUPS` to 7 in the same file.
-- Subscribe to the additional control group in the output library under
-  https://github.com/PX4/Firmware/blob/master/src/lib/mixer_module/mixer_module.cpp#L52. It should look like this:
+- Subscribe to the additional control group in the output library ([/src/lib/mixer_module/mixer_module.cpp#L52](https://github.com/PX4/Firmware/blob/master/src/lib/mixer_module/mixer_module.cpp#L52)) in the `MixingOutput` constructor.
+  It should look like this:
   ```
 	{&interface, ORB_ID(actuator_controls_0)},
 	{&interface, ORB_ID(actuator_controls_1)},
 	{&interface, ORB_ID(actuator_controls_2)},
 	{&interface, ORB_ID(actuator_controls_3)},
+  ```
+  ```
 	{&interface, nullptr},
 	{&interface, nullptr},
 	{&interface, ORB_ID(actuator_controls_6)},
   ```
 
-Putting an output on group 6 works by publishing actuator control group 6. First you have to create the publication.
+Putting an output on group 6 works by publishing actuator control group 6.
+First you have to create the publication.
 This should happen once when the PX4 module is initialized (look for places where this pattern is already being used):
 ```
 uORB::Publication<actuator_controls_s> _actuator_controls_pub{ORB_ID(actuator_controls_6)};
--- a/en/concept/custom_mixer_payload.md
+++ b/en/concept/custom_mixer_payload.md
@@ -1,0 +1,77 @@
+# Custom Payload Mixers
+
+This topic shows how to add a custom [mixer](../concept/mixing.md) for programmatically controlling a custom payload (e.g., an electromagnetic gripper).
+
+The topic is intended for developers who want to support payload types that do not have existing control group definitions (e.g. gimbals have a control group, but grippers do not).
+You should already have read [Mixing and Actuators](../concept/mixing.md).
+
+
+## Payload Mixer Example
+
+A payload mixer is just a [simple mixer](../concept/mixing.md#simple-mixer) that maps any of the function values from [Control Group #6 (First Payload)](../concept/mixing.md#control_group_6) to a particular output.
+You can then publish uORB topics to the selected control group function and their value will be mapped to the specified output.
+
+For this example, we'll create a custom mixer based on the *RC passthrough mixer* ([pass.aux.mix](https://github.com/PX4/Firmware/blob/master/ROMFS/px4fmu_common/mixers/pass.aux.mix)).
+This mixer is commonly loaded into the AUX PWM ports on large multicopters). 
+It passes through the values of 4 user-defined RC channels (set using the [RC_MAP_AUXx/RC_MAP_FLAPS](../advanced/parameter_reference.md#RC_MAP_AUX1) parameters) to the first four outputs on the AUX PWM output.
+
+```
+# Manual pass through mixer for servo outputs 1-4
+
+# AUX1 channel (select RC channel with RC_MAP_AUX1 param)
+M: 1
+S: 3 5  10000  10000      0 -10000  10000
+
+# AUX2 channel (select RC channel with RC_MAP_AUX2 param)
+M: 1
+S: 3 6  10000  10000      0 -10000  10000
+
+# AUX3 channel (select RC channel with RC_MAP_AUX3 param)
+M: 1
+S: 3 7  10000  10000      0 -10000  10000
+
+# FLAPS channel (select RC channel with RC_MAP_FLAPS param)
+M: 1
+S: 3 4  10000  10000      0 -10000  10000
+```
+
+> **Note** The file defines a [simple mixer](../concept/mixing.md#simple-mixer) with four outputs.
+> - `M: 1` indicates an output that is defined by one control input (the following `S` line).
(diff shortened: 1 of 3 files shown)


en/concept/mixing.md
```diff
@@ -74,16 +76,16 @@ For a multicopter things are a bit different: control 0 (roll) is connected to a
 * 1: RC pitch
 * 2: RC yaw
 * 3: RC throttle
-* 4: RC mode switch
-* 5: RC aux1
-* 6: RC aux2
-* 7: RC aux3
+* 4: RC mode switch (Passthrough of RC channel mapped by [RC_MAP_FLAPS](../advanced/parameter_reference.md#RC_MAP_FLAPS))
```
hamishwillee, line 79: FYI this makes it clear what this actually maps to!

en/concept/mixing.md
```diff
@@ -74,16 +76,16 @@ For a multicopter things are a bit different: control 0 (roll) is connected to a
...
+* 7: RC aux3 (Passthrough of RC channel mapped by [RC_MAP_AUX3](../advanced/parameter_reference.md#RC_MAP_AUX3))
 
 > **Note** This group is only used to define mapping of RC inputs to specific outputs during *normal operation* (see [quad_x.main.mix](https://github.com/PX4/Firmware/blob/master/ROMFS/px4fmu_common/mixers/quad_x.main.mix#L7) for an example of AUX2 being scaled in a mixer).
   In the event of manual IO failsafe override (if the PX4FMU stops communicating with the PX4IO board) only the mapping/mixing defined by control group 0 inputs for roll, pitch, yaw and throttle are used (other mappings are ignored).
 
 
-### Control Group #6 (First Payload)
+### Control Group #6 (First Payload) {#control_group_6}
 
 * 0: function 0 (default: parachute)
```
hamishwillee: @bkueng One big question that is not answered is what values can you use?

My assumption would be that the payload group is for whatever custom stuff you want. If we had a defined payload group we wouldn't want people to mess with it and overload its use. But then I see default is marked as "parachute" so I am wondering why this isn't called "parachute" to reserve it?

Can you clarify?

en/concept/mixing.md
```diff
@@ -161,6 +163,19 @@ A mixer file must be named **XXXX._main_.mix** if it is responsible for the mixi
 > **Note** `XXXX` in the mixer file name is set in the airframe definition (e.g. see `set MIXER quad_w` in [airframes/10015_tbs_discovery](https://github.com/PX4/Firmware/blob/master/ROMFS/px4fmu_common/init.d/airframes/10015_tbs_discovery))
 
 
+### Loading a Custom Mixer {#loading_custom_mixer}
```
hamishwillee, line 184: FYI, I thought that this entry point is probably the one thing people will definitely read, so this should be part of it. We could however have as its own doc if you think that is better.

IN that case I'd probably still  have this section - but renamed "Mixer File Loading"

en/concept/custom_mixer_payload.md
```diff
@@ -0,0 +1,77 @@
...
+# FLAPS channel (select RC channel with RC_MAP_FLAPS param)
+M: 1
+S: 3 4  10000  10000      0 -10000  10000
+```
+
+> **Note** The file defines a [simple mixer](../concept/mixing.md#simple-mixer) with four outputs.
+> - `M: 1` indicates an output that is defined by one control input (the following `S` line).
+> - `S: 3 `_`n`_ indicates that the input is the n<>th<> input of [Control Group 3 (Manual Passthrough)](../concept/mixing.md#control-group-3-manual-passthrough).
+>  So for `S: 3 5` the input is called "RC aux1", and will output the value of the channel mapped by parameter `RC_MAP_AUX1`.
+> - The order of output declaration of
```
hamishwillee: Note to me. Delete this line

en/concept/custom_mixer_payload.md
```diff
@@ -0,0 +1,77 @@
+# Custom Payload Mixers
```
hamishwillee, line 1: So I've tried not to explain too much about mixers here. That should be covered elsewhere ... properly. 
The model is to show how a simple mixer and payloads work in practise, starting from the pas.aux.mix example.

en/concept/custom_mixer_payload.md
```diff
@@ -0,0 +1,77 @@
...
+  S: 6 1  10000  10000      0 -10000  10000
+  ```
+
+Because this output is in the first position in the file it will map to the first AUX PWM output (unless UAVCAN is enabled).
+This output will now respect updates to the payload control group (6) output 1.
+
+Putting an output on group 6 works by publishing actuator control group 6. First you have to create the publication.
+This should happen once when the PX4 module is initialized (look for places where this pattern is already being used):
+```
+uORB::Publication<actuator_controls_s> _actuator_controls_pub{ORB_ID(actuator_controls_0)};
```
hamishwillee: @bkueng This confuses me. How does the system know I'm intersted in group 5? Is the last param a typo - ie should it be `actuator_controls_6`?

en/concept/custom_mixer_payload.md
```diff
@@ -0,0 +1,77 @@
...
+```
+
+Then you need to publish the first message:
+```
+actuator_controls_s _act_controls{};
+_act_controls.timestamp = hrt_absolute_time();
+// set the first output to 50% positive (this would rotate a servo halfway into one of its directions)
+_act_controls.control[0] = 0.5f;
```
hamishwillee, line 96: And further to the comment above - https://github.com/PX4/Devguide/pull/973/files#r392069246 is this a typo too? I still don't know how we know we have the right control group, but again would have thought that the index would be 1 not 0.

en/concept/custom_mixer_payload.md
```diff
@@ -0,0 +1,77 @@
...
+```
+
+> **Note** The file defines a [simple mixer](../concept/mixing.md#simple-mixer) with four outputs.
+> - `M: 1` indicates an output that is defined by one control input (the following `S` line).
+> - `S: 3 `_`n`_ indicates that the input is the n<>th<> input of [Control Group 3 (Manual Passthrough)](../concept/mixing.md#control-group-3-manual-passthrough).
+>  So for `S: 3 5` the input is called "RC aux1", and will output the value of the channel mapped by parameter `RC_MAP_AUX1`.
+> - The order of output declaration of 
+
+
+Start by copying the mixer file and putting it onto the SD Card at: **/etc/mixers/pass.aux.mix** (see [Mixing and Actuators > Loading a Custom Mixer](../concept/mixing.md#loading_custom_mixer).
```
bkueng, line 45: Worth noting that `pass.aux.mix` depends on the airframe. `pass` is typically the default, but in particular VTOL's typically have their own.
This can be checked in the airframe config: https://github.com/PX4/Firmware/blob/master/ROMFS/px4fmu_common/init.d/airframes/13006_vtol_standard_delta#L58.
I now see you explained this further down though.

en/concept/custom_mixer_payload.md
bkueng: You're right, it should be.
But 6 is not even implemented: https://github.com/PX4/Firmware/blob/master/msg/actuator_controls.msg#L17.

en/concept/custom_mixer_payload.md
bkueng: Looks like both are 0-based indexes, so you are right, the index should match with the mixer definition.

en/concept/mixing.md
bkueng: :+1:

en/concept/mixing.md
bkueng: Yes parachute is simply incorrect, and should be removed. As you also mention, only one module is allowed to publish a certain control_group topic at a given time, otherwise you will run into problems.

en/concept/mixing.md
```diff
@@ -161,6 +163,19 @@ A mixer file must be named **XXXX._main_.mix** if it is responsible for the mixi
...
+
+PX4 loads appropriately named mixer files from the SD card directory **etc/mixers**, by preference, and then the version in Firmware.
+
+To load a custom mixer, you should give it the same name as a "normal" mixer file (that is going to be loaded by your airframe) and put it in the **etc/mixers** directory on your flight controller's SD card.
+
+Most commonly you will use the mixer file for your current airframe (**XXXX._aux_.mix**), where the `XXXX` prefix is the set in the airframe configuration file (e.g. [airframes/10015_tbs_discovery](https://github.com/PX4/Firmware/blob/master/ROMFS/px4fmu_common/init.d/airframes/10015_tbs_discovery) loads **quad_w._aux_.mix** - see `set MIXER quad_w`).
```
bkueng: Might be worth noting that you can override the main and/or aux mixer

en/concept/mixing.md
```diff
@@ -161,6 +163,19 @@ A mixer file must be named **XXXX._main_.mix** if it is responsible for the mixi
...
+
+**Note**
+- The airframe mixer file is preferred because it is always loaded (it is not always obvious which other mixer files are loaded, if any).
```
bkueng: The airframe mixer as opposed to what? This is not clear to me.

en/concept/mixing.md
bkueng: Yes, this is good. Answers some comments I had above.

bkueng (commented):
Thanks @hamishwillee.
Personally I think if we implemented the/a MAVLink servo command onto the RC AUXn channels (so that both can be used, MAVLink or RC), it would solve most of the use-cases and people would not have to touch mixers or even code.
For anyone who wants to go beyond, this is good though.

en/concept/custom_mixer_payload.md
hamishwillee: Actually I didn't completely explain this because it isn't clear what the default is in all cases. Based on comments and what I see in the airfames reference I think it looks like this:

Is this right
1. Set MAIN mixer : `set MIXER xxxx`.
1. Explicitly set AUX MIXER: `set MIXER_AUX yyyyy`
1. If no MIXER_AUX explicitly set:
   - **FOR copter, FW, etc. (not VTOL)**, then `set MIXER` name is used as prefix for expected AUX file.
   - **FOR VTOL airframe** then `pass.aux.mix` used for AUX.

Is that right?

en/concept/custom_mixer_payload.md
hamishwillee: @bkueng So I think you're saying that there is no way to publish to this topic - ie this guide is pointless? Correct?

What do I do about that then? Can you publish the topic, or do we go back to Lorenz or what?

en/concept/custom_mixer_payload.md
hamishwillee: @bkueng But (even) if I change this to `_act_controls.control[1] = 0.5f;` and we publish the topic for group 6 (as discussed above https://github.com/PX4/Devguide/pull/973/files#r392069246) how does the system know that I want to publish to group 6? ie.. `actuator_controls_s _act_controls{};` would be *all groups*, and we would want to publish to group 6, index 1.

Does my question make sense?

en/concept/mixing.md
hamishwillee: I've clarified that you'll most commonly use the airframe *AUX* file, and added the comment about main in first note below. The point being that you can overload anything you like, but AUX in your airframe is a a good target.

en/concept/mixing.md
hamishwillee: As opposed to _any other mixer_ (e.g. pass.aux.mix). The problem here is that there are a bunch of other mixers and it isn't clear which of them might be loaded. The only one you *know* will be loaded is airframe. The stuff in brackets was supposed to make that clear. 

I'll revisit this after you answer https://github.com/PX4/Devguide/pull/973/files#r392770530 - might just remove the line.

en/concept/custom_mixer_payload.md
bkueng: I think it's simple enough that we can just add the additional required steps:
- define the topic in https://github.com/PX4/Firmware/blob/master/msg/actuator_controls.msg#L17
- `uint8 NUM_ACTUATOR_CONTROL_GROUPS = 5`
- subscribe the topic in https://github.com/PX4/Firmware/blob/master/src/lib/mixer_module/mixer_module.cpp#L52

That would make it control group #4 though. And maybe some more bits are required.

en/concept/custom_mixer_payload.md
bkueng: Almost, for AUX it is:
- fw+mc use pass by default
- vtol uses none by default
- if MIXER_AUX is none (so vtol+airframe does not set it), then `set MIXER` name is used as prefix for expected AUX file.
- there's also some logic for gimbals: https://github.com/PX4/Firmware/blob/master/ROMFS/px4fmu_common/init.d/rc.interface#L25

en/concept/custom_mixer_payload.md
bkueng: The topic name defines the group: `uORB::Publication<actuator_controls_s> _actuator_controls_pub{ORB_ID(actuator_controls_0)};` is group #0, `uORB::Publication<actuator_controls_s> _actuator_controls_pub{ORB_ID(actuator_controls_1)};` is group #1. Does this make sense?

en/concept/mixing.md
bkueng: I see. You're right, the airframe mixer is the one that's easier to know which one is actually used.

hamishwillee:
> Personally I think if we implemented the/a MAVLink servo command onto the RC AUXn channels (so that both can be used, MAVLink or RC), it would solve most of the use-cases and people would not have to touch mixers or even code.
For anyone who wants to go beyond, this is good though.

Yes. Awesome idea. When can you do it ? :-)

coderkalyan:
Hey guys, just realized this thread is a thing! How can I help? How much of it relates to what we were discussing about the DO_SET_SERVO command @hamishwillee @bkueng ?

en/concept/custom_mixer_payload.md
hamishwillee: Awesome, thanks. I've re-written this section to split out mixer loading and custom mixer loading. I think very clear now what gets loaded when. Please recheck.

en/concept/custom_mixer_payload.md
hamishwillee: I think my C++ is a bit rusty. I can see 
```
#publish to group 6
uORB::Publication<actuator_controls_s> _actuator_controls_pub{ORB_ID(actuator_controls_6)};
```
But I can't see how you know that this now means `_act_controls` now points to the same set of controls (actuator_controls_6)
```
actuator_controls_s _act_controls{};
```
Anyway, can I ask YOU to fix up this whole section ie after "Putting an output on group 6 works" - and showing how to publish (https://github.com/PX4/Devguide/pull/973/files#r392899091 )

Too much back and forth otherwise. (i.e. I don't know how to fix address "That would make it control group #4 though. And maybe some more bits are required.", and I don't want to).

hamishwillee:
> Hey guys, just realized this thread is a thing! How can I help? How much of it relates to what we were discussing about the DO_SET_SERVO command @hamishwillee @bkueng ?

@coderkalyan All and none of it :-). Beat is suggesting this fix for do_set_servo: https://github.com/PX4/Devguide/pull/973#issuecomment-599814861
This topic is attempting to capture Lorenz document usefully. But if we fixed up 973 most people would not need most of this.

en/concept/custom_mixer_payload.md
bkueng: `_act_controls` is a generic data structure. It can be published on any control group topic.
I pushed a commit to add it.

bkueng:
@coderkalyan You're welcome to help, you can take https://github.com/PX4/Firmware/pull/10320#issuecomment-561119536 as a starting point for `DO_SET_SERVO`.

en/concept/custom_mixer_payload.md
hamishwillee: Thanks!

hamishwillee: Thanks. I'll just have to trust this works.

hamishwillee:
@bkueng Thanks for all your help on this. I'm assuming you're otherwise happy and have merged. I think the mixer loading section alone makes this work worthwhile.

coderkalyan:
@bkueng Do you want me to open a separate PR or something? What's up with the current PR on make do set servo work - seems like a really long discussion!

PR #978, opened 2020-03-17
Mavlink: Point to startup script locations

hamishwillee:
@julianoes As discussed on slack with @RicardoM17 this makes it more clear what startup file you should add mavlink startup to - pointing to startup scripts doc and some example files you might update.

I figure rcs file is right place, but actually you could put it in any of the linked configs including the airframe right?

Anything else to add.

### DIFF
--- a/en/middleware/mavlink.md
+++ b/en/middleware/mavlink.md
@@ -118,9 +118,7 @@ nullptr
 };
 ```
 
-Then make sure to enable the stream, for example by adding the following line to
-the startup script (`-r` configures the streaming rate, `-u` identifies the
-MAVLink channel on UDP port 14556):
+Then make sure to enable the stream, for example by adding the following line to the [startup script](../concept/system_startup.md) (e.g. [/ROMFS/px4fmu_common/init.d-posix/rcS](https://github.com/PX4/Firmware/blob/master/ROMFS/px4fmu_common/init.d-posix/rcS) on NuttX or [ROMFS/px4fmu_common/init.d-posix/rcS](https://github.com/PX4/Firmware/blob/master/ROMFS/px4fmu_common/init.d-posix/rcS)) on SITL. Note that `-r` configures the streaming rate and `-u` identifies the MAVLink channel on UDP port 14556).
 
 ```
 mavlink stream -r 50 -s CA_TRAJECTORY -u 14556


en/middleware/mavlink.md
```diff
@@ -118,9 +118,7 @@ nullptr
 };
 ```
 
-Then make sure to enable the stream, for example by adding the following line to
-the startup script (`-r` configures the streaming rate, `-u` identifies the
-MAVLink channel on UDP port 14556):
+Then make sure to enable the stream, for example by adding the following line to the [startup script](../concept/system_startup.md) (e.g. [/ROMFS/px4fmu_common/init.d-posix/rcS](https://github.com/PX4/Firmware/blob/master/ROMFS/px4fmu_common/init.d-posix/rcS) on NuttX or [ROMFS/px4fmu_common/init.d-posix/rcS](https://github.com/PX4/Firmware/blob/master/ROMFS/px4fmu_common/init.d-posix/rcS)) on SITL. Note that `-r` configures the streaming rate and `-u` identifies the MAVLink channel on UDP port 14556).
```
RicardoM17, line 121: The link for NuttX is wrong (same as the one for SITL). I believe you meant to use this instead? https://github.com/PX4/Firmware/blob/master/ROMFS/px4fmu_common/init.d/rcS . Either way for this one it would be good to indicate the suggested line like so https://github.com/PX4/Firmware/blob/5e343b104ebee28e0d1ce0597e774a248677cb26/ROMFS/px4fmu_common/init.d/rcS#L423 (Julian's suggestion)

RicardoM17 (changes requested):

en/middleware/mavlink.md
hamishwillee: Thanks - fixed the nuttx link.
I don't use specific lines because it a) AFAIK it doesn't really matter where you put the line b) the lines move as the files get modified. So the link might work now, but point to the wrong place in a few months. I'd rather not set a bad expectation.

RicardoM17: That makes sense. It's just because for SITL it's quite obvious where to place but not really for the nuttx equivalent. 

Either way all good with me and many thanks for making this easier for the not so experienced folks like me :D

julianoes (approved):
Thx

en/middleware/mavlink.md
hamishwillee: Well, thanks for not just ignoring this stuff and helping us validate fixes. Lots of people just silently accept the help and move on.

RicardoM17: No problem @hamishwillee I'm glad to help and I hope to do so more and more in the future. Thank you for all the help as well. :slightly_smiling_face:

PR #980, opened 2020-03-18
SD card layout: change to recommendation to format

hamishwillee:
As discussed in https://github.com/PX4/Devguide/pull/972#discussion_r393535675, Nuttx will format the SD card if mounting fails. However generally safer just to tell them to do this themselves.

### DIFF
--- a/en/concept/sd_card_layout.md
+++ b/en/concept/sd_card_layout.md
@@ -2,7 +2,7 @@
 
 The PX4 SD Card is used for storing configuration files, flight logs, mission information etc. 
 
-> **Tip** We recommend that you format the SD Card as FAT32 before use.
+> **Tip** We recommend that you format the SD Card as FAT32 before first use with PX4.
 
 The directory structure/layout is shown below.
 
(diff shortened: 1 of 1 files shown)


bkueng:
But this now tells everyone to format, even for brand new cards, which is not required.

hamishwillee:
Hmmm. It's a recommendation, not an instruction. But I get your point.

We could change from

> We recommend that you format the SD Card as FAT32 before first use with PX4.

To

> The SD card should be FAT32 formatted for use with PX4. 

or 

> The SD card should be FAT32 formatted for use with PX4. We recommend you reformat cards that are using a different file system.

If not, what do you suggest?

bkueng:
This:
> The SD card should be FAT32 formatted for use with PX4 (which is the default for SD cards). We recommend you reformat cards that are using a different file system.

?

hamishwillee:
sure: https://github.com/PX4/Devguide/commit/6d602fc02989bc23b04a929e53a5108807ddfb54

PR #976, opened 2020-03-17
Drivers: Add log verbosity info

hamishwillee:
Adds info on how to increase the log verbosity when driver debugging. 

This info should apply more generally to modules (and the example I use is a module template). However it is untested (by me) so applying here where suggested by @davids5 .

PR #983, opened 2020-03-19
Drivers: adding define to force debug build

hamishwillee:
Suggestion from here: https://github.com/PX4/Devguide/pull/976#discussion_r394139986

@bkueng Are there pros/cons to each method? I.e. why/when would you do this instead of updating the cmake info? My assumption is that in either case this is a temporary testing change, so perhaps it doesn't matter?

### DIFF
--- a/en/middleware/drivers.md
+++ b/en/middleware/drivers.md
@@ -143,4 +143,4 @@ px4_add_module(
 	)
 ```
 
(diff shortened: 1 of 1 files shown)


bkueng (approved):
Both are similar, but if you already have the .cpp file open, it's quicker, and adding it on the CMake level enables more output (which you might not want) from all files in the module.

hamishwillee:
Thanks @bkueng - I added "on a per-file basis" which is the most important point (ie not getting this for the whole module as you pointed out)

PR #986, opened 2020-03-19
param docs: do clean build after adding new parameter file

hamishwillee:
As discussed on slack.

Note, this does not include the other comments around use of .get() and .update() as it isn't clear when you would use them. That discussion continued separately on slack.

### DIFF
--- a/en/advanced/parameters_and_configurations.md
+++ b/en/advanced/parameters_and_configurations.md
@@ -224,6 +224,9 @@ Typically it is stored alongside its associated module.
 
 The build system extracts the metadata (using `make parameters_metadata`) to build the [parameter reference](../advanced/parameter_reference.md) and the parameter information used by ground stations.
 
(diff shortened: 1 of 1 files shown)


en/advanced/parameters_and_configurations.md
```diff
@@ -224,6 +224,9 @@ Typically it is stored alongside its associated module.
 
 The build system extracts the metadata (using `make parameters_metadata`) to build the [parameter reference](../advanced/parameter_reference.md) and the parameter information used by ground stations.
 
+> **Warning** After adding a *new* parameter file you must call `make clean` before building or the new parameters will not be generated (parameter files are added as part of the *cmake* configure step, which only happens for clean builds).
```
bkueng: technically a reconfigure also happens when one of the cmake files change.

bkueng (approved):

en/advanced/parameters_and_configurations.md
hamishwillee: Thanks. I believe this is now clarified. I'm going to merge. If you think this needs further work, please propose a change.

PR #993, opened 2020-03-25
Explain how to define locations in the world file

Jaeyoung-Lim:
This PR explains how to define the location of the world using .world files.

This explains how to use https://github.com/PX4/sitl_gazebo/pull/441


TSC21 (changes requested):
Should it be stated that the LLA can be set by env variables? And if they are set, they take priority over setting coordinates in the `world` file.

en/simulation/gazebo.md
```diff
@@ -258,19 +260,58 @@ The next time you build/restart Gazebo it will use the new GPS noise setting.
...
+## Set World Location {#set_world_location}
+
+By default, a world is loaded at the current vehicle location.
+If using a world that simulates a real location, this can result in a mismatch between what is displayed from the vehicle camera and the what is shown on the ground station map.
+To overcome this problem you can set the location of the world as shown in this section.
+
+> **Note** You can also set a [Custom Takeoff Location](#custom_takeoff_location) that matches the real-world location of the map.
+  However adding the location to the map is easier (and can still be over-ridden if needed).
+
+The location of the world is defined in the **.world** file by specifying the location of the origin using the `spherical_coordinates` tag.
```
hamishwillee, line 228: What is the origin? i.e. it is clear in the general sense how to align the model, but not in the very specifics. I also assume that the orientation of the model will be right?

en/simulation/gazebo.md
```diff
@@ -258,19 +260,58 @@ The next time you build/restart Gazebo it will use the new GPS noise setting.
...
-
 You can also specify the full path to a world to load using the `PX4_SITL_WORLD` environment variable.
 This is useful if testing a new world that is not yet included with PX4.
 
+> **Tip** If the loaded world does not align with the map, you may need to [set the world location](#set_world_location). 
+  
+
+## Set World Location {#set_world_location}
+
+By default, a world is loaded at the current vehicle location.
```
hamishwillee: @Jaeyoung-Lim Is this correct?

My assumption is that what happens is that setting the world causes the world to be loaded in the right place AND sets the vehicle to load at that location. 
So if you set the world AND a different custom takeoff location the world/map are still correctly aligned - just the vehicle may not be near them
(the other possibility is that if you set the different location the vehicle position is set there, and the world loads at that location - which would make less sense).

hamishwillee:
I've re-organised (and made sure that Nuno's comment addressed). Might need further changes based on https://github.com/PX4/Devguide/pull/993#discussion_r398248299

en/simulation/gazebo.md
Jaeyoung-Lim: The origin is (0, 0, 0) of the world - and we can define the gps coordinates(location) of the origin. The orientation of the model doesn't matter, since coordinate in gazebo is ENU(East North Up)

en/simulation/gazebo.md
Jaeyoung-Lim: The vehicle always get spawned at the origin and the world just defines where the origin lies on gps coordinates.

Jaeyoung-Lim:
@hamishwillee Should I move the changes to the new section? or does this live here and more information lives in the new page?

hamishwillee:
> Should I move the changes to the new section? or does this live here and more information lives in the new page?

By "new section" you mean the worlds doc? No. My thinking is that the world doc is at this point just an image repository - interesting, but you don't "need" it. All key info goes in this page.

That will change if we get lots of info that is very specific to worlds. But we're not there yet.

en/simulation/gazebo.md
hamishwillee: @Jaeyoung-Lim I think that you're saying the vehicle always gets spawned at the origin of the world, and by setting the GPS co-ordinates **in the** world you're making that origin be at a certain GPS co-ordinate - right?

And by setting the position "manually" (using the env variables) you're doing the same thing? I.e. you're not saying "the vehicle is at GPS x, y, z. You're saying "set the origin of the world to position X, Y, Z, and the vehicle will appear there?" Is that correct?

en/simulation/gazebo.md
hamishwillee: OK. See my other clarification. I'm pretty sure I got it - just want to be sure: https://github.com/PX4/Devguide/pull/993/files#r399863369

hamishwillee:
Thanks @Jaeyoung-Lim , 
I've clarified "locations in the world file" and merged :-)

PR #996, opened 2020-03-27
Computer vision: Cross link to PX4 Vision Kit

hamishwillee:
@Jaeyoung-Lim @jinger26 This is cross link to the vision kit. I'll also add it to the path-planning doc.
 Anywhere else we should add it in your opinion?

@mrivi @jkflying Should we (you!) add something similar to https://github.com/PX4/avoidance#obstacle-detection-and-avoidance ?


hamishwillee:
Nothing controversial in this. Merging.

PR #1000, opened 2020-03-31
Video Streaming in QGC Odroid - Still Useful?

hamishwillee:
@bkueng This links to/relies on a link to now dead tutorial: https://pixhawk.org/peripherals/onboard_computers/odroid_c1 

The bit it references is setting up Ubuntu 14.04 on Odroid C1
Does this article still have value? i.e. could we reasonably remove it, and if we did, what would we lose?

My feeling is that generally it is useful to understand how to set up a camera to attach to a companion computer and stream to QGC. Not sure whether this particular computer is popular, whether the transmitter is still available, or that this article is up to date enough to be useful. 

Thoughts? If it is useful? If so, would it be recoverable by replacing the links with these (say)
- https://www.hardkernel.com/blog-2/ubuntu-18-04-lts-for-odroid-c1/
- https://wiki.odroid.com/odroid-c1/os_images/ubuntu/minimal

Or is there anyone else you know who might help

### DIFF
--- a/en/SUMMARY.md
+++ b/en/SUMMARY.md
@@ -114,7 +114,7 @@
   * [ULog File Format](log/ulog_file_format.md)
 * [Tutorials](tutorials/tutorials.md)
   * [Ground Control Station](qgc/README.md)
-  * [Video Streaming in QGC](qgc/video_streaming.md)
+  * [Video Streaming from Odroid C1 to QGC](qgc/video_streaming.md)
   * [Long-distance Video Streaming](qgc/video_streaming_wifi_broadcast.md)
   * [S.Bus Driver for Linux](tutorials/linux_sbus.md)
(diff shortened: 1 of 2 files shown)


bkueng:
I think so too, it is generally useful to have a QGC video tutorial.
Odroid's are generally popular, somewhat less than RPi's though.
@Jaeyoung-Lim might have good resources.

hamishwillee:
@bkueng @Jaeyoung-Lim Thanks for your advice. In the short term I have done a quick and dirty fix - getting the links from wayback machine. This is enough that someone can use the instructions as documented. I'm going to merge that for now.

More generally, can we discuss our companion story in https://github.com/PX4/Devguide/issues/1004 . I think it is far more work than it should be to get a companion system working. If we had "something" then things like Pixhawk vision could be done by others.

PR #1001, opened 2020-03-31
Remove Windows Msys Installer Docs

hamishwillee:
The px4driver.msi is now gone from QGC. Instead it uses https://github.com/mavlink/qgroundcontrol/blob/master/deploy/driver.msi?raw=true

1. Do we know whether I can direct replace like this?
2. Is there any value in this toolchain anymore. I like the idea in theory, but what value does it have, if any over cygwin? I'm pretty sure it doesn't work any more, so wondering if we should just remove it from docs? Who can agree that?

Ah, even the link for the installer seems to be gone: http://firmware.diydrones.com/Tools/PX4-tools/px4_toolchain_installer_v14_win.exe


julianoes:
I have no idea and removing myself as a reviewer :smile:.

hamishwillee:
@MaEtUgR @bkueng I want to remove the old msys windows toolchain. As far as I can tell it is not available or working at all. Thoughts?

MaEtUgR:
Yeah, I wasn't sure who uses it and how. I was going to rework it but when I found out I can get NuttX **and SITL** working on Cygwin I completely switched over my efforts. I have no capacity to test or maintain it. The advantage would be it's more native and potentially smaller/faster for building NuttX targets than Cygwin but I don't have any real-world comparison and the value is gone if it doesn't work. So from my side we can remove it.

MaEtUgR:
btw the QGC installer still contains the driver installer it's apparently just a different installer now. But on Windows 10 I never installed any driver (I cancel the installation every time) and all PX4 powered drones always worked out of the box 🤷‍♂ .
image

MaEtUgR:
@hamishwillee I brought it up in the dev call and the conclusion was quite quickly that we should remove the MSYS based toolchain from the docs since no one uses or maintains it.

hamishwillee:
@MaEtUgR Thanks, in particular for chasing this on the DevCall. I'll sort it!

FYI I investigated the drivers a while ago (thanks to Michael Oborne). They are _required_ for Vista, but not for Windows 10. 

On Windows 10 they can be useful if you have multiple connections to multiple vehicles. IN this case they label the COM pots with the name of the connected vehicle - ie not just COM1.

hamishwillee:
This just removes the whole lot. Merging :-)

MaEtUgR:
Cool, thanks @hamishwillee for taking care!

> On Windows 10 they can be useful if you have multiple connections to multiple vehicles.

So maybe I should try it too.

PR #1009, opened 2020-04-07
add work queue example

bkueng:
From https://github.com/PX4/Firmware/pull/14597

@hamishwillee I'm sure you can bring this in a nicer form :)

### DIFF
--- a/en/apps/module_template.md
+++ b/en/apps/module_template.md
@@ -1,9 +1,21 @@
 # Module Template for Full Applications
 
-The PX4 Firmware contains a template for writing a new application (module) that runs as a [task](../concept/architecture.md#runtime-environment) on its own stack:
-[src/templates/module](https://github.com/PX4/Firmware/tree/master/src/templates/module).
+There are 2 different approaches to writing a full application: either running as a task or as a work queue item.
+See [here](../concept/architecture.md#runtime-environment) for the differences.
+In most cases a work queue item can be used, as this minimizes resource usage.
 
 > **Note** All the things learned in the [First Application Tutorial](../apps/hello_sky.md) are relevant for writing a full application.
+
+
+## Work Queue Item
+
+The PX4 Firmware contains a template for writing a new application (module) that runs as a work queue item:
+[src/examples/work_item](https://github.com/PX4/Firmware/tree/master/src/examples/work_item).
+
+
+## Tasks
+The PX4 Firmware contains a template for writing a new application (module) that runs as a task on its own stack:
+[src/templates/module](https://github.com/PX4/Firmware/tree/master/src/templates/module).
 
 The template demonstrates the following additional features/aspects that are required or are useful for a full application:
 
--- a/en/apps/module_template.md
+++ b/en/apps/module_template.md
@@ -1,21 +1,38 @@
 # Module Template for Full Applications
 
-An application can be written to run as either a task (a module with its own stack and process priority) or as a work queue item (multiple modules share the same stack).
-In most cases a work queue item can be used, as this minimizes resource usage.
+An application can be written to run as either a *task* (a module with its own stack and process priority) or as a *work queue task* (a module that runs on a work queue thread, sharing the stack and thread priorit with other tasks on the work queue).
+In most cases a work queue task can be used, as this minimizes resource usage.
 
-> **Note** [Architectural Overview > Runtime Environment](../concept/architecture.md#runtime-environment) provides more information about stacks and work queues.
+> **Note** [Architectural Overview > Runtime Environment](../concept/architecture.md#runtime-environment) provides more information about tasks and work queue tasks.
 
 <span></span>
 > **Note** All the things learned in the [First Application Tutorial](../apps/hello_sky.md) are relevant for writing a full application.
 
 
-## Work Queue Item
+## Work Queue Task
 
-The PX4 Firmware contains a template for writing a new application (module) that runs as a work queue item: 
+The PX4 Firmware contains a template for writing a new application (module) that runs as a work queue task: 
 [src/examples/work_item](https://github.com/PX4/Firmware/tree/master/src/examples/work_item).
+
+A work queue task application is just the same as an ordinary (task) application, except that it needs to specify that it is a work queue task, and schedule itself to run during initialisation.
+
+The example shows how.
+In summary:
+1. Specify the dependency on the work queue library in the cmake definition file ([CMakeLists.txt](https://github.com/PX4/Firmware/blob/master/src/examples/work_item/CMakeLists.txt)):
+   ```
+   ...
+   DEPENDS
+      px4_work_queue
+```
+1. In addition to `ModuleBase`, the task should also derive from `ScheduledWorkItem` (included from [ScheduledWorkItem.hpp]( https://github.com/PX4/Firmware/blob/master/platforms/common/include/px4_platform_common/px4_work_queue/ScheduledWorkItem.hpp))
+1. Implement the `ScheduledWorkItem::Run()` method to perform "work"
+1. Implement the `task_spawn` method, specifying that the task is a work queue (using the `task_id_is_work_queue` id.
+1. Schedule the work queue task using one of the scheduling methods (in the example we use `ScheduleOnInterval` from within the `init` method).
+
 
 
 ## Tasks
+
 The PX4 Firmware contains a template for writing a new application (module) that runs as a task on its own stack:
 [src/templates/module](https://github.com/PX4/Firmware/tree/master/src/templates/module).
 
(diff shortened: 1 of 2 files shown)


en/concept/architecture.md
```diff
@@ -127,20 +127,21 @@ The whole PX4 middleware runs in a single address space, i.e. memory is shared b
 > to be changed include `uORB`, `parameter interface`, `dataman` and `perf`).
 
 There are 2 different ways that a module can be executed:
-- **Tasks**: The module runs in its own task with its own stack and process priority
-  (this is the more common way). 
+- **Tasks**: The module runs in its own task with its own stack and process priority.
 - **Work queues**: The module runs on a shared task, meaning that it does not own a stack. 
-  Multiple tasks run on the same stack with a single priority per work queue.
+  Multiple tasks can run on the same stack with a single priority per work queue.
```
dagar:
```suggestion
  Multiple modules can run on the same stack with a single priority per work queue.
```

en/concept/architecture.md
```diff
@@ -113,34 +113,27 @@ considerably slower.
...
+> **Info** The system is designed such that with minimal effort it would be possible to run each module in separate address space (parts that would need to be changed include `uORB`, `parameter interface`, `dataman` and `perf`).
 
 There are 2 different ways that a module can be executed:
-- **Tasks**: The module runs in its own task with its own stack and process priority
-  (this is the more common way). 
-- **Work queues**: The module runs on a shared task, meaning that it does not own a stack. 
-  Multiple tasks run on the same stack with a single priority per work queue.
+- **Tasks**: The module runs in its own task with its own stack and process priority.
+- **Work queues**: The module runs on a shared task, meaning that it does not own a stack.
+  Multiple modules can run on the same stack with a single priority per work queue.
```
hamishwillee: YOu say "same stack with a single priority per work queue." - do you mean "all work queues share a stack and priority?"

en/concept/architecture.md
```diff
@@ -113,34 +113,27 @@ considerably slower.
...
 
-  A task is scheduled by specifying a fixed time in the future.
-  The advantage is that it uses less RAM, but the task is not allowed to sleep
-  or poll on a message.
+  A task is scheduled by specifying a fixed time in the future, or via uORB topic update callback.
```
hamishwillee: 1. This is indented, which means that it looks like it belongs with the Work queue information. Is that the case?
2. It says "A task is scheduled ..." - do you mean a task or a work queue?

So I "think" you probably mean (Please confirm):
- Task 
  - runs in its own stack and priority.
  - It can sleep or poll on a message and do blocking IO.
- Work queue:
  - shares a stack.
  - shares a priority with other work queues
  - cannot  sleep or poll on a message or do blocking IO.
- long running tasks should run in own task or own work queue - is this because queues are cooperative? or if not, why?

Further questions:
- Above implies you can have multiple work queues? If so, how do you choose which one? How many items should there be on a queue?
- Are work queues cooperative? i.e. do they have to cede control to other queues? If so we should explain that.

en/apps/module_template.md
```diff
@@ -1,10 +1,24 @@
...
 
+<span></span>
 > **Note** All the things learned in the [First Application Tutorial](../apps/hello_sky.md) are relevant for writing a full application.
 
+
+## Work Queue Item
+
+The PX4 Firmware contains a template for writing a new application (module) that runs as a work queue item: 
+[src/examples/work_item](https://github.com/PX4/Firmware/tree/master/src/examples/work_item).
+
```
hamishwillee, line 16: It is not clear to me what it is that you do that makes the module behave as a work queue or as a task. We can save users a lot of time by spelling this out.

E.g. Just enough for people not to have to guess what is relevant.
1. In Tasks section we should explain what makes a module - derives from X, Y, Z. uses CMAKE X, Y, Z.
2. Here explain the delta - again , implements interface X, does Y. 

By inspection I THINK that for a work queue item;
1. CMAKE
   ```
	DEPENDS
		px4_work_queue
   ```
1. Includes https://github.com/PX4/Firmware/blob/master/platforms/common/include/px4_platform_common/px4_work_queue/ScheduledWorkItem.hpp
2. Derives from `ScheduledWorkItem`
3. Implements the `ScheduledWorkItem::Run()` method and sets the scheduling rate

It also looks like you might have to do something in the ModuleBase::taskSpawn to actually create it on the work queue?

Can you clarify the essentials and essential differences?

hamishwillee:
Thanks @bkueng - Perhaps, but not until some of the things I'm finding confusing are cleared up :-0

en/concept/architecture.md
bkueng: No, there are N work queues, each of which has a priority + stack, and on each work queue there can be an arbitrary number of tasks/modules/work queue items running (a task needs to be assigned to a specific work queue).

en/concept/architecture.md
bkueng: - I'd change `Work queue: ` -> `Work queue task: `, as mentioned above, the work queue itself specifies essentially the thread, on which the tasks/items/modules are running. Not sure what the best term is.

> long running tasks should run in own task or own work queue - is this because queues are cooperative? or if not, why?

Yes, and everything is serialized, so a work queue item cannot be interrupted by another item on the same work queue.

> Above implies you can have multiple work queues? If so, how do you choose which one?

Correct, you specify on initialization: https://github.com/PX4/Firmware/blob/master/src/modules/load_mon/load_mon.cpp#L121
Here are all of them: https://github.com/PX4/Firmware/blob/master/platforms/common/include/px4_platform_common/px4_work_queue/WorkQueueManager.hpp#L49.

> How many items should there be on a queue?

As many as you want - at some point you'll notice negative effects, but we're not there yet.

> Are work queues cooperative? i.e. do they have to cede control to other queues? If so we should explain that.

A work queue is essentially a task/thread: it can be interrupted by higher-priority tasks.

en/apps/module_template.md
bkueng: Yes this is all correct. In spawning the tasks you say it's a work queue item: https://github.com/PX4/Firmware/blob/master/src/examples/work_item/WorkItemExample.cpp#L100
Then it schedules itself in `init`: https://github.com/PX4/Firmware/blob/master/src/examples/work_item/WorkItemExample.cpp#L54.

en/concept/architecture.md
hamishwillee: Thanks @bkueng 

I couldn't see an answer for this bit below. Can you check it is correct? Note modified in bits to be more clear

1. This is indented, which means that it looks like it belongs with the Work queue information. Is that the case?
2. It says "A task is scheduled ..." - do you mean a task or a work queue?

So I "think" you probably mean (Please confirm):
- Task 
  - runs in its own stack and priority.
  - It can sleep or poll on a message and do blocking IO.
- Work queue task:
  - shares a stack with other work queue tasks.
  - shares a priority with other work queue tasks on the queue
  - cannot  sleep or poll on a message or do blocking IO. Basically it gets its turn eventually (I assume it is round robin?)

bkueng: > This is indented, which means that it looks like it belongs with the Work queue information. Is that the case?

Correct

> It says "A task is scheduled ..." - do you mean a task or a work queue?

This is where we need clarity on the naming. It's the work queue task, but not the work queue thread itself.

> So I "think" you probably mean (Please confirm):

Confirmed.

> Basically it gets its turn eventually (I assume it is round robin?)

Yes added to a linked list. We might add priorities in future.

hamishwillee:
@bkueng I have re-written this in line with our discussion - please re-review.

It is now clear to me how you identify that this is a work queue task - using the id `task_id_is_work_queue`, and how it is scheduled. I don't understand the detail of the code (i.e. what the object.store is doing) but the code could be copied easily enough to capture the pattern:

```
int WorkItemExample::task_spawn(int argc, char *argv[])
{
	WorkItemExample *instance = new WorkItemExample();

	if (instance) {
		_object.store(instance);
		_task_id = task_id_is_work_queue;

		if (instance->init()) {
			return PX4_OK;
		}

	} else {
		PX4_ERR("alloc failed");
	}

	delete instance;
	_object.store(nullptr);
	_task_id = -1;

	return PX4_ERROR;
}
```

What isn't clear is what queue this will be added to? I.e. how to I create a new queue and add it to that vs, how to I find out what other queues I might add to?

hamishwillee:
@bkueng When you get a chance, see above ^^^ - how do I find out queues available and choose one? I can't infer from the example.

bkueng:
Didn't I answer that above?
You specify on initialization: https://github.com/PX4/Firmware/blob/master/src/modules/load_mon/load_mon.cpp#L121
Here are all of them: https://github.com/PX4/Firmware/blob/master/platforms/common/include/px4_platform_common/px4_work_queue/WorkQueueManager.hpp#L49.

hamishwillee:
Thanks @bkueng - the hint at where the options were was what I needed to understand this. I've updated and I think this is good to merge now. Sorry for delay in getting to this.

PR #1011, opened 2020-04-07
Forward Fast-RTPS install to setup/fast-rtps-installation.md

TSC21:
This is generating confusion (https://discuss.px4.io/t/stuck-on-ros2-rtps-interface-install/16029/13) and needs to be fixed.


hamishwillee:
Thanks @TSC21 - completely agree, and we shouldn't duplicate the instructions. Ideally we should include as much of this as possible in our standard setup scripts/environment and document inline there.

PR #1013, opened 2020-04-08
cygwin: add file monitors slowing down operation

MaEtUgR:
@bys1123 reminded me friendly that this might be a significant issue for certain people and while I myself optimized my Windows setup for speed we should mention the correlation in the documentation.

### DIFF
--- a/en/setup/dev_env_windows_cygwin.md
+++ b/en/setup/dev_env_windows_cygwin.md
@@ -64,7 +64,9 @@ The ordinary workflow consists of starting a console window by double clicking o
 
 ### File Monitoring Tools vs Toolchain Speed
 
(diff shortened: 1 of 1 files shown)


bys1123:
Maybe "slow down the toolchain operation `and installation`" better? 
I install this toolchain need 20 minutes with anti-virus tool open.

MaEtUgR:
True, I added your suggestion. It can be squashed.

bys1123 (approved):

hamishwillee:
@MaEtUgR @MaEtUgR Great. I've simplified the text to my taste and merged.
- I think that "at own risk" is sufficient disclaimer
- The only "lost" information is mention of Windows defender and that the posix tools are file access hungry. I think the first is not strictly necessary, and the second is clearly inferred :-)

PR #1025, opened 2020-04-24
Update git_example.md with a build error pitfall

francelico:
Provides solution due to a build error caused by not having the updated list of tags in your forked repo.

### DIFF
--- a/en/contribute/git_examples.md
+++ b/en/contribute/git_examples.md
@@ -138,19 +138,26 @@ git checkout <branch name>
 ## Common pitfalls
 
 ### Force push to forked repository
+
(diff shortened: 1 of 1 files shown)


hamishwillee:
@francelico - I've seen that before myself and not added this much needed note, so thank you very much. Note I modified slightly to just fetch upstream tags, as that is just a little more lightweight for usres.

francelico:
Hi @hamishwillee , thanks for the review and fast merging! It's my first time contributing to the documentation, I'm just wonderign when does the https://dev.px4.io/master/en/contribute/git_examples.html page updates to the version on this repo?

hamishwillee:
@francelico Yes it should update. There is some problem with the build. Checking.

francelico:
Yes looks like it built now! I will be deleting my branch then, thanks.

PR #1036, opened 2020-04-30
Simulation: Add info on simulating failur

hamishwillee:
It is a crime that this has not been documented before now :-). 

Julian, I think important enough to also go in jmavsim and gazebo docs so it is not missed right?


julianoes (approved):
I did not know this existed 😲

hamishwillee:
Thanks @julianoes . Merged!

PR #1045, opened 2020-06-09
scripts: ubuntu_sim_ros_melodic: install python-rosdep

TSC21:
Addresses https://github.com/PX4/Firmware/issues/14984 and https://github.com/PX4/Firmware/issues/15064.


hamishwillee:
Did not fully test it, but makes sense :-)

PR #1046, opened 2020-06-11
Update fast-rtps-installation.md

AndyBlightLeeds:
Use the packaged gradlew command instead of using the installed version.

### DIFF
--- a/en/setup/fast-rtps-installation.md
+++ b/en/setup/fast-rtps-installation.md
@@ -84,11 +84,8 @@ Then install Fast-RTPS-Gen 1.0.4 (Gradle is required for this):
 git clone --recursive https://github.com/eProsima/Fast-RTPS-Gen.git -b v1.0.4 ~/Fast-RTPS-Gen \
     && cd ~/Fast-RTPS-Gen \
     && ./gradlew assemble \
-    && ./gradlew install
+    && sudo ./gradlew install
 ```
-
-> **Note** You might require `sudo` permissions on the "install" step
-
 
 ## Installation from Binaries
 
--- a/en/setup/fast-rtps-installation.md
+++ b/en/setup/fast-rtps-installation.md
@@ -83,8 +83,8 @@ Then install Fast-RTPS-Gen 1.0.4 (Gradle is required for this):
 ```
 git clone --recursive https://github.com/eProsima/Fast-RTPS-Gen.git -b v1.0.4 ~/Fast-RTPS-Gen \
     && cd ~/Fast-RTPS-Gen \
-    && gradle assemble \
-    && gradle install
+    && ./gradlew assemble \
+    && ./gradlew install
 ```
 
 > **Note** You might require `sudo` permissions on the "install" step


TSC21 (changes requested):
I don't see a real good reason to do this besides not using the system gradle. Can you elaborate?

AndyBlightLeeds:
This fix was mentioned on the issue <https://github.com/eProsima/Fast-RTPS-Gen/issues/31> as I had exactly the same problem and am running the same version of gradle (4.4.1) as reported by @Alex031544.

@LuisGP said:
> Fast-RTPS-Gen comes with an embedded gradle to avoid version issues, to use it, try executing from the Fast-RTPS-Gen directory:
>
>./gradlew assemble

So this proposed change to the documentation just adds the information that Luis said we should use.

TSC21:
> This fix was mentioned on the issue [eProsima/Fast-RTPS-Gen#31](https://github.com/eProsima/Fast-RTPS-Gen/issues/31) as I had exactly the same problem and am running the same version of gradle (4.4.1) as reported by @Alex031544.
> 
> @LuisGP said:
> 
> > Fast-RTPS-Gen comes with an embedded gradle to avoid version issues, to use it, try executing from the Fast-RTPS-Gen directory:
> > ./gradlew assemble
> 
> So this proposed change to the documentation just adds the information that Luis said we should use.

This shouldn't be required if local gradle is accepted with sudo. For example, in the containers we use, we do not need to use gradlew. I would assume this as an alternative, rather than a replacement.

AndyBlightLeeds:
My guess is that the containers you use have the latest versions of gradle installed so the problem doesn't arise.  Using the packaged version of gradlew will **always** work as the version used is under the control of the development team.  This is why I propose this change.

AndyBlightLeeds:
I have done some more checking about gradle versions.  The dockerfile <https://hub.docker.com/r/px4io/px4-dev-base-bionic/dockerfile> has the following:

```bash
# Gradle (Required to build Fast-RTPS-Gen)
RUN wget -q "https://services.gradle.org/distributions/gradle-5.6.2-bin.zip" -O /tmp/gradle-5.6.2-bin.zip \
	&& mkdir /opt/gradle \
	&& cd /tmp \
	&& unzip -d /opt/gradle gradle-5.6.2-bin.zip \
	&& rm -rf /tmp/*

ENV PATH "/opt/gradle/gradle-5.6.2/bin:$PATH"
```
So this is why there is no problem with the gradle version when using the dockers.  The gradle version used in the dockers is 5.6.2 and the one I installed from Ubuntu is 4.4.1.  So either we add something about installing version 5.6.2 of gradle or use gradlew as suggested by Luis.

TSC21:
> I have done some more checking about gradle versions. The dockerfile https://hub.docker.com/r/px4io/px4-dev-base-bionic/dockerfile has the following:
> 
> ```shell
> # Gradle (Required to build Fast-RTPS-Gen)
> RUN wget -q "https://services.gradle.org/distributions/gradle-5.6.2-bin.zip" -O /tmp/gradle-5.6.2-bin.zip \
> 	&& mkdir /opt/gradle \
> 	&& cd /tmp \
> 	&& unzip -d /opt/gradle gradle-5.6.2-bin.zip \
> 	&& rm -rf /tmp/*
> 
> ENV PATH "/opt/gradle/gradle-5.6.2/bin:$PATH"
> ```
> 
> So this is why there is no problem with the gradle version when using the dockers. The gradle version used in the dockers is 5.6.2 and the one I installed from Ubuntu is 4.4.1. So either we add something about installing version 5.6.2 of gradle or use gradlew as suggested by Luis.

We are not storing the containers in docker hub anymore. The current version of Gradle being used is 6.3: https://github.com/PX4/containers/blob/master/docker/Dockerfile_base-bionic#L80

TSC21:
Also the problem you are getting is no related with the Gradle version but rather with permissions to copy files to system dirs.

Anyway, do you confirm that using `./gradlew install` doesn't come with this permission issue?

AndyBlightLeeds:
I have just tried this again. 
`./gradlew install` fails with the error that I mentioned.  
`sudo ./gradlew install` downloads version 5.6.2 of gradle and then the installation succeeds. 

There is an existing comment beneath my proposed changes that states:
> You might require sudo permissions on the "install" step

So the `sudo` is not needed in the proposed changes.

TSC21:
> I have just tried this again.
> `./gradlew install` fails with the error that I mentioned.
> `sudo ./gradlew install` downloads version 5.6.2 of gradle and then the installation succeeds.
> 
> There is an existing comment beneath my proposed changes that states:
> 
> > You might require sudo permissions on the "install" step
> 
> So the `sudo` is not needed in the proposed changes.

How is it not needed when you stated that with `sudo ./gradlew install` the installation succeeds?

AndyBlightLeeds:
No problem.  I'll update the change to use `sudo ./gradlew install` and remove the line below the code about sudo permissions.

TSC21:
> No problem. I'll update the change to use `sudo ./gradlew install` and remove the line below the code about sudo permissions.

@AndyBlightLeeds when it's stated that it might be needed, it means it might, not that is mandatory. In this case, if you don't have permissions in gradle to copy files to places where you require special permissions, it's natural that you need `sudo`. But it's not mandatory if you have that setup before.

AndyBlightLeeds:
@TSC21 I'm happy with either of the proposed changes.  

On balance, I slightly prefer the `sudo ./gradlew` change as most newbies like me will get it right first time.  People who have a setup  where gradle is run with sudo permissions should not be affected by the extra sudo call.  

In the end, it is your choice :-)

hamishwillee:
@TSC21 This is waiting on you to respond to ^^^

hamishwillee:
@TSC21 ^^^ You are being watched :-)

hamishwillee:
@TSC21 Can you please look at this and decide if it should be merged or discarded. I am going to merge devguide and user guide soon, so outstanding PRs sitting around are a pain.

TSC21 (approved):
Yep this is a better option.

PR #1056, opened 2020-07-08
Add note about spawn location in Gazebo

vorobotics:
Vehicles are spawned at (1.01, 0.98, 0.83) as noted [here](https://github.com/PX4/Firmware/issues/15280).


hamishwillee:
Thanks very much @vorobotics ! Merged.

PR #1057, opened 2020-07-08
Fix MAVProxy broken link

RicardoM17:


hamishwillee:
Thanks @RicardoM17 Good catch. I changed this to the docs link :-)

PR #1060, opened 2020-07-15
setup: no more Java 8 required

julianoes:
We can just use Java 11 or 14 after https://github.com/PX4/jMAVSim/pull/119.

FYI @unipheas @jinger26


hamishwillee:
Thanks @julianoes 
I subedited to make it a little more readable. Also hinted Java 14 is an option, not "the" option.

PR #1081, opened 2020-09-11
rename module template

bkueng:
From https://github.com/PX4/Firmware/pull/15715

### DIFF
--- a/en/apps/module_template.md
+++ b/en/apps/module_template.md
@@ -45,7 +45,7 @@ In summary:
 ## Tasks
 
 The PX4 Firmware contains a template for writing a new application (module) that runs as a task on its own stack:
-[src/templates/module](https://github.com/PX4/Firmware/tree/master/src/templates/module).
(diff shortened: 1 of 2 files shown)


hamishwillee:
Hi @bkueng 
This is fine, merged. 

1. I assume rename was just to make it clear this is a template, not an example?
2. Looking at the cmake, the format of module declaration is a bit awkward `MODULE templates__template_module`. That's fine, but I noticed we don't explain the cmake file. Is it just path back to src? ie if we had `src/fred/jim/somemodule` cmake would expect fred__jim__somemodule?

bkueng:
1. it's for easier search-and-replacing/renaming
2. Yes it follows the dir & file names. This is not strict though, it's just importent to keep it unique across the tree.

hamishwillee:
Thanks. Have tried note point 2 in https://github.com/PX4/Devguide/commit/2d7569fe21eed9af1660c6d5b9d77e07ae8278c7 - would appreciate your sanity check

PR #1082, opened 2020-09-11
posix: update px4 startup arguments

bkueng:
from https://github.com/PX4/Firmware/pull/15714


hamishwillee:
@bkueng Thanks for the update. You've obviously run through this and tested; I have not, but I trust you and it makes sense.

PR #1085, opened 2020-09-21
param show-for-airframe

hamishwillee:
@bkueng This change just updates the normal param guide docs.

Any idea why this would not be showing up in https://dev.px4.io/master/en/middleware/modules_command.html#param ?

### DIFF
--- a/en/advanced/parameters_and_configurations.md
+++ b/en/advanced/parameters_and_configurations.md
@@ -36,6 +36,9 @@ You can use the `-c` flag to show all parameters that have changed (from their d
 param show -c
 ```
 
+You can use `param show-for-airframe` to show all parameters that have changed from their defaults for just the current airframe's definition file (and defaults it imports).
+
+
 ### Exporting and Loading Parameters
 
 You can save any parameters that have been *touched* since all parameters were last reset to their firmware-defined defaults (this includes any parameters that have been changed, even if they have been changed back to their default).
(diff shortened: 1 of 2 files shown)


bkueng:
Thanks Hamish. Don't you want to add it to https://dev.px4.io/master/en/airframes/adding_a_new_frame.html#config-file? That's what it's intended for.

> Any idea why this would not be showing up in https://dev.px4.io/master/en/middleware/modules_command.html#param ?

Is CI not updating automatically? It gets generated corrrectly when running `make module_documentation` in Firmware.

hamishwillee:
> Don't you want to add it to https://dev.px4.io/master/en/airframes/adding_a_new_frame.html#config-file? That's what it's intended for.

@bkueng 

I was a bit lazy here sorry. Can you outline why this is useful and how you "really" use it/when you really need it? I.e. I understand that you can use this to work out which of the values defined in the airframe file have changed (also those included from the defaults). But I don't understand when you need to be able to identify these changes. I guess maybe after tuning so you can easily work out what changed and update the config file? 

Hope the question makes sense?

hamishwillee:
@bkueng 
1. Updated add for airframe - just a note capturing that you can use show-for-airframes to get the delta of params of a tuned airframe with respect to the generic one. Hope this is correct!
2. FYI the param module doc now updated. That must only happen every 24 hours.

OK?

bkueng (approved):

PR #1088, opened 2020-09-28
Document verbosity increase for gazebo

Jaeyoung-Lim:
This documents increasing the verbosity in the simulation through `VERBOSE_SIM` when running SITL simulations

Related PR: https://github.com/PX4/Firmware/pull/15817 https://github.com/PX4/Firmware/pull/15720

### DIFF
--- a/en/simulation/gazebo.md
+++ b/en/simulation/gazebo.md
@@ -375,6 +375,11 @@ The code is available on the [sitl_gazebo repository](https://github.com/px4/sit
 > **Note** The build system enforces the correct GIT submodules, including the simulator.
   It will not overwrite changes in files in the directory.
 
+In case something is not working as expected, you can increase the verbosity of the gazebo simulation through `VERBOSE_SIM`
+```
+export VERBOSE_SIM=1
+make px4_sitl gazebo
+```
 
 ## Further Information
 
--- a/en/simulation/gazebo.md
+++ b/en/simulation/gazebo.md
@@ -367,6 +367,23 @@ gst-launch-1.0  -v udpsrc port=5600 caps='application/x-rtp, media=(string)video
 ! rtph264depay ! avdec_h264 ! videoconvert ! autovideosink fps-update-interval=1000 sync=false
 ```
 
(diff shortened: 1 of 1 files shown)


en/simulation/gazebo.md
```diff
@@ -375,6 +375,11 @@ The code is available on the [sitl_gazebo repository](https://github.com/px4/sit
 > **Note** The build system enforces the correct GIT submodules, including the simulator.
   It will not overwrite changes in files in the directory.
 
+In case something is not working as expected, you can increase the verbosity of the gazebo simulation through `VERBOSE_SIM`
```
TSC21: Should we specify if this is to `gzserver` or `gzclient` or both?

hamishwillee: @Jaeyoung-Lim 
Thanks. I have moved to its own heading/section. This isn't a bad place at all - I just think it is useful to have all the info about flags in one place, and I think a search is more likely to find this. 

FWIW it this only logs the server or the client we should say that (ie I agree with Nuno). If it does both we can ignore.

en/simulation/gazebo.md
```diff
@@ -367,6 +367,23 @@ gst-launch-1.0  -v udpsrc port=5600 caps='application/x-rtp, media=(string)video
 ! rtph264depay ! avdec_h264 ! videoconvert ! autovideosink fps-update-interval=1000 sync=false
 ```
 
+### Verbose Logging
+
```
hamishwillee, line 371: @Jaeyoung-Lim Please check - and if this logs specifically only part of gazebo (server or client) or any other very specific part, we should identify that part. If it is for general use we can probably leave as is.

RicardoM17: @Jaeyoung-Lim wink wink :point_up:

hamishwillee:
@Jaeyoung-Lim I am merging this despite https://github.com/PX4/PX4-Devguide/pull/1088#discussion_r496249197 . This is still useful. Would be good if you could address though.

PR #1091, opened 2020-09-29
Update gimbal_control.md

ghost:
added mavlink gimbal setup, this setup can be used with these set of gimbals.
I tested this one with http://www.viewprotech.com/index.php?ac=article&at=read&did=279
this gimbal and a EO/IR gimbal.

### DIFF
--- a/en/advanced/gimbal_control.md
+++ b/en/advanced/gimbal_control.md
@@ -19,6 +19,15 @@ After selecting the input mode, reboot the vehicle so that the mount driver star
 If the input mode is set to `AUTO`, the mode will automatically be switched based on the latest input.
 To switch from MAVLink to RC, a large stick motion is required.
 
+## MAVLINK Gimbal Support
+To enable a mavlink supported gimbal, (`MNT_MODE_IN`) parameter can be set to **MAVLINK_DO_MOUNT** and (`MNT_MODE_OUT`) should be set to **MAVLINK**. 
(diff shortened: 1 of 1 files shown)


hamishwillee:
Awesome thanks @yuthikasagarage ! I have subedited it, mostly to cross link to relevant resource.

@julianoes Can you please confirm you also are happy with this. One question - "The tricky part is that not all airframes have aux properly mapped which means you're lucky when it works." - according to these docs the mixer is mapped irrespective of what the airframe does. Can you confirm?

julianoes:
I think the instructions are good. Thanks @yuthikasagarage.

> The tricky part is that not all airframes have aux properly mapped which means you're lucky when it works.

I stand by this. I wish it was better.

hamishwillee:
Thanks @julianoes . Unfortunately his section indicates that it doesn't matter what the airframe has mapped. So while I completely believe you that this is tricky, I'm left with inconsistent docs. A mystery.

> A mixer file is required to define the mapping for the output pins and the [mount mixer](https://github.com/PX4/Firmware/blob/master/ROMFS/px4fmu_common/mixers/mount.aux.mix) is automatically selected (this overrides any AUX mixer provided by the airframe configuration).

I am merging anyway, on the basis that this problem pre-exists the change and adds additional value. 

Thank you @yuthikasagarage !

PR #1093, opened 2020-10-03
Fix spelling of actuator

jkemp101:
Just a small fixup. Replaced actautor with actuator


hamishwillee:
@jkemp101 thanks for the fix. Can you please revert all the changes except for the one in **en/simulation/README.md**?

The reason is that the other language versions are auto-generated by round trip through a translation service (this allows us to automatically keep translations in sync with changes to the main source). If I were to merge this, then I would then have to do a merge with the translation service created docs (crowdin).

jkemp101:
@hamishwillee Fixed

hamishwillee:
Thanks @jkemp101 - merged.

PR #1095, opened 2020-10-07
[UPDATE] - Add ROS installation links to ROS menu

TheLegendaryJedi:


hamishwillee:
@TheLegendaryJedi Thanks so much - definite improvement.

I made some minor changes - FYI:
- Pushed into own section - that was just personal taste.
- Removed to some degree the specific naming of versions - ie "Melodic", Gazebo 9. This is for simplified maintenance - it means we will not need to update _this_ doc when we update the installation script to other versions later on. Anyone who wants to can find out what versions are installed by going to the link. Hope that makes sense.

Issue #1061, opened 2020-07-23
RTPS/ROS2 Interface page typo


nichwall:
In the "Creating a ROS2 Advertiser" section (https://dev.px4.io/master/en/middleware/micrortps.html#creating-a-ros2-advertiser), the example file `debug_vect_advertiser.cpp` is said to be in `px4_ros_com/src/listeners` instead of `px4_ros_com/src/advertisers`. It's simple to find where the source file is actually stored, but inconsistent with documentation.

RicardoM17:
Thanks for the heads up @nichwall . I've pushed a fix in #1096 .

FYI next time if you want you can make the changes yourself. Just to make that clear, but raising awareness is the next best thing and the whole community thanks you for that.

@hamishwillee FYI once the PR is merged, this issue can be closed.

hamishwillee:
Closed by https://github.com/PX4/Devguide/commit/6009b061dac10141ff4b799e18d16b9c4d031eb7

hamishwillee:
Thanks @RicardoM17 - much appreciated. 

@nichwall Thanks for posting this! As Ricardo says, easy to fix these yourself - if you click this icon on the top of every page it takes you to github and opens the page for editing. You can then create a PR which includes the actual change for us to review.

![image](https://user-images.githubusercontent.com/5368500/95924115-aa9be900-0e02-11eb-914a-3bda120bcd55.png)

PR #1096, opened 2020-10-12
micrortps: Fix debug_vect_advertiser.cpp path

RicardoM17:
Closes issue https://github.com/PX4/Devguide/issues/1061


TSC21 (approved):
Thanks!

PR #1103, opened 2020-10-28
Update RPi toolchain installation guide

SalimTerryLi:
Update the correct GCC version and aarch64 toolchain.

### DIFF
--- a/en/setup/dev_env_linux_ubuntu.md
+++ b/en/setup/dev_env_linux_ubuntu.md
@@ -78,7 +78,8 @@ To get the build toolchain for Raspberry Pi:
 
 ### GCC (armhf)
 
-Ubuntu software repository provides a set of pre-compiled toolchains. Note that Ubuntu Focal comes up with `gcc-9-arm-linux-gnueabihf` as its default installation, so we must manually install `gcc-8-arm-linux-gnueabihf` and set it as the default toolchain. The following instruction assumes you haven't installed any version of arm-linux-gnueabihf, and will set up the default executable with `update-alternatives`.
+Ubuntu software repository provides a set of pre-compiled toolchains. Note that Ubuntu Focal comes up with `gcc-9-arm-linux-gnueabihf` as its default installation, so we must manually install `gcc-8-arm-linux-gnueabihf` and set it as the default toolchain.
+The following instruction assumes you haven't installed any version of arm-linux-gnueabihf, and will set up the default executable with `update-alternatives`.
 Install them with the terminal command:
 
 ```sh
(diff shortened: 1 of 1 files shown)


hamishwillee:
@bkueng @dagar What @SalimTerryLi  has done here looks great to me. Can you confirm that the versions of compilers indicated etc are what we expect?

SalimTerryLi:
All of those toolchains(armhf and aarch64) won't work on Ubuntu 20.04 host machine, as it depends on libc6-2.29, which is not available on RPi now(still 2.27). Anyone who want to build for RPi must use older Ubuntu 18.04 or docker build. A lot of documentation work.

They can pass the build and provide binary files, but is extreme new so cannot run on actual Pi. The really working one is documented separately in user docs.

hamishwillee:
@SalimTerryLi 

Re https://github.com/PX4/PX4-Devguide/pull/1103#issuecomment-721486508 I added a Warning at the beginning of the RasPi section which just says which Ubuntu version to use. Does that fix this consideration?

If so, do you think this is safe/good to merge?

SalimTerryLi:
Excellent! Thanks for your supplement.
It's generally right in my view, and is OK to merge.

en/setup/dev_env_linux_ubuntu.md
```diff
@@ -63,6 +63,11 @@ sudo add-apt-repository --remove ppa:team-gcc-arm-embedded/ppa
 
 <!-- NOTE: RaPi docker toolchain (for comparison) here: https://github.com/PX4/containers/blob/master/docker/Dockerfile_armhf -->
 
+> **Warning** Use Ubuntu **18.04** or [Docker](https://github.com/PX4/containers/blob/master/docker/Dockerfile_armhf) for building RasPi!
+  *Do not use Ubuntu 20.04*. The *armhf* or *aarch64* toolchains on a Ubuntu 20.04 host machine depend on *libc6-2.29* which is not available on RPi at time of writing (it has *libc6-2.27*).
+
+They can pass the build and provide binary files, but is extreme new so cannot run on actual Pi. The really working one is documented separately in user docs.
```
bkueng: Add link to the user doc?

bkueng (approved):

en/setup/dev_env_linux_ubuntu.md
SalimTerryLi: [This section](https://dev.px4.io/master/en/setup/building_px4.html#other-boards) is a little disorganized. Should all of those boards locate somewhere else?(for example, user docs).
I think only this rough link should be provided here: https://docs.px4.io/master/en/flight_controller/

hamishwillee: @SalimTerryLi 
1. "The really working one is documented separately in user docs." like @bkueng says, where is that link to go to?
2. The information should be in one place. I would be happy for this section you point to go into the user docs. What we would leave here is a link to that. 

We should fix this now and do that other work as a separate action. OK?

SalimTerryLi:
Modified. @hamishwillee

hamishwillee:
Thanks @SalimTerryLi ! I did a minor subedit and will now merge.

Note that the PX4 user guide and Devguide will shortly merge. At that point we might look at moving to having just ONE source of build information :-)

PR #1131, opened 2020-11-16
Fix shell script typo in 'Adding a new airframe'

mortenfyhn:
Just noticed this when reading. Using `==` works with bash, but not with sh.

By the way, the "edit page" button leading to automatically creating a fork and letting me suggest changes is very sleek!

### DIFF
--- a/en/airframes/adding_a_new_frame.md
+++ b/en/airframes/adding_a_new_frame.md
@@ -56,7 +56,7 @@ The next section specifies vehicle-specific parameters, including [tuning gains]
 ```bash
 sh /etc/init.d/rc.fw_defaults
 
-if [ $AUTOCNF == yes ]
+if [ $AUTOCNF = yes ]
 then
 	param set BAT_N_CELLS 2
 	param set FW_AIRSPD_MAX 15


hamishwillee:
Thanks @mortenfyhn . Can you confirm the change works with both bash _and_ sh?


Glad you like the edit button. FYI next week the devguide should merge into the user guide. At that point the edit link will do same thing, but is at end of the page - see the test tree here: https://docs.px4.io/v1.devmerge/en/dev_setup/getting_started.html

mortenfyhn:
Sure, I can confirm that this script:

```sh
AUTOCNF="yes"
if [ $AUTOCNF == yes ]
then
    echo "yes"
fi
```

runs fine with `bash script.sh` but fails with `sh script.sh`, error `script.sh: 2: [: yes: unexpected operator`

While this script:
```sh
AUTOCNF="yes"
if [ $AUTOCNF = yes ]
then
    echo "yes"
fi
```
runs properly with both bash and sh.

hamishwillee:
Thanks very much @mortenfyhn - merged.
FYI I hope to merge devguide into user guide in next couple of days. This has been manually copied across.

PR #1137, opened 2020-11-30
Apply some fixes

elpiel:


hamishwillee:
Hi @elpiel 

I'm going to have to reject this:
1. I can't easily tell what you have changed, because most of these changes just seem to be modifications of the line endings. Could you point to the relevant content changes?
2. We only change the documents in the /en/ tree. The translated versions are then automatically updated with the new strings. 
3. All this content has moved to the user guide. I am going to add a note suggesting that future changes should go in the user guide.

If you could please put changes into the user guide, affecting just the parts of the content that change in the English version (/en/), I'd be more than happy to review again!

elpiel:
@hamishwillee my bad, my editor applied those changes, I will revert them and open it again.
Thanks for letting me know about the translations. It was just easier to make the changes and open a PR :P instead of asking first

hamishwillee:
No worries - usually it is easier! But can you make the change in the user guide as above?

PR #1141, opened 2020-12-25
Fix URL to download ubuntu_sim_ros_melodic.sh

themaverikk:


hamishwillee:
Thanks. FYI all the docs have merged into User guide, so if URL is wrong it would be fixed there.

